Issue #5, opened 2023-05-10
Update dependencies to stay current and remove high security vulnerabilities


d-ryan-ashcraft:
* Update to Krausening 16
* Remove poetry.lock files as we are simulating libraries with those test modules
* Update to at least cryptography 39.0.1 (resolves several high and medium vulnerabilities)
![image](https://github.com/TechnologyBrewery/habushu/assets/118617093/d6d3549a-af14-47ed-870d-d7602a45ac56)

d-ryan-ashcraft:
Alerts closed after commit:
![image](https://github.com/TechnologyBrewery/habushu/assets/118617093/0603a635-1321-4860-9146-eb60f7730776)

PR #6, opened 2023-05-10
#5: :arrow_up: version updates

d-ryan-ashcraft:


habushu-mixology/pyproject.toml
```diff
@@ -7,7 +7,8 @@ license = "MIT License"
 
 [tool.poetry.dependencies]
 python = "^3.9"
-krausening = "13"
+krausening = "16"
+cryptography = "^39.0.1"
```
kdendtler, line 11: Q/A: Do we want to consider using a specific version of cryptography?

kdendtler (approved):

habushu-mixology/pyproject.toml
d-ryan-ashcraft: We could - but my understanding is the due to transitive dependency challenges within PIP/Python in general, using ranges for libraries like this is preferrable

ekoniec1: Long but good read on taking this approach a step forward and **not** capping the upper bounds of dependency versions unless needed: https://iscinumpy.dev/post/bound-version-constraints/

d-ryan-ashcraft:
2.4.0 Release Notes updated to reflect changes.

Issue #8, opened 2023-05-14
Pyenv not activated when cleaning, leading to virtual environment hits


d-ryan-ashcraft:
When running the clean build lifecycle, pyenv is not activated and poetry is assumed to be installed.  This can create challenges with installs being present as well as false positive clean steps where virtual environments are not removed despite setting the `habushu.deleteVirtualEnv` parameter to `true`.

When running clean in this fashion when poetry reports an existing virtual environment via the `poetry env list --full-path` command, we see that Habushu is not detecting nor cleaning the virtual environment:
```
[INFO] -----------< org.technologybrewery.habushu:habushu-mixology >-----------
[INFO] Building habushu::Mixing Drinks with Habushu 2.4.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] ------------------------------[ habushu ]-------------------------------
[INFO] 
[INFO] --- habushu:2.4.0-SNAPSHOT:clean-habushu (default-clean-habushu) @ habushu-mixology ---
[INFO] Executing Poetry command: poetry env info --path
[WARNING] No Poetry virtual environment was detected for deletion.
...
```

d-ryan-ashcraft:
Refactor pyenv and poetry validation and configuration to happen in clean as well as validate.  For now, we'll duplicate the calls - slightly less efficient, but covers all the permutations and still reasonably fast.

d-ryan-ashcraft:
**Test Steps**
- Run `mvn clean install` to ensure that a virtual environment has been created
- Switch the the `habushu-mixology` module in the command line via `cd ./habushu-mixology`
- Run `poetry env list --full-path` to emit what Poetry says is the current virtual environment - should look something like the following:
   ```
   /Users/<usename>/Library/Caches/pypoetry/virtualenvs/habushu-mixology-gy-zFMPc-py3.9
   ```
- Run `mvn clean install -Dhabushu.deleteVirtualEnv=true` and ensure you see the following output for the `clean` phase of the build - particularly the part about removing the virtual environment reported above:
   ```
   [INFO] -----------< org.technologybrewery.habushu:habushu-mixology >-----------
   [INFO] Building habushu::Mixing Drinks with Habushu 2.4.0-SNAPSHOT
   [INFO]   from pom.xml
   [INFO] ------------------------------[ habushu ]-------------------------------
   [INFO] 
[...]

Issue #7, opened 2023-05-14
Bump default version of Python 3.9 to latest patch release


d-ryan-ashcraft:
Update to 3.9.16

d-ryan-ashcraft:
**Test Steps:**
- Ensure pyenv is installed on in your environment
- run `pyenv version` to validate the current version the is the default for pyenv
- If 3.9.16, change it to a different version (e.g., 3.9.14)
- Run `mvn clean install -Pbootstrap` to build the latest version of `habushu-maven-plugin` as this is a snapshot (development) test
- Run `mvn clean install` to run the habushu test projects
- Ensure that INFO output for clean and validate phases for both `habushu-mixology` and `habushu-mixology-consumer` look like:
```
--- habushu:2.4.0-SNAPSHOT:... ---
[INFO] Using Python 3.9.16 (managed by pyenv)
```

d-ryan-ashcraft:
**Final Test Results:**
- `pyenv version` result: 
   ```
   3.9.14 (set by /Users/********/dev/technology-brewery/habushu/.python-version)
   ```
- `mvn clean install -Pbootstrap` result:
   ```
   [INFO] ------------------------------------------------------------------------
   [INFO] BUILD SUCCESS
   [INFO] ------------------------------------------------------------------------
   [INFO] Total time:  10.472 s
   [INFO] Finished at: 2023-05-15T09:58:39-04:00
   [INFO] ------------------------------------------------------------------------
   ```
- `mvn clean install` on test modules result:
   ```
   ...
   [INFO] --- habushu:2.4.0-SNAPSHOT:validate-pyenv-and-poetry (default-validate-pyenv-and-poetry) @ habushu-mixology ---
   [INFO] Using Python 3.9.16 (managed by pyenv)
   ...
   [INFO] --- habushu:2.4.0-SNAPSHOT:validate-pyenv-and-poetry (default-validate-pyenv-and-poetry) @ habushu-mixology-consumer ---
   [INFO] Using Python 3.9.16 (managed by pyenv)
   ...
   ```

d-ryan-ashcraft:
Added entry to release notes: `Updated to default to Python 3.9.16 (from 3.9.14)`

PR #9, opened 2023-05-14
#8 :bug: activate pyenv and python before clean; #7 :arrow_up: bump p…

d-ryan-ashcraft:
…ython 3.9 to latest patch release

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
@@ -1,0 +1,219 @@
+package org.technologybrewery.habushu;
+
+import com.vdurmont.semver4j.Semver;
+import com.vdurmont.semver4j.Semver.SemverType;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugin.logging.Log;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.technologybrewery.habushu.exec.PoetryCommandHelper;
+import org.technologybrewery.habushu.exec.PyenvCommandHelper;
+import org.technologybrewery.habushu.exec.PythonVersionHelper;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Common class that ensures pre-requisite tools that Habushu leverages are installed and available on the
+ * developer's machine to support the same functionality across multiple Mojo implementations. These include:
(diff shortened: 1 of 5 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -0,0 +1,219 @@
...
+ * <li>Required Poetry plugins (currently only {@code poetry-monorepo-dependency-plugin})</li>
+ * </ul>
+ */
+class PyenvAndPoetrySetup {
+
+    /**
+     * Specifies the semver compliant requirement for the default version of Python that
+     * must be installed and available for Habushu to use.
+     */
+    protected static final String PYTHON_DEFAULT_VERSION_REQUIREMENT = "3.9.16";
```
d-ryan-ashcraft, line 35: I: this is the bump to 3.9.16 - now two places where Python can be instantiated, but both use this constant

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Common class that ensures pre-requisite tools that Habushu leverages are installed and available on the
+ * developer's machine to support the same functionality across multiple Mojo implementations. These include:
+ * <ul>
+ * <li>pyenv</li>
+ * <li>Poetry (installed version must satisfy {@link #POETRY_VERSION_REQUIREMENT})</li>
+ * <li>Required Poetry plugins (currently only {@code poetry-monorepo-dependency-plugin})</li>
+ * </ul>
+ */
+class PyenvAndPoetrySetup {
```
d-ryan-ashcraft, line 29: I (informational): this is almost exclusively refactoring prior implementation for reuse in two plugins that cannot share a base class.  Thus we are using a composition approach.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PyenvCommandHelper.java
```diff
@@ -25,29 +27,31 @@ public class PyenvCommandHelper {
...
     public boolean isPyenvInstalled() {
-	try {
-	    execute(Arrays.asList("--version"));
-	} catch (Throwable e) {
-	    return false;
-	}
-	return true;
+        try {
+            String foundVersion = executeWithDebugLogging(Arrays.asList("--version"));
+            logger.debug("Found " + foundVersion);
```
d-ryan-ashcraft, line 39: I: just made some of the debugging with pyenv debug level so it's a more crisp in the info output what is happening that people typically care about

d-ryan-ashcraft (commented):
Recommend ignoring whitespace

kdendtler (approved):

ekoniec1 (approved):

Issue #12, opened 2023-05-25
Array right hand side values break managed dependency functionality


d-ryan-ashcraft:
When working with managed dependencies that are inline tables, Habushu throws an exception:
```
Caused by: java.lang.ClassCastException: class com.electronwill.nightconfig.core.SimpleCommentedConfig cannot be cast to class java.lang.String (com.electronwill.nightconfig.core.SimpleCommentedConfig is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @6eb82908; java.lang.String is in module java.base of loader 'bootstrap')
    at org.technologybrewery.habushu.InstallDependenciesMojo.executeDetailedManagedDependencyMismatchActions (InstallDependenciesMojo.java:284)
    at org.technologybrewery.habushu.InstallDependenciesMojo.processManagedDependencyMismatches (InstallDependenciesMojo.java:251)
    at org.technologybrewery.habushu.InstallDependenciesMojo.execute (InstallDependenciesMojo.java:125)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
```

This occurs when both reading and replacing such values.

d-ryan-ashcraft:
Recreated the defect - by adding the following into `habushu-mixology`:
```
[tool.poetry.dependencies]
...
uvicorn = {version = "^0.18.0", extras = ["standard"]}
```

d-ryan-ashcraft:
Validating the fix on the current SNAPSHOT version with the person that reported it.  Will cut the patch release on confirmation from them.

d-ryan-ashcraft:
Validated.

PR #13, opened 2023-05-25
#12 :bug: support array right hand side values with managed dependencies

d-ryan-ashcraft:


habushu-maven-plugin/pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.technologybrewery.habushu</groupId>
         <artifactId>habushu</artifactId>
-        <version>2.5.0-SNAPSHOT</version>
+        <version>2.4.1-SNAPSHOT</version>
```
kdendtler, line 7: Q: Was this just reverting an incorrect value?

jocobtt (approved):

habushu-maven-plugin/pom.xml
d-ryan-ashcraft: No, this is a patch vice minor release

kdendtler: Should it be done on the master branch then? If there are no other dev changes then it won't have any functional change, but just verifying.

d-ryan-ashcraft: there's not really a need for a master branch on a library.  2.5.0 was just set as the next dev version.  i'll set it back after this to 2.5.0 again.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -386,6 +388,16 @@ protected void performPendingDependencyReplacements(Map<String, TomlReplacementT
...
     }
 
+    /**
+     * Handles escaping with double quotes only if the value is not an array.
+     *
+     * @param valueToEscape value to potentially escape
+     * @return value ready to write to toml file
+     */
+    protected static String escapeTomlRightHandSide(String valueToEscape) {
+        return (!valueToEscape.contains("{")) ? DOUBLE_QUOTE + valueToEscape + DOUBLE_QUOTE : valueToEscape;
```
kdendtler, line 398: Q: For clarity, this isn't actually checking if there is an array but a table instead? 

https://toml.io/en/v1.0.0#inline-table

epwilkins (approved):

habushu-maven-plugin/pom.xml
kdendtler: I'm fine with the management on this versioning scheme but I think there's still a case to be made that we should have a master/main branch specifically for this kind of thing in the future. If we want to release a minor patch quickly but have inflight changes on dev it seems like that would be much easier and safer with separate branches. Can move that to another conversation though.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
d-ryan-ashcraft: I guess that's what I get for commenting a protected method.  Happy to make the change.

habushu-maven-plugin/pom.xml
d-ryan-ashcraft: When we have that problem, I'd be happy to discuss.  In the very short run, I think we try to keep things simple.

kdendtler (approved):

Issue #17, opened 2023-06-01
Improve ability to manage dependencies with dev (snapshot) versions


d-ryan-ashcraft:
When working with a multi-module build leveraging Habushu, it would be nice to handle the conversion of `SNAPSHOT` versions to `dev` versions automatically.  Consider the following scenario:

- `framework`
  - `framework-python-component-a`
  - `framework-python-component-b`
  - `framework-parent-pom` (uses Habushu dependency management and wants to specify the prior 2 components to make migrations easier)

- `consumer-project` (`pom.xml` extends the `framework-parent-pom`)
   - `consumer-project-module-1`

When not using a `SNAPSHOT` version of `framework-parent-pom`, pushing a version would automatically update any dependencies to `framework-python-component-a` or `framework-python-component-b` without issue, making upgrades much more similar to a typical Maven managed dependency situation.

However, when using a `SNAPSHOT` version, we have to make these changes manually, which is a drag and opportunity to get things out of whack.

To resolve this situation, it would be nice if the Habushu managed dependencies also key off of `overridePackageVersion` and replaced `SNAPSHOT` with `dev`.  It would need to also skip managing local development references, such as:
```
habushu-mixology = {path = "../habushu-mixology", develop = true}
```

d-ryan-ashcraft:
Implemented the changes - sample output from newly added BDD tests for this ticket:
- Managed dependency from a standard version to a `-SNAPSHOT` that is auto-corrected to `.dev`:
   ```
   [info] Updated package-a: "1.0.0" --> "1.1.0.dev"
   [info] Updated package-b: "1" --> "2.dev"
   ```
- Managed dependency from a standard version to a `-SNAPSHOT`, left that way b/c `overridePackageVersion=false`
   ```
   [info] Updated package-a: "1.0.0" --> "1.1.0-SNAPSHOT"
   [info] Updated package-b: "1" --> "2-SNAPSHOT"
   ```
- Managed dependency skipped when no version is directly specified:
   ```
   [info] local-dev-package-example-a does not have a specific version to manage - skipping
   [debug]          SimpleCommentedConfig:{path=../local-dev-package-example-a, develop=true}
   [info] local-dev-package-example-b does not have a specific version to manage - skipping
   [debug]          SimpleCommentedConfig:{path=../local-dev-package-example-b, develop=true}
   ```

d-ryan-ashcraft:
[...]

PR #18, opened 2023-06-02
#17: :sparkles: ability to handle managed dependency SNAPSHOT to dev con…

d-ryan-ashcraft:
…version when working with in-flight managed dependencies


habushu-mixology-consumer/pom.xml
```diff
@@ -21,6 +21,22 @@
 				<extensions>true</extensions>
 				<configuration>
 					<pythonVersion>${version.python.default}</pythonVersion>
+					<managedDependencies>
+							<packageDefinition>
+								<packageName>black</packageName>
+								<operatorAndVersion>^23.3.0</operatorAndVersion>
+                                <!-- optional - active is true by default:-->
+                                <active>true</active>
```
aaron-gary: Super minor, but the indenting is off.  At least in GH

aaron-gary (approved):
Looks good

jocobtt (approved):

habushu-mixology-consumer/pom.xml
d-ryan-ashcraft: Good catch - fixed here and in `habushu-mixology`.   Both had tab indentation and now have space-based indentation with reformatting applied.

Issue #38, opened 2023-08-23
Update Habushu to demonstrably leverage the Maven Build Cache


d-ryan-ashcraft:
With Maven's relatively recent addition of the Maven Build Cache, we can get achieve Gradle-like build speed with Maven-like build consistency and (relative) intuitiveness.  This ticket should apply the Maven Build Cache to all of Habushu as well as create the documentation needed for projects to enable Maven Build Cache locally.

This is a significant update as Python dependency management is substantially slower, on average, than Java.  Being able to include and smartly skip as well as only trigger appropriate modules will be a substantial time savings.

**Test Steps**
- Build the habuahu plugin: `mvn clean install -Pbootstrap -Dmaven.build.cache.skipCache=true` - note that it runs a full build and forces a full run in case you had a previously cached version
- Re-run the same command - note that the cache skips the build
- Run habushu module manual test: `mvn clean install -Dmaven.build.cache.skipCache=true` - note that it runs a full build and forces a full run in case you had a previously cached version
- Re-run the same command 
   - Note that the cache skips the build
   - Note that the `dist` directory has been restored with 3 files in it
- Make a change to the `habushu-mixology` module (e.g., to `pyproject.toml`)
- Re-run the same command
   - Note that it runs a full build (that is, the changed module _and_ the dependent module rebuilt automatically downstream

d-ryan-ashcraft:
CI build passed: 
![image](https://github.com/TechnologyBrewery/habushu/assets/118617093/c2ab30b8-46b2-4e3a-92a0-6d58c93d3373)

d-ryan-ashcraft:
**Test results:**
- ✅ Run `mvn clean install -Pbootstrap`
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 2.7.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.532 s]
[INFO] habushu::Maven Plugin for Python Projects .......... SUCCESS [ 12.281 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  14.451 s
[INFO] Finished at: 2023-08-24T09:15:42-04:00
[INFO] ------------------------------------------------------------------------
```
- ✅ Re-run `mvn clean install -Pbootstrap` to validate cache hit (11s faster this round)
```
[...]

PR #41, opened 2023-08-24
#38 :bug: fix missing build directory and update documentation to ref…

d-ryan-ashcraft:
…lect this change

### DIFF
--- a/habushu-mixology-consumer/pom.xml
+++ b/habushu-mixology-consumer/pom.xml
@@ -14,6 +14,7 @@
     <packaging>habushu</packaging>
 
     <build>
+        <directory>dist</directory>
         <plugins>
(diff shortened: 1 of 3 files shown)


habushu-mixology-consumer/pom.xml
```diff
@@ -14,6 +14,7 @@
     <packaging>habushu</packaging>
 
     <build>
```
d-ryan-ashcraft, line 16: I (Informational): I messed this up and took it out of the prior PR as I thought it was superfluous.  I think the test passed in that case b/c the cached version had this configuration.  So I didn't find it until I forced the final test via a clean cache.

kdendtler (approved):

Issue #39, opened 2023-08-23
Update default Python version used by Habushu for easier PyEnv installations


d-ryan-ashcraft:
Habushu currently defaults to Python 3.9.16. However, Both Pyton 3.9.x and 3.10.x have stopped released regular binary installers. This creates a couple of challenges:
- Installation requires a local build of Python, which is relatively slow
- These installations are reportedly unstable on Apple M2 chips

We should strive to stay up to date with the latest stable binary installer.  As of today, that is 3.11.4.

d-ryan-ashcraft:
CI build passed:
![image](https://github.com/TechnologyBrewery/habushu/assets/118617093/a1ad2ed9-77dc-4f8f-86ee-30050947bf9a)

PR #42, opened 2023-08-24
#39 :arrow_up: update default python version from 3.9.16 to 3.11.4 to…

d-ryan-ashcraft:
… ensure latest binary installer is available; :sparkles: catch poetry check error and perform remediation

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -22,7 +22,7 @@ jobs:
         steps:
             - uses: actions/checkout@v3
             - name: Setup Pyenv
-              uses: gabrielfalcao/pyenv-action@v13
+              uses: gabrielfalcao/pyenv-action@v16
             - name: Install Poetry
               uses: snok/install-poetry@v1
             - name: Set up JDK 11
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
@@ -21,27 +21,35 @@ public class InitializeHabushuMojo extends AbstractHabushuMojo {
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
 
-	getLog().info("Validating Poetry-based project structure...");
-	PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
-	poetryHelper.execute(Arrays.asList("check"));
+        getLog().info("Validating Poetry-based project structure...");
+        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        try {
+            poetryHelper.execute(Arrays.asList("check"));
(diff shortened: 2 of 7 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -21,27 +21,35 @@ public class InitializeHabushuMojo extends AbstractHabushuMojo {
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
```
d-ryan-ashcraft, line 23: I (Informational): recommend hiding whitespace

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -21,27 +21,35 @@ public class InitializeHabushuMojo extends AbstractHabushuMojo {
...
-			"Poetry package version set to %s in pyproject.toml does not align with expected POM-derived version of %s",
-			currentPythonPackageVersion, expectedPythonPackageVersion));
-	    }
-
-	}
+        getLog().info("Validating Poetry-based project structure...");
+        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        try {
+            poetryHelper.execute(Arrays.asList("check"));
+        } catch (HabushuException e) {
```
d-ryan-ashcraft, line 28: Seems to occur when you update the minimum python version in your `pyproject.toml`.  It is easy to remiediate, so we catch the error and have poetry resync `pypoject.toml` and `poetry.lock`.  If it were to fail during this, the error would be throw to the user.

.github/workflows/maven.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
         steps:
             - uses: actions/checkout@v3
             - name: Setup Pyenv
```
d-ryan-ashcraft, line 24: I (Informational): Need newer version of pyenv to install python 3.11.4

carter-cundiff (approved):

Ross-Jeffrey (commented):
Change seems fine in my limited experience. I've reviewed the build check and based on that clearing I'd say this is an approved change. I do still see a few spots in the habushu repo that reference python version of 3.10.4 like 
cleanHabushuMojo.java and ValidatePyenvAndPoetryMojo.java, unless I'm misunderstanding things?

d-ryan-ashcraft:
> Change seems fine in my limited experience. I've reviewed the build check and based on that clearing I'd say this is an approved change. I do still see a few spots in the habushu repo that reference python version of 3.10.4 like cleanHabushuMojo.java and ValidatePyenvAndPoetryMojo.java, unless I'm misunderstanding things?

Those are examples of how to specify a non-default version of Python - I _think_ it is useful they are different than the default version, but could be convinced otherwise as well.

Issue #45, opened 2023-09-01
habushu clean no longer also cleans `target`


d-ryan-ashcraft:
Up to 2.6.0, Habushu's clean phase would clean both `dist` and `target`.  Now it appears that only `dist` is being cleaned.  This is a challenge for any standard Maven plugins being used in conjunction with Habushu that might leave things in the target folder.

d-ryan-ashcraft:
Recreated locally.  

### Steps to Recreate
* Add a `target` folder to `habushu-mixology`
* Add a file to the folder
* Run `mvn clean install`
* Validate that the folder still contains the file that was added - should have been cleaned out

d-ryan-ashcraft:
CI build passed: 
<img width="1710" alt="image" src="https://github.com/TechnologyBrewery/habushu/assets/118617093/fed7f05b-be45-4a07-a1c5-12c945ecf488">

PR #46, opened 2023-09-01
#45 :bug: update to also clean target in addition to dist

d-ryan-ashcraft:


README.md
```diff
@@ -597,6 +597,21 @@ normal circumstances.
...
 
+
+#### distDirectory ####
+
+Controls where the clean plugin will delete dist artifacts.  
+
+Default: `${project.basedir}/dist`
+
+
+#### targetirectory ####
```
peter-mcclonski: A: Typo

peter-mcclonski (approved):

tylerashcraft (approved):
Looks good.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -125,11 +131,15 @@ public void execute() throws MojoExecutionException {
...
             filesetsToDelete.add(distArchivesFileset);
+
+            Fileset targetArchivesFileset = createFileset(targetDirectory);
+            filesetsToDelete.add(targetArchivesFileset);
         } catch (IllegalAccessException e) {
             throw new MojoExecutionException("Could not write to private field in Fileset class.", e);
         }
 
         getLog().info(String.format("Deleting distribution archives at %s", distDirectory));
+        getLog().info(String.format("Deleting distribution archives at %s", targetDirectory));
```
epwilkins: A: "distribution archives" seems like a misnomer

README.md
```diff
@@ -597,6 +597,21 @@ normal circumstances.
 
 Default: `${project.basedir}/target/habushu.placeholder.txt`
 
+
+#### distDirectory ####
```
epwilkins, line 601: Q/S: It seems like this should be responsible for more than just the clean plugin functionality. Wouldn't we want to use the same property to drive _where_ Habushu generates artifacts, whether distribution or intermediary/target artifacts?

epwilkins (approved):

README.md
d-ryan-ashcraft: Yeah - it is.  This was actually already in use for that, I just noticed it wasn't documented and added the documentation here.

README.md
d-ryan-ashcraft: Good catch - fixed ✅

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
d-ryan-ashcraft: Good catch - fixed ✅

Issue #47, opened 2023-09-06
Catch pyenv-python mismatches


d-ryan-ashcraft:
## Defect 
If you are using pyenv to manage your python versions but don't have the path configured correctly, a false-positive situation can arrive where pyenv reports the correct target version, but the underlying python version does not match.  

For instance, consider the following scenario: 
- pyenv is configured to use 3.11.4 but not correctly configured on your path
- python version on the path is 3.11.0

When you build, habushu will delegate to pyenv and it will report version 3.11.4 is being using.  But in fact, it will be using 3.11.0.  It's hard to detect because the virtual environment only uses 3.11 in the name.  We noticed only when python was set to an older minor version (3.9), making the virtual environment mismatch more apparent.  

Given how difficult this is to detect, it would be nice if Habushu checked for the situation and errored out with a useful error message to help the user correct the situation.

## Test Steps
* Update path to include a python version you aren't using in pyenv
  * For example: `export PATH=/<pathToPyenv>/.pyenv/versions/<version>/bin:/<pathToPyenv>/.pyenv/bin:<the rest of your path - BUT NOT the valid pyenv shims path!>`
* Run habushu
* Receive an appropriate error message

d-ryan-ashcraft:
Test results:
```
[INFO] --- habushu:2.8.0-SNAPSHOT:validate-pyenv-and-poetry (default-validate-pyenv-and-poetry) @ habushu-mixology ---
[INFO] Using Python 3.11.4 (managed by pyenv)
[INFO] Found Poetry 1.6.1
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 2.8.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.455 s]
[INFO] habushu::Mixing Drinks with Habushu ................ FAILURE [  0.720 s]
[INFO] habushu::Mixing Drinks with Habushu Consumer ....... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.161 s
[INFO] Finished at: 2023-09-07T10:53:03-04:00
[INFO] ------------------------------------------------------------------------
[...]

PR #48, opened 2023-09-07
#47 :goal_net: catch pyenv vs. system python configuration mismatch

d-ryan-ashcraft:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -139,7 +139,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         }
 
         if (!missingRequiredToolMsgs.isEmpty()) {
```
d-ryan-ashcraft, line 141: I (Informational): just making it easier to read the error

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -191,6 +191,17 @@ private String validatateAndConfigurePyenv(List<String> missingRequiredToolMsgs,
                 currentPythonVersion = pyenvHelper.getCurrentPythonVersion();
             }
```
d-ryan-ashcraft, line 193: I: this is the meat of the change - when pyenv isn't configured correctly but we have configured habushu to use it, pyenv says your python version is one thing, but python calls on the path say another.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PythonVersionHelper.java
```diff
@@ -21,17 +22,17 @@ public class PythonVersionHelper {
 
     private static final String PYTHON_COMMAND = "python";
     private static final String PYTHON_3_COMMAND = "python3";
```
d-ryan-ashcraft, line 24: I: some minor sonarlint tidying while I was here - this class is easier to review if you hide whitespace

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PythonVersionHelper.java
```diff
@@ -41,50 +42,81 @@ public PythonVersionHelper(File workingDirectory, String desiredPythonVersion) {
...
-	String version = execute(Collections.singletonList("--version"));
-	return version.replaceAll(extractVersionRegex, "");
+        String version = quietlyExecute(Collections.singletonList("--version"));
+        return version.replaceAll(EXTRACT_VERSION_REGEX, "");
     }
 
     /**
      * Executes a python command with the given arguments, logs the executed
-     * command, and returns the resultant process output as a string.
+     * at DEBUG level, and returns the resultant process output as a string.
```
d-ryan-ashcraft, line 51: I: creature comfort so we just have less output since this now runs every time we get the pyenv version

peter-mcclonski (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PythonVersionHelper.java
```diff
@@ -21,17 +22,17 @@ public class PythonVersionHelper {
 
     private static final String PYTHON_COMMAND = "python";
     private static final String PYTHON_3_COMMAND = "python3";
-    private static final String pythonVersion3Regex = "^3.*";
-    private static final String extractVersionRegex = "^.*?(?=([0-9]))";
+    private static final String PYTHON_VERSION_3_REGEX = "^3.*";
+    private static final String EXTRACT_VERSION_REGEX = "^.*?(?=([\\d]))";
```
epwilkins: A: Does it need to be in a character class block if we're using the built-in character class?
```suggestion
    private static final String EXTRACT_VERSION_REGEX = "^.*?(?=(\\d))";
```

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PythonVersionHelper.java
d-ryan-ashcraft: Good catch - updated ✅

Issue #62, opened 2023-11-14
Support Baton-drive automatic migrations, apply to monorepo dependencies moving into the `[tool.poetry.group.monorepo.dependencies]` group


d-ryan-ashcraft:
Integrate Baton so we can perform common migration implictly.

Specifically, configure migrating monorepo dependencies moving into the `[tool.poetry.group.monorepo.dependencies]` group as the first Baton migration.

d-ryan-ashcraft:
### Test Steps
 * Check out habushu project at https://github.com/TechnologyBrewery/habushu
 * Run `mvn clean install -Pbootstrap`
    * Verify that all unit tests passed successfully (newly added BDD specification `monorepo-migrations.feature` is included, so this validated functionality in an automated fashion)
 * Update `habushu/habushu-mixology-consumer/pyproject.toml` as follows:
     * Comment out the following lines (add the `#` to the front):
     ```toml
     #[tool.poetry.group.monorepo.dependencies]
     #habushu-mixology = {path = "../habushu-mixology", develop = true} 
     ```
    * Run `mvn clean install -pl :habushu-mixology-consumer`
    * Validate that `pyproject.toml` has no further changes 
* Update `habushu/habushu-mixology-consumer/pyproject.toml` as follows:
    * Move the `habushu-mixology` dependency to the `[tool.poetry.dependencies]` group:
     ```toml
     [tool.poetry.dependencies]
     python = "^3.11"
     habushu-mixology = {path = "../habushu-mixology", develop = true} 
     ```
    * Validate that `pyproject.toml` has a new `[tool.poetry.group.monorepo.dependencies]` that now contains the `habushu-mixology` dependency:
     ```toml
     [tool.poetry.dependencies]
     python = "^3.11"
     
     [tool.poetry.group.monorepo.dependencies]
     habushu-mixology = {path = "../habushu-mixology", develop = true}
     ```
* Do not `habushu/habushu-mixology-consumer/pyproject.toml` to verify that if the toml file is already in a desired state, no changes are made:
    * Run `mvn clean install -pl :habushu-mixology-consumer`
    * Validate that `pyproject.toml` has no further changes (monorepo group still exists with the same dependency)
* Update `habushu/habushu-mixology-consumer/pyproject.toml` as follows:
    * Move the `habushu-mixology` dependencies back to the `[tool.poetry.dependencies]`, but leave the now-empty group `[tool.poetry.group.monorepo.dependencies]` there:
     ```toml
[...]

PR #63, opened 2023-11-14
#62 :sparkles: initial Baton integration to support automated migrati…

d-ryan-ashcraft:
…on for monorepo dependencies

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
@@ -1,0 +1,147 @@
+package org.technologybrewery.habushu.migration;
+
+import com.electronwill.nightconfig.core.CommentedConfig;
+import com.electronwill.nightconfig.core.Config;
+import com.electronwill.nightconfig.core.file.FileConfig;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.baton.AbstractMigration;
+import org.technologybrewery.baton.BatonException;
+import org.technologybrewery.habushu.HabushuException;
+import org.technologybrewery.habushu.util.TomlReplacementTuple;
+import org.technologybrewery.habushu.util.TomlUtils;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Automatically migrates any monorepo dependencies (e.g., foo = {path = "../foo", develop = true}) in the main
+ * [tool.poetry.dependencies] group into the [tool.poetry.group.monorepo.dependencies] group instead.  As noted in the
+ * project's README.md, this prevents these dependencies from causing issues when Poetry projects are exported in
+ * development releases.
+ */
+public class CustomMonorepoGroupMigration extends AbstractMigration {
+
+    public static final Logger logger = LoggerFactory.getLogger(CustomMonorepoGroupMigration.class);
+
+    protected Map<String, TomlReplacementTuple> replacements = new HashMap<>();
+
+    protected boolean hasExistingMonoRepoDependenciesGroup;
+
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        replacements.clear();
+        hasExistingMonoRepoDependenciesGroup = false;
+
+        boolean shouldExecute = false;
+        try (FileConfig tomlFileConfig = FileConfig.of(file)) {
+            tomlFileConfig.load();
+
+            Optional<Config> toolPoetryDependencies = tomlFileConfig.getOptional(TomlUtils.TOOL_POETRY_DEPENDENCIES);
+            if (toolPoetryDependencies.isPresent()) {
+                Config foundDependencies = toolPoetryDependencies.get();
+                Map<String, Object> dependencyMap = foundDependencies.valueMap();
+
+                for (Map.Entry<String, Object> dependency : dependencyMap.entrySet()) {
+                    String packageName = dependency.getKey();
+                    Object packageRhs = dependency.getValue();
+                    if (TomlUtils.representsLocalDevelopmentVersion(packageRhs)) {
+                        String packageRshAsString = TomlUtils.convertCommentedConfigToToml((CommentedConfig) packageRhs);
+                        logger.info("Found local dependency not within monorepo group! ({} = {})", packageName, packageRshAsString);
+                        TomlReplacementTuple replacementTuple = new TomlReplacementTuple(packageName, packageRshAsString, "");
+                        replacements.put(packageName, replacementTuple);
+                        shouldExecute = true;
+                    }
+                }
+            }
+
+            Optional<Config> toolPoetryMonorepoDependencies = tomlFileConfig.getOptional(TomlUtils.TOOL_POETRY_GROUP_MONOREPO_DEPENDENCIES);
+            if (toolPoetryMonorepoDependencies.isPresent()) {
+                hasExistingMonoRepoDependenciesGroup = true;
+            }
+        }
+
+        return shouldExecute;
+    }
+
+    @Override
+    protected boolean performMigration(File pyProjectTomlFile) {
+        String fileContent = StringUtils.EMPTY;
+        try (BufferedReader reader = new BufferedReader(new FileReader(pyProjectTomlFile))) {
+            String line = reader.readLine();
+            boolean injectAfterNextEmptyLine = false;
+
+            while (line != null) {
+                boolean addLine = true;
+                boolean isEmptyLine = line.isBlank();
+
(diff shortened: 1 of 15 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Automatically migrates any monorepo dependencies (e.g., foo = {path = "../foo", develop = true}) in the main
+ * [tool.poetry.dependencies] group into the [tool.poetry.group.monorepo.dependencies] group instead.  As noted in the
+ * project's README.md, this prevents these dependencies from causing issues when Poetry projects are exported in
+ * development releases.
+ */
+public class CustomMonorepoGroupMigration extends AbstractMigration {
```
epwilkins, line 29: This touches on a comment I almost left on the baton PR, but what if someone wants to opt out of this?  Since the migration runs on every build, you couldn't just change it back.  I guess just shut off the migration mojo by setting the phase to none, but then you wouldn't get other/future migrations.

habushu-maven-plugin/src/test/resources/specifications/monorepo-migrations.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  Scenario: Migrate monorepo dependencies does not change a file with non applicable dependencies
+    Given an existing pyproject.toml without any monorepo dependencies
+    When Habushu migrations execute
+    Then no migration was perform
+
+  Scenario: Migrate monorepo dependencies does not change a file with all monorepo dependencies already in monorepo group
+    Given an existing pyproject.toml file with three monorepo dependencies already in the tool.poetry.group.monorepo.dependencies group
+    When Habushu migrations execute
+    Then no migration was perform
```
jocobtt: Small typo on lines 23 & 18 it seems. Should be performed?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
```diff
@@ -0,0 +1,147 @@
...
+                    if (key == null) {
+                        key = line.substring(0, line.indexOf(TomlUtils.EQUALS));
+                    }
+
+                    if (key != null) {
+                        key = key.strip();
+
+                        TomlReplacementTuple matchedTuple = replacements.get(key);
+                        if (matchedTuple != null) {
+                            // skip this line, we will add it back to [tool.poetry.group.monorepo.dependencies] later
```
peter-mcclonski, line 96: A: Do we have an open issue for this yet, just so we keep track?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
d-ryan-ashcraft: Correct - for 0.1.0, that's how they will have to do it.  But in a subsequent change to Baton, we'll let people inactivate migrations in a "line item veto" style.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
d-ryan-ashcraft: No need - it does that later in the code block.  It's just a note about how the processing will work.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
```diff
@@ -0,0 +1,147 @@
...
+        try (BufferedReader reader = new BufferedReader(new FileReader(pyProjectTomlFile))) {
+            String line = reader.readLine();
+            boolean injectAfterNextEmptyLine = false;
+
+            while (line != null) {
+                boolean addLine = true;
+                boolean isEmptyLine = line.isBlank();
+
+                if (line.contains(StringUtils.SPACE) || line.contains(TomlUtils.EQUALS)) {
+                    String key = line.substring(0, line.indexOf(StringUtils.SPACE));
```
peter-mcclonski, line 85: S: Somewhat nervous of exceptions in the case where `line.contains(StringUtils.SPACE) == false`.  I want to say (could be very wrong) that `indexOf` will return `-1`, and then `substring(0, -1)` throws an exception.

peter-mcclonski: I could also be completely wrong here, but I feel like I remember substring not liking that.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
```diff
@@ -0,0 +1,147 @@
...
+                        if (matchedTuple != null) {
+                            // skip this line, we will add it back to [tool.poetry.group.monorepo.dependencies] later
+                            addLine = false;
+                        }
+                    }
+
+                } else if (line.contains("[") && line.contains("]")) {
+                    String key = line.strip();
+
+                    if (hasExistingMonoRepoDependenciesGroup && (key.equals("[" + TomlUtils.TOOL_POETRY_GROUP_MONOREPO_DEPENDENCIES + "]"))) {
```
peter-mcclonski, line 104: A: We use this same string multiple times (including the very next condition).  For readability, maybe extract to a var or a string-builder method?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
peter-mcclonski: Ignore, just got to the part where this is addressed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/TomlUtils.java
```diff
@@ -0,0 +1,110 @@
...
+        if (config.get(VERSION) != null) {
+            sb.append(VERSION).append(" = \"").append(config.get(VERSION).toString()).append("\"");
+            List<String> extras = config.get(EXTRAS);
+            if (CollectionUtils.isNotEmpty(extras)) {
+                sb.append(", ").append(EXTRAS).append(" = [");
+                // NB: if we expect more complex values, such as multiple extras, more work would need to be done for
+                // both consistent formatting and comparison of these values.  However, at the time of initially writing
+                // this method, there isn't a clear demand signal, so we are going to KISS for now:
+
+                for (int i = 0; i < extras.size(); i++) {
```
peter-mcclonski, line 89: A: At a first glance, this seems like it could be abbreviated to a `String.join(...)` call?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
d-ryan-ashcraft: Good catch - though I think the core issue is actually this statement:
```
if (line.contains(StringUtils.SPACE) || line.contains(TomlUtils.EQUALS)) {
```

Which should be changed to an `&&` instead, which will protect against this case and is functionally more correct as well.

d-ryan-ashcraft: ✅ Resolved

habushu-maven-plugin/src/test/resources/specifications/monorepo-migrations.feature
d-ryan-ashcraft: ✅ Good catch - resolved

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomMonorepoGroupMigration.java
d-ryan-ashcraft: The next condition is actually a different group id. (monorepo dependencies vs. normal dependencies)

This is one of these scenarios where nothing is super readable.  Could add a method that only ever gets executed once, but i tend to think what is there is more readable in the situation (though certainly arguable).  I borrowed this from another location in Habushu and remember having this same debate with I PR-ed that code - but ultimately was okay with it.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/TomlUtils.java
d-ryan-ashcraft: Seemed fair, but in applying the update, it gets even more ugly w/ the double quotes, so leaving as is.

jocobtt (approved):
Looks good to me.

jocobtt (approved):

peter-mcclonski (approved):

Issue #59, opened 2023-10-16
Skip if required behave directories if they do not exist


d-ryan-ashcraft:
When executing a project, if the needed behave directories do not exist, the build will fail.  It would be nice to skip execution of behave in this instance rather than be forced to create and commit empty directories.

For example:
```
[ERROR] Failed to execute goal org.technologybrewery.habushu:habushu-maven-plugin:2.8.0:behave-bdd-test 
(default-behave-bdd-test) on project foo-bar-blah: Could not load behave features directory: 
/Users/tgwynn/dev/foo-bar-blah/tests/features
```

d-ryan-ashcraft:
CI build passed:
![image](https://github.com/TechnologyBrewery/habushu/assets/118617093/bf84bc87-7800-4b97-b36d-f13fc53c7f40)

PR #64, opened 2023-11-17
#59 :bug: skip test execution if features directory does not exist

d-ryan-ashcraft:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -1,15 +1,5 @@
 package org.technologybrewery.habushu;
```
d-ryan-ashcraft, line 1: I (Informational): ignore whitespace

jacksondelametter (approved):

PR #65, opened 2023-12-05
Update the Poetry Core version in the Habushu project to the required version

satishmanda:
AIOPS-2767 Added functionality to upgrade the poetry-core version required by the Habushu

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
@@ -1,0 +1,162 @@
+package org.technologybrewery.habushu.migration;
+
+import com.electronwill.nightconfig.core.CommentedConfig;
+import com.electronwill.nightconfig.core.Config;
+import com.electronwill.nightconfig.core.file.FileConfig;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.baton.AbstractMigration;
+import org.technologybrewery.baton.BatonException;
+import org.technologybrewery.habushu.HabushuException;
+import org.technologybrewery.habushu.util.TomlReplacementTuple;
+import org.technologybrewery.habushu.util.TomlUtils;
+import org.technologybrewery.habushu.util.PoetryUtil;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+
+/**
+ * Automatically migrates Poetry-core version in the [build-system] group to the
+ * version where the major version and minor version are inline with the
+ * Poetry-core version required by Habushu defined by the POETRY_CORE_VERSION_REQUIREMENT in the
+ * PoetryUtil. As noted in the project's README.md, this prevents these dependencies from causing issues
+ * when Poetry projects are exported in development releases.
+ */
+public class CustomPoetrycoreVersionMigration extends AbstractMigration {
+
+    public static final Logger logger = LoggerFactory.getLogger(CustomPoetrycoreVersionMigration.class);
+
+    protected Map<String, TomlReplacementTuple> replacements = new HashMap<>();
+
+    private boolean isMajorVersionUpdateRequired;
+    private boolean isMinorVersionUpdateRequired;
+
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        try (FileConfig tomlFileConfig = FileConfig.of(file)) {
+            tomlFileConfig.load();
+            Optional<Config> toolBuildSystem = tomlFileConfig.getOptional(TomlUtils.BUILD_SYSTEM);
+            if (toolBuildSystem.isPresent()) {
+                Config buildSystem = toolBuildSystem.get();
+                Map<String, Object> dependencyMap = buildSystem.valueMap();
+                for (Map.Entry<String, Object> dependency : dependencyMap.entrySet()) {
+                    isMajorVersionUpdateRequired = false;
+                    isMinorVersionUpdateRequired = false;
+                    // check if we need to upgrade the poetry-core version.
+                   if(isPoetrycoreUpgradeRequired(dependency)) {
+                       return true;
+                   }
+                }
+            }
+        }
+        return false;
+    }
+
+    @Override
+    protected boolean performMigration(File pyProjectTomlFile) {
+        String fileContent = StringUtils.EMPTY;
+        try (BufferedReader reader = new BufferedReader(new FileReader(pyProjectTomlFile))) {
+            String line = reader.readLine();
+            while (line != null) {
+                boolean isEmptyLine = line.isBlank();
+
+                if (line.contains(StringUtils.SPACE) && line.contains(TomlUtils.EQUALS)) {
+                    String key = line.substring(0, line.indexOf(StringUtils.SPACE));
+                    if (key == null) {
+                        key = line.substring(0, line.indexOf(TomlUtils.EQUALS));
+                    }
+
+                    if (key != null) {
+                        key = key.strip();
+
+                        TomlReplacementTuple matchedTuple = replacements.get(key);
+                        if ((matchedTuple != null) && (line.contains(TomlUtils.POETRY_CORE)) && (line.contains(TomlUtils.EQUALS))) {
+                            // update the poetry-core major version upgrade if required.
+                            StringBuilder stringBuilder = new StringBuilder(line);
+                            if(isMajorVersionUpdateRequired){
+                                int majorIndex = line.lastIndexOf(TomlUtils.EQUALS) + 1;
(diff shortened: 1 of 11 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
```diff
@@ -0,0 +1,162 @@
...
+            tomlFileConfig.load();
+            Optional<Config> toolBuildSystem = tomlFileConfig.getOptional(TomlUtils.BUILD_SYSTEM);
+            if (toolBuildSystem.isPresent()) {
+                Config buildSystem = toolBuildSystem.get();
+                Map<String, Object> dependencyMap = buildSystem.valueMap();
+                for (Map.Entry<String, Object> dependency : dependencyMap.entrySet()) {
+                    isMajorVersionUpdateRequired = false;
+                    isMinorVersionUpdateRequired = false;
+                    // check if we need to upgrade the poetry-core version.
+                   if(IsPoetrycoreUpgradeRequired(dependency)) {
```
cpointe-ibllanos: A: Minor, the if statement is unnecessary.
`return IsPoetrycoreUpgradeRequired(dependency)`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
```diff
@@ -0,0 +1,162 @@
...
+                            }
+
+                            // update the poetry-core minor version if required.
+                            if (isMinorVersionUpdateRequired) {
+                                int minorVersionIndex = line.indexOf(TomlUtils.DOT) + 1;
+                                int endMinorVersionIndex = line.lastIndexOf(TomlUtils.DOT);
+                                String strPoetryCoreReqMinorVersion = PoetryUtil.getMinorReqVersion(PoetryUtil.POETRY_CORE_VERSION_REQUIREMENT);
+                                stringBuilder = stringBuilder.replace(minorVersionIndex, endMinorVersionIndex, strPoetryCoreReqMinorVersion);
+                                line = stringBuilder.toString();
+                                logger.info("Updating the Poetry-core major version to {}.", strPoetryCoreReqMinorVersion);
```
cpointe-ibllanos: A: Typo? The string refers to the major version even though this is for the minor. 

Also, would the logs on line 99 and 89 be better suited for logger.debug? Doesn't make sense to say it's upgrading twice in the info logs every time; I think the single info statement on line 101 is the most helpful info log for a normal build.

cpointe-ibllanos (approved):
Looks good to me, just minor comments. Thanks for utilizing good practices like clear variable naming, properly defined constants, and clear test steps.

I'd get a second opinion since I'm pretty new to Habushu's codebase however.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
```diff
@@ -0,0 +1,162 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+
+/**
```
d-ryan-ashcraft, line 25: A (Administrative): minor, but best if we don't note the versions here as they will likely drift when we upgrade the constants but miss them here.  Recommend you refine this to say "inline with major and minor versions specified in `PoetryUtil`" or similar.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
d-ryan-ashcraft: S (Substantive): comment says major, but seems like it should be minor.

d-ryan-ashcraft: Seems like 3 info level output messages is a little heavy too - maybe the major and minor lines should become debug or be removed?

Also seems like major should spit out the version it is updating to for consistency, if kept.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
```diff
@@ -0,0 +1,162 @@
...
+                            logger.info("Updated the Poetry-core version to {} required by Habushu.", PoetryUtil.POETRY_CORE_VERSION_REQUIREMENT);
+                        }
+                    }
+                }
+                fileContent += line + "\n";
+                line = reader.readLine();
+            }
+
+        } catch (IOException e) {
+            throw new HabushuException("Problem reading pyproject.toml while updating version for poetry-core dependencies!", e);
```
d-ryan-ashcraft: S: This and the next exception - seems like `poetry-core dependencies` should be something like `the build system's poetry-core version`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
```diff
@@ -0,0 +1,162 @@
...
+
+        try {
+            TomlUtils.writeTomlFile(pyProjectTomlFile, fileContent);
+        } catch (IOException e) {
+            throw new BatonException("Problem while writing dependencies to TomlFile during upgrade of poetry-core version!", e);
+        }
+        return true;
+    }
+
+    private boolean IsPoetrycoreUpgradeRequired(Map.Entry<String, Object> dependency) {
```
d-ryan-ashcraft: S: `IsPoetrycoreUpgradeRequired" should be lowercased to "isPoetrycoreUpgradeRequired"

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
```diff
@@ -0,0 +1,162 @@
...
+
+                if ((poetryCoreMajorVer == poetryCoreReqMajorVersion && poetryCoreMinorVer < poetryCoreReqMinorVersion)) {
+                    isMinorVersionUpdateRequired = true;
+                }
+
+                // Return true if major or minor version upgarde is required.
+                if(isMajorVersionUpdateRequired || isMinorVersionUpdateRequired) {
+                    TomlReplacementTuple replacementTuple = new TomlReplacementTuple(packageName, poetrycoreVer, "poetry-core>=1.6.0");
+                    replacements.put(packageName, replacementTuple);
+                    logger.info("Found Poetry-core dependency with version : {} which is less than the required version of 1.6.0 or higher for Habushu. Poetry-core version dependency will be updated to 1.6.0.", poetrycoreVer);
```
d-ryan-ashcraft: A: seems like a 4th place this is getting logged - let's keep it tidy, 1 info level log.  We can be more verbose at debug level, if it makes sense.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/PoetryUtil.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public static final String POETRY_VERSION_REQUIREMENT = "^1.5.0";
+
+    /**
+     * Specifies the semver compliant requirement for the version of Poetry-core that
+     * must be installed and available for Habushu to use.
+     */
+    public static final String POETRY_CORE_VERSION_REQUIREMENT = "^1.6.0";
+
+    /**
```
d-ryan-ashcraft: S: these don't seem like Poetry related Util methods, maybe they belong elsewhere?

d-ryan-ashcraft (changes requested):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
satishmanda: Yes. It should be at the debug level. Made the change to log as debug.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/PoetryUtil.java
satishmanda: I am thinking of adding them to the TomlUtil class as these helper methods can be used to extract major and minor version.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/CustomPoetrycoreVersionMigration.java
satishmanda: We are looping through the dependency map and iterate through all the elements to see if we any of it requires the upgrade. We need the if statement here incase any one of them requires upgrade.

d-ryan-ashcraft (approved):

PR #73, opened 2023-12-19
67 secure install

georgit:
The goal of this PR is to allow Habushu to use maven username/password when pulling dependencies.

Currently, Habushu does not use user credentials when pulling dependencies and instead pulls them anonymously. This is leading to some issues for dependencies that require credentials. We would like to fix this. These credentials are in the settings.xml file in the .m2 directory.

Along the way, we found that the HabushuUtil file was missing the first level of indentation for all methods. After some though, we think the most likely reason for this is a copy/paste error and decided it was worth fixing. Maybe there is some convention of not indenting in a final class in Java, but we didn't think that likely.


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -147,6 +193,24 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             log.info("Checking for updates to poetry-monorepo-dependency-plugin...");
             poetryHelper.installPoetryPlugin("poetry-monorepo-dependency-plugin");
         }
+
+        String username = null;
+        String password = null;
+
+        if (StringUtils.isNotEmpty(this.pypiRepoId)) {
+            username = HabushuUtil.findUsernameForServer(this.settings, this.pypiRepoId);
+            password = HabushuUtil.findPasswordForServer(settings, pypiRepoId, decryptPassword);
```
aaron-gary: minor:  Just for consistency, can you use --> this.settings and this.pypiRepoId and  this.decryptPassword.  Or, alternatively, change the line above to drop --> this.blahblah

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -147,6 +193,24 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+        }
+
+        if (StringUtils.isNotEmpty(pypiRepoUrl)) {
```
aaron-gary: If the pypiRepoUrl is empty, should we log an error or throw and exception?  Not sure what the best approach is here.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -82,50 +94,50 @@ public static void runBashScript(String bashScriptPath) {
...
-	    String[] command;
-	    if (parameters != null && parameters.length > 0) {
-		command = new String[parameters.length + 1];
-
-		for (int i = 0; i < parameters.length; i++) {
-		    command[i + 1] = parameters[i];
+		logger.debug("Running bash script located at {}.", bashScriptPath);
+
+		try {
+			String[] command;
```
aaron-gary, line 51: Looks like the indentation is off.  Can you reformat this file/

aaron-gary (approved):
Left some minor comments

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
georgit: Done.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
georgit: Sorry I thought I had gotten all of the indentation sorted out, I gave it another pass just now.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -147,6 +193,24 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             log.info("Checking for updates to poetry-monorepo-dependency-plugin...");
             poetryHelper.installPoetryPlugin("poetry-monorepo-dependency-plugin");
         }
+
+        String username = null;
+        String password = null;
+
+        if (StringUtils.isNotEmpty(this.pypiRepoId)) {
+            username = HabushuUtil.findUsernameForServer(this.settings, this.pypiRepoId);
+            password = HabushuUtil.findPasswordForServer(this.settings, this.pypiRepoId, this.decryptPassword);
```
epwilkins: S: I would move these two methods to `AbstractHabushuMojo` since they are highly specific to Maven execution, and just pass the credentials into the constructor of this object.

epwilkins: This also allows you to remove `pypiRepoId`, `decryptPassword`, and, `settings` from the constructor so it ends up simplifying this class.

georgit: If I understand your suggestion, my latest push should have addressed this. Those methods are now in AbstractHabushuMojo and are called by ValidatePyEndAndPoetryMojo, which then passes the values down to PyEnvAndPoetrySetup

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -110,7 +110,8 @@ private void clean() throws MojoExecutionException {
         if (deleteVirtualEnv) {
             try {
                 PyenvAndPoetrySetup configureTools = new PyenvAndPoetrySetup(pythonVersion, usePyenv,
-                        patchInstallScript, workingDirectory, rewriteLocalPathDepsInArchives, getLog());
+                        patchInstallScript, workingDirectory, rewriteLocalPathDepsInArchives,
+                        null, null, null, null, getLog());
```
epwilkins: A: Recommend overloading the constructor so we don't have pass in a bunch of `nulls`. Also gives us the ability in the javadocs to specify that without the credentials/server info, authentication capabilities will be non-existent/limited.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -147,6 +160,15 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
             poetryHelper.installPoetryPlugin("poetry-monorepo-dependency-plugin");
         }
+
+        if (StringUtils.isEmpty(username) || StringUtils.isEmpty(password)) {
+            log.info(String.format("Did not find username and password for the <server> with <id> %s. Will pull anonymously", pypiRepoId));
+        }
+        else {
+            log.info(String.format("TESTING CHANGE Adding username and password configuration for %s at %s", pypiRepoId, pypiRepoUrl));
+            poetryHelper.execute(
+                    Arrays.asList("config", String.format("http-basic.%s", pypiRepoId), username, password));
```
epwilkins, line 192: C: If it's already configured with authentication information we shouldn't overwrite it.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -91,13 +99,18 @@ class PyenvAndPoetrySetup {
...
         this.pythonVersion = pythonVersion;
         this.usePyenv = usePyenv;
         this.patchInstallScript = patchInstallScript;
         this.baseDir = baseDir;
         this.rewriteLocalPathDepsInArchives = rewriteLocalPathDepsInArchives;
         this.log = log;
+        this.username = username;
+        this.password = password;
+        this.pypiRepoId = pypiRepoId;
+        this.pypiRepoUrl = pypiRepoUrl;
```
epwilkins: A: If we're only using this for a log statement, I'd drop it completely.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -91,13 +99,18 @@ class PyenvAndPoetrySetup {
...
+                               String username, String password, String pypiRepoId, String pypiRepoUrl, Log log) {
         this.pythonVersion = pythonVersion;
         this.usePyenv = usePyenv;
         this.patchInstallScript = patchInstallScript;
         this.baseDir = baseDir;
         this.rewriteLocalPathDepsInArchives = rewriteLocalPathDepsInArchives;
         this.log = log;
+        this.username = username;
+        this.password = password;
+        this.pypiRepoId = pypiRepoId;
```
epwilkins, line 112: A: If it simplifies the code, it would make sense to wrap the repo credentials and ID in an object.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -147,6 +160,15 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             log.info("Checking for updates to poetry-monorepo-dependency-plugin...");
             poetryHelper.installPoetryPlugin("poetry-monorepo-dependency-plugin");
         }
+
+        if (StringUtils.isEmpty(username) || StringUtils.isEmpty(password)) {
+            log.info(String.format("Did not find username and password for the <server> with <id> %s. Will pull anonymously", pypiRepoId));
```
epwilkins: A: This is a bit of a lie. If poetry already has auth configured for the server then even if the username/password are empty here doesn't mean we'll use anonymous auth.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
epwilkins: On follow-up investigation, the only auth method that poetry currently supports is http-basic auth. So there's no risk that we would be preempting a more secure authentication method by always overwriting it, and by doing so we make it easier for the user to update the auth info.

This feedback can be ignored.

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
georgit: Done

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
georgit: Done

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
georgit: For now I will leave it because the object could sometimes use the repo url and sometimes not, but this could be good for the future. I'm just not sure it simplifies the code right now.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
georgit: I edited the message in my latest commit.

PR #75, opened 2023-12-21
Add new `cacheWheels:boolean` configuration to the habushu-maven-plugin that caches build wheel files. 

JeffreyRoss:
In order to make mono-repository wheel files more available across multiple modules we want to cache them into a reliable location. This change introduces a new **habushu-maven-plugin** configuration **_cacheBuildWheelFiles:boolean_**

When set to true will copy the build's wheel file into poetry's cache directory
`{poetry-cache-dir}/cache/wheels/{artifactId}/*.whl `
When set to false nothing will be cached.
Default is false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,12 @@ Folder in which Python test files are located - should align with Poetry's proje
 
 Default: `${project.basedir}/tests`
 
+#### cacheBuildWheelFiles ####
+
+Enables or Disables the copying of poetry build wheel files into Poetry cache.
+
+Default: `false`
+
 #### managedDependencies ####
 
 Optional set of dependencies to manage across modules extending a parent pom. This allows packages to be managed to a 
@@ -693,4 +699,3 @@ If you are working on Habushu, please be aware of some nuances in working with a
 * `mvn clean install -Pbootstrap`: Builds the `habushu-maven-plugin` such that the custom `habushu` lifecycle may be utilized within subsequent builds.
   * **NOTE:** If updates are made to the `habushu` lifecycle (i.e. updates to the `habushu` lifecycle mapping configuration made in `habushu-maven-plugin/src/main/resources/META-INF/plexus/components.xml`), developers **MUST**  changes require two builds to test - one to build the lifecycle, then a second to use that updated lifecycle.  Code changes to `Mojo` classes within the existing `habushu` lifecycle work via normal builds without the need for a second pass.
 * `mvn clean install -Pdefault`: (ACTIVE BY DEFAULT - `-Pdefault` does not need to be specified) builds all modules.  Developers may use this profile to build and apply changes to existing `habushu-maven-plugin` `Mojo` classes
-
--- a/README.md
+++ b/README.md
@@ -542,9 +542,9 @@ Folder in which Python test files are located - should align with Poetry's proje
 
 Default: `${project.basedir}/tests`
 
-#### cacheBuildWheelFiles ####
+#### cacheWheels ####
 
-Enables or Disables the copying of poetry build wheel files into Poetry cache.
+Enables or Disables the copying of wheels into Poetry cache.
 
 Default: `false`
 
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
@@ -1,21 +1,30 @@
 package org.technologybrewery.habushu;
 
 import com.electronwill.nightconfig.core.Config;
+import com.electronwill.nightconfig.core.conversion.Path;
 import com.electronwill.nightconfig.core.file.FileConfig;
+
+import org.apache.commons.io.FileUtils;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.technologybrewery.habushu.exec.PoetryCommandHelper;
+import org.technologybrewery.habushu.util.HabushuUtil;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Delegates to Poetry during the {@link LifecyclePhase#PACKAGE} build phase to
@@ -36,6 +45,12 @@ public class BuildDeploymentArtifactsMojo extends AbstractHabushuMojo {
      */
     @Parameter(property = "habushu.exportRequirementsFile", required = false, defaultValue = "true")
     protected boolean exportRequirementsFile;
+
+    /**
+     * By default, do not cache wheel (*.whl) file.
+     */
+    @Parameter(property = "habushu.cacheBuildWheelFiles", required = false, defaultValue = "false")
+    protected boolean cacheBuildWheelFiles;
 
     /**
      * By default, do not include the --without-urls flag when exporting.
@@ -112,6 +127,10 @@ public class BuildDeploymentArtifactsMojo extends AbstractHabushuMojo {
 
             setUpPlaceholderFileAsMavenArtifact();
         }
+
+        if(cacheBuildWheelFiles){
+            cacheWheelFiles();
+        }
     }
 
     private void logLocalMonorepoCaveats() {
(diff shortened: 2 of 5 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -149,4 +168,25 @@ protected void setUpPlaceholderFileAsMavenArtifact() {
...
+            File poetryWheelCacheDirectory = new File(String.format("%s/cache/repositories/wheels/%s", poetryCacheDirectoryPath, project.getArtifactId()));
+
+            if(poetryWheelCacheDirectory.exists() || poetryWheelCacheDirectory.mkdirs()){
+                List<File> wheelFiles = Stream.of(wheelSourceDirectory.listFiles())
+                                              .filter(file -> file.getAbsolutePath().endsWith(".whl"))
+                                              .map(File::getAbsoluteFile)
+                                              .collect(Collectors.toList());
+                for(File file : wheelFiles){
+                    HabushuUtil.copyFile(file.getPath(), String.format("%s/%s", poetryWheelCacheDirectory.toPath().toString(), file.getName()));
+                    getLog().info(String.format("Cached the %s file", file.getName()));
```
cpointe-ibllanos, line 185: A: Do we always want the cached files printed or would this be a better case for getLog().debug(...)?

cpointe-ibllanos (commented):
Looks good to me, one small question  I'll let someone with more Habushu experience give the approval.

georgit (approved):
Looks good to me.

README.md
```diff
@@ -542,6 +542,12 @@ Folder in which Python test files are located - should align with Poetry's proje
 
 Default: `${project.basedir}/tests`
 
+#### cacheBuildWheelFiles ####
+
+Enables or Disables the copying of poetry build wheel files into Poetry cache.
```
epwilkins: A: I'd just say "wheels" instead of "poetry build wheel files" (same for the config name).  A bit more readable and definitely something you'd expect someone using python to understand easily.  It might also be nice to say why this option would be useful vs not.  This should make most poetry installs work out-of-the-box, even if they aren't using a dev/path dependency right?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -112,6 +127,10 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
 
             setUpPlaceholderFileAsMavenArtifact();
         }
+
+        if(cacheBuildWheelFiles){
+            cacheWheelFiles();
```
epwilkins: S: Traditionally this kind of action would happen during the `install` phase instead of the `package` phase. If we feel strongly it should stay in the package phase by default, we should call out it out in the docs so it doesn't trip people up. Otherwise, I'd just introduce a new Mojo whose sole responsibility is to install the wheel files into the poetry cache.

epwilkins: Possible to push this to a follow-on as well. "As a Habushu user, I want to be able my wheels installed to the local cache separately from the package step, so I have more control over the build lifecycle."

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -149,4 +168,25 @@ protected void setUpPlaceholderFileAsMavenArtifact() {
         project.getArtifact().setFile(mavenArtifactFile);
     }
 
-}
+    private void cacheWheelFiles() {
+        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        try{
+            File wheelSourceDirectory = new File(String.format("%s/dist", project.getBasedir()));
```
epwilkins: C: I believe the `dist` directory is configurable. We could pull the configuration parameter that currently exists on the Clean mojo up to abstract mojo and re-use it here.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
JeffreyRoss: image

This is a one liner that gets tacked onto the existing action's log output and only shows when the config is set to true. I think this follows the existing patterns within habushu.

README.md
JeffreyRoss: @ewilkins-csi this is a step in the direction to make wheels available without directory calculations. It doesn't actually exercise the config in modules that need the wheel. That's next. 

Should I express that future intend at this point?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
JeffreyRoss: There is a configuration in the clean mojo but it's purpose is to identify where the clean plugin will delete artifacts. Doesn't appear to define where the are placed? Can you confirm? 

image

epwilkins: The actual configuration happens as part of Maven's functionality I believe, but may need a little research.  If you look at the `<build>` configuration of the POM in habushu-mixology you'll see we set `dist` with the `<directory>` tag. I'm assuming if you changed that value it would change it to a different location.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -149,4 +168,25 @@ protected void setUpPlaceholderFileAsMavenArtifact() {
...
 
-}
+    private void cacheWheelFile() {
+        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        try{
+            File wheelSourceDirectory = new File(String.format("%s/dist", project.getBasedir()));
+            String poetryCacheDirectoryPath = poetryHelper.getPoetryCacheDirectoryPath();
+            File poetryWheelCacheDirectory = new File(String.format("%s/cache/repositories/wheels/%s", poetryCacheDirectoryPath, project.getArtifactId()));
+
+            if(poetryWheelCacheDirectory.exists() || poetryWheelCacheDirectory.mkdirs()){
```
epwilkins, line 178: Q: Do we need an else that prints an error or fails the build?  Not sure under what circumstances `mkdirs` would return false if every.

epwilkins: If it basically always throws an exception or returns true, it's worth putting that in a comment so future readers aren't confused.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -57,6 +58,16 @@ public Pair<Boolean, String> getIsPoetryInstalledAndVersion() {
         }
     }
 
+    /**
+     * Returns a {@link String} indicating the relative path to the poetry 
+     * to cache directory.
```
epwilkins: A:
```suggestion
     * cache directory. This is equivalent to {@code poetry config cache-dir}.
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -181,4 +182,20 @@ public static void giveFullFilePermissions(String filePath) {
 	}
     }
 
-}
+    /**
+     * Copies specified file into specified path.
+     * 
+     * @param filePath the path to the file to copy
+     * @param destinationFilePath the path to the file to copy
```
epwilkins: A:
```suggestion
     * @param destinationFilePath the path to where the new copy should be created
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -181,4 +182,20 @@ public static void giveFullFilePermissions(String filePath) {
...
+     */
+    public static void copyFile(String sourceFilePath, String destinationFilePath) {
+        try{
+            File sourceFile = new File(sourceFilePath);
+            File destinationFile = new File(destinationFilePath);
+            FileUtils.copyFile(sourceFile, destinationFile);
+        } catch(IOException ioe){
+            throw new HabushuException("Could not copy the "+ sourceFilePath +" file!", ioe);
```
epwilkins: A:
```suggestion
            throw new HabushuException("Could not copy the file ["+ sourceFilePath +"] to [" + destinationFilePath +"]!", ioe);
```

epwilkins (approved):

README.md
epwilkins: Ah yeah it's fine to wait to add that detail until we actually implement it.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
JeffreyRoss: Added comment.

PR #91, opened 2024-02-26
#91 Behave test results should output a test results in the Cucumber reports JSON format

### LINKED ISSUE #89: Behave test suite should output test results in cucumber report format to establish consistency between test reports and show Python code coverage
Habushu currently configures Behave as the Python test suite. It functions similarly to the Java Cucumber test suite in that it parses the Gherkin language, but outputs the test result in a differently structure JSON. The idea is that since Habushu is intended to be used within the Maven compiler, we should standardize reporting to match the Cucumber report style. This will also allow us to more easily determine the code coverage for Python tests.

cpointe-ibllanos:
Causes the test outputs for Behave tests to be outputted to ~/target/cucumber-reports/cucumber.json in a format compatible with Cucumber reports. This creates more consistency between the Java Cucumber and Python Behave test suites.

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
@@ -29,6 +29,7 @@ import java.util.List;
 public class BehaveBddTestMojo extends AbstractHabushuMojo {
 
     protected static final String BEHAVE_PACKAGE = "behave";
+    protected static final String BEHAVE_CUCUMBER_FORMATTER = "behave-cucumber-formatter";
 
     /**
      * Options that should be passed to the behave command. <b>NOTE:</b> If this
@@ -36,6 +37,20 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
      */
     @Parameter(property = "habushu.behaveOptions", required = false)
     protected String behaveOptions;
+
+    /**
+     * By default, format Behave test results in a JSON compatible with Cucumber Reports plugin
+     */
+    @Parameter(property = "habushu.outputCucumberStyleTestReports", required = false, defaultValue = "true")
+    protected boolean outputCucumberStyleTestReports;
+
+    /**
+     * By default, Behave marks Scenarios and Features with skipped steps as failures in its reports even if the
+     * other test steps themselves pass. To match Cucumber's default logic, this setting will prevent capturing skipped
+     * tests if there are no failures.
+     */
+    @Parameter(property = "habushu.dontFailBehaveTestsWithSkippedTestSteps", required = false, defaultValue = "true")
+    protected boolean dontFailBehaveTestsWithSkippedTestSteps;
 
 
     /**
@@ -82,6 +97,17 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
             List<String> executeBehaveTestArgs = new ArrayList<>();
             executeBehaveTestArgs
                     .addAll(Arrays.asList("run", BEHAVE_PACKAGE, getCanonicalPathForFile(behaveDirectory)));
+
+            if (outputCucumberStyleTestReports) {
+                poetryHelper.installDevelopmentDependency(BEHAVE_CUCUMBER_FORMATTER);
+                executeBehaveTestArgs.add("--format=behave_cucumber_formatter:PrettyCucumberJSONFormatter");
+                executeBehaveTestArgs.add("--outfile=dist/cucumber-reports/cucumber.json");
+            }
+
+            if (dontFailBehaveTestsWithSkippedTestSteps) {
+                executeBehaveTestArgs.add("--no-skipped");
+            }
+
 
             if (StringUtils.isNotEmpty(behaveOptions)) {
                 executeBehaveTestArgs.addAll(Arrays.asList(StringUtils.split(behaveOptions)));
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
@@ -49,8 +49,8 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
      * other test steps themselves pass. To match Cucumber's default logic, this setting will prevent capturing skipped
      * tests if there are no failures.
      */
-    @Parameter(property = "habushu.dontFailBehaveTestsWithSkippedTestSteps", required = false, defaultValue = "true")
-    protected boolean dontFailBehaveTestsWithSkippedTestSteps;
+    @Parameter(property = "habushu.omitSkippedTests", required = false, defaultValue = "true")
+    protected boolean omitSkippedTests;
 
 
     /**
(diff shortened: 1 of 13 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
```diff
@@ -72,6 +72,13 @@ public abstract class AbstractHabushuMojo extends AbstractMojo {
     @Parameter(property = "habushu.testDirectory", required = true, defaultValue = "${project.basedir}/tests")
     protected File testDirectory;
 
+    /**
+     * Folder in which Python build files are located - should align with Poetry's
+     * project structure conventions.
+     */
+    @Parameter(property = "habushu.targetDirectory", required = true, defaultValue = "${project.basedir}/target")
+    protected File targetDirectory;
```
cpointe-ibllanos: A: this should no longer be needed - was from attempting automated testing for this feature

habushu-maven-plugin/src/test/resources/specifications/dependency-management.feature
```diff
@@ -53,7 +53,7 @@ Feature: Test dependency management capabilities to help align package versions
       | packageFooTest | ^1.2.7             |
       | packageBarTest | ^0.9.0             |
 
-  Scenario: Inactive managed dependencies are skipped
+  Scenario Outline: Inactive managed dependencies are skipped
```
cpointe-ibllanos, line 56: Simple syntax fixes. Cucumber seems smart enough to interpret Scenarios as Outlines based on whether Examples are given either way.

habushu-maven-plugin/src/test/resources/junit-platform.properties
cpointe-ibllanos, line 1: I: Using the reports from the plugin module to make sure reporting works against both Java and Python with the new plugin management configurations.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -37,6 +38,20 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
+     */
+    @Parameter(property = "habushu.outputCucumberStyleTestReports", required = false, defaultValue = "true")
+    protected boolean outputCucumberStyleTestReports;
+
+    /**
+     * By default, Behave marks Scenarios and Features with skipped steps as failures in its reports even if the
+     * other test steps themselves pass. To match Cucumber's default logic, this setting will prevent capturing skipped
+     * tests if there are no failures.
+     */
+    @Parameter(property = "habushu.dontFailBehaveTestsWithSkippedTestSteps", required = false, defaultValue = "true")
```
cpointe-ibllanos: I: Considering renaming this "omitSkippedTests" or inversing the logic and going with "showSkippedTests" - curious what others think.

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/TestSpecifications.java
```diff
@@ -15,6 +13,8 @@
 @Suite
 @IncludeEngines("cucumber")
 @SelectClasspathResource("specifications")
+@ExcludeTags("manual")
+@ConfigurationParameter(key = FILTER_TAGS_PROPERTY_NAME, value = "not manual")
```
cpointe-ibllanos: I: Habushu uses the junit platform suite integrated with cucumber for its testing; I had to include these settings to omit @manual tests.

habushu-maven-plugin/src/test/resources/junit-platform.properties
```diff
@@ -1,2 +1,6 @@
 cucumber.publish.quiet=true
-cucumber.publish.enabled=false
\ No newline at end of file
+cucumber.publish.enabled=false
+cucumber.glue=org.technologybrewery.habushu
+cucumber.filter.tags=~@manual
```
cpointe-ibllanos: A: I may be able to define this as "not manual" and remove one of the changes above.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
d-ryan-ashcraft: Sounds good - sort of a mouthful with how it is now

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/TestSpecifications.java
cpointe-ibllanos: Was able to get rid of the second one

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/ReportFormattingSteps.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class ReportFormattingSteps {
+    @When("the Habushu project is built")
+    public void the_habushu_project_is_built() {
+        // Manual
+    }
+
+    @Then("a Cucumber report file {string} exists")
+    public void a_cucumber_report_file_exists(String string) {
+        // Manual
```
Cho-William: what is meant by just the manual comment? some more context in the comment would help with knowing why these steps are empty

cpointe-ibllanos: New comment:
        // Manual - test is dependent on files generated after Behave runs and cannot be automated

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -83,6 +98,17 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
             executeBehaveTestArgs
                     .addAll(Arrays.asList("run", BEHAVE_PACKAGE, getCanonicalPathForFile(behaveDirectory)));
 
+            if (outputCucumberStyleTestReports) {
+                poetryHelper.installDevelopmentDependency(BEHAVE_CUCUMBER_FORMATTER);
+                executeBehaveTestArgs.add("--format=behave_cucumber_formatter:PrettyCucumberJSONFormatter");
+                executeBehaveTestArgs.add("--outfile=dist/cucumber-reports/cucumber.json");
```
Cho-William, line 104: this installs a dependency that formats and creates the cucumber.json needed to create the final html report

habushu-mixology-consumer/pyproject.toml
```diff
@@ -11,6 +11,10 @@ python = "^3.11"
 [tool.poetry.group.monorepo.dependencies]
 habushu-mixology = {path = "../habushu-mixology", develop = true}
 
+
+[tool.poetry.group.dev.dependencies]
+behave-cucumber-formatter = "^1.0.1"
```
d-ryan-ashcraft: S: generally, we are trying to go with `>=` over `^` given that it plays more nicely general Python upgrade approaches (if there is a good reason to use `^` instead, that's fine - but recommend throwing a comment next to it)

habushu-mixology/pyproject.toml
```diff
@@ -17,6 +17,9 @@ behave = "^1.2.6"
 grpcio-tools = "^1.48.0"
 python-dotenv = "^0.20.0"
 
+[tool.poetry.group.dev.dependencies]
+behave-cucumber-formatter = "^1.0.1"
```
d-ryan-ashcraft: S: see prior comment

d-ryan-ashcraft (approved):

habushu-mixology-consumer/pyproject.toml
cpointe-ibllanos: You're right -- doing ^ would not fetch major and minor releases e.g. 2.0.0 or 1.1.0

Issue #94, opened 2024-03-04
Modify Retrieving of cached Wheels to allow non-habushu packaging


JeffreyRoss:
Currently the RetrieveWheels functionality requires that the executing packaging is Habushu, however, there are some use cases where we'd want to use these functionality in non-habushu packaging.

PR #95, opened 2024-03-05
#94 Modifying retrieve caching logic to allow for non-habushu packaging

JeffreyRoss:
Currently the RetrieveWheels functionality requires that the executing packaging is Habushu, however, there are some use cases where we'd want to use these functionality in non-habushu packaging.

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RetrieveWheelsMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RetrieveWheelsMojo.java
@@ -32,6 +32,9 @@ public class RetrieveWheelsMojo extends AbstractHabushuMojo {
     @Parameter(property = "habushu.wheelDependencies", required = false)
     protected List<WheelDependency> wheelDependencies;
 
+    /**
+    * Overriding to allow execution in non-habushu projects
+    */      
     @Override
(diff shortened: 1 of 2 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RetrieveWheelsMojo.java
```diff
@@ -32,6 +32,11 @@ public class RetrieveWheelsMojo extends AbstractHabushuMojo {
     @Parameter(property = "habushu.wheelDependencies", required = false)
     protected List<WheelDependency> wheelDependencies;
 
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException {
```
cpointe-ibllanos, line 39: A: Probably worth clarifying the logic being skipped in a comment.

cpointe-ibllanos (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RetrieveWheelsMojo.java
JeffreyRoss: @cpointe-ibllanos added.

epwilkins (approved):

PR #97, opened 2024-03-08
Change output directory for Behave tests

### LINKED ISSUE #96: Habushu should output Behave test to the Python module's target directory, not dist
The distribution directory of a Python module, like the name implies, is meant to be packaged and shared. Since we only care about the reports in the test scope, Behave tests and reports should be output to the target directory just like Cucumber test.

cpointe-ibllanos:
Configures Behave test JSONs and Cucumber-style reports to generate in the target directory rather than dist for consistent behavior across both testing frameworks.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,8 +88,8 @@
 								<checkBuildResult>true</checkBuildResult>
 								<treatUndefinedAsFailed>true</treatUndefinedAsFailed>
 								<treatSkippedAsFailed>false</treatSkippedAsFailed>
-								<outputDirectory>${project.build.directory}/cucumber-reports</outputDirectory>
-								<inputDirectory>${project.build.directory}/cucumber-reports</inputDirectory>
+								<outputDirectory>${project.basedir}/target/cucumber-reports</outputDirectory>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -88,8 +88,8 @@
 								<checkBuildResult>true</checkBuildResult>
 								<treatUndefinedAsFailed>true</treatUndefinedAsFailed>
 								<treatSkippedAsFailed>false</treatSkippedAsFailed>
-								<outputDirectory>${project.build.directory}/cucumber-reports</outputDirectory>
-								<inputDirectory>${project.build.directory}/cucumber-reports</inputDirectory>
+								<outputDirectory>${project.basedir}/target/cucumber-reports</outputDirectory>
+								<inputDirectory>${project.basedir}/target/cucumber-reports</inputDirectory>
```
cpointe-ibllanos, line 92: I: The build directory for Python projects is set to the distribution folder; this changing this allows us to target both Cucumber and Behave tests without having to do separate path configurations in the reporting plugin for each testing framework.

jacksondelametter:
Looks good to me, very simple!

JeffreyRoss (approved):

PR #102, opened 2024-03-13
Reverse monorepo migration

### LINKED ISSUE #72: Remove `tool.poetry.group.monorepo.dependencies` to facilitate proper transitive installation; keep semantics of export at the same time
As a side effect of moving monorepo dependencies to their own `[tool.poetry.group.monorepo.dependencies]` group, they no longer get installed by default.  It would be useful if they did get installed by default, but continued to not be exported by default.

### Definition of Done:
* Remove migration to auto-include `path`-based dependencies int the `[tool.poetry.group.monorepo.dependencies]` group
* Reverse migrate any `[tool.poetry.group.monorepo.dependencies]` back into the standard `[tool.poetry.dependencies]` group 
* Remove export updates related to #56 add `[tool.poetry.group.monorepo.dependencies]` group and restrict export of `path` dependencies
* Use functionality in `poetry-monorepo-plugin` to facilitate this instead: https://github.com/TechnologyBrewery/poetry-monorepo-dependency-plugin/issues/17

cpointe-ibllanos:
https://github.com/TechnologyBrewery/habushu/issues/72


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/RemoveMonorepoGroupMigration.java
```diff
@@ -26,45 +26,40 @@
...
-                Config foundDependencies = toolPoetryDependencies.get();
+            Optional<Config> toolPoetryMonorepoDependencies = tomlFileConfig.getOptional(TomlUtils.TOOL_POETRY_GROUP_MONOREPO_DEPENDENCIES);
+            if (toolPoetryMonorepoDependencies.isPresent()) {
+                Config foundDependencies = toolPoetryMonorepoDependencies.get();
                 Map<String, Object> dependencyMap = foundDependencies.valueMap();
 
                 for (Map.Entry<String, Object> dependency : dependencyMap.entrySet()) {
                     String packageName = dependency.getKey();
                     Object packageRhs = dependency.getValue();
                     if (TomlUtils.representsLocalDevelopmentVersion(packageRhs)) {
```
epwilkins: S: Since we're removing the monorepo group entirely, I don't think we care to check whether it's a path dependency here -- we just want to move _all_ dependencies in that group. (In practice, it should only be path deps)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/RemoveMonorepoGroupMigration.java
```diff
@@ -26,45 +26,40 @@
...
                         String packageRshAsString = TomlUtils.convertCommentedConfigToToml((CommentedConfig) packageRhs);
-                        logger.info("Found local dependency not within monorepo group! ({} = {})", packageName, packageRshAsString);
+                        logger.info("Found local dependency within monorepo group! ({} = {})", packageName, packageRshAsString);
                         TomlReplacementTuple replacementTuple = new TomlReplacementTuple(packageName, packageRshAsString, "");
                         replacements.put(packageName, replacementTuple);
+                        Optional<Config> toolPoetryDependecies = tomlFileConfig.getOptional(TomlUtils.TOOL_POETRY_DEPENDENCIES);
+                        hasExistingDependenciesGroup = toolPoetryDependecies.isPresent();
                         shouldExecute = true;
```
epwilkins: A: The above 3 lines can be moved outside of the for loop

epwilkins (approved):

PR #103, opened 2024-03-14
Configure Habushu to use latest monorepo dependency and omit pathed dependencies from requirements.txt by default

cpointe-ibllanos:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -63,6 +63,9 @@ public class BuildDeploymentArtifactsMojo extends AbstractHabushuMojo {
     @Parameter(property = "habushu.exportRequirementsWithHashes", required = false, defaultValue = "true")
     protected boolean exportRequirementsWithHashes;
 
+    @Parameter(property = "habushu.exportRequirementsWithoutPathedDependencies", required = false, defaultValue = "true")
```
epwilkins: A: Poetry calls them "path dependencies"
```suggestion
    @Parameter(property = "habushu.exportRequirementsWithoutPathDependencies", required = false, defaultValue = "true")
```

epwilkins (approved):

Issue #101, opened 2024-03-12
Configure Habushu to have print/logger statements visible by default


colinpalmer-pro:
Currently, one must add the following configuration to see print/logger statements in one's python code that is built/packaged by Habushu.

```
<plugin>
   <groupId>org.technologybrewery.habushu</groupId>
   <artifactId>habushu-maven-plugin</artifactId>
   <configuration>
      <behaveOptions>--no-capture --no-capture-stderr</behaveOptions>
   </configuration>
</plugin> 
```
This should be the default behavior. Suppressing logs can be an optional configuration.

Tech Version: 2.13.0

colinpalmer-pro:
PR merged and default logging functionality added. Issue resolved.

PR #104, opened 2024-03-20
#101 Habushu Default Logging

colinpalmer-pro:
- Added new disableOutputCapture parameter that will print behave stdout, stderr and logs to console by default
- Updated stderr to be an OutputStream so logs are visible when tests pass and they appear during the run instead of after

Example Output:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,9 +283,18 @@ Default: `true`
 
 Options that should be passed to the `behave` command when executing tests. If this value is provided, then **behaveExcludeManualTag** is ignored. 
 
-`behave` supports a [number of command line options](https://behave.readthedocs.io/en/stable/behave.html#command-line-arguments) - developers may adjust the default test execution behavior to optimize productivity, such as selectively executing features associated with a specific in-flight tag (`mvn clean test -Dhabushu.behaveOptions="--tags wip-feature"`) or changing logging behavior (`mvn clean test -Dhabushu.behaveOptions="--no-logcapture --no-capture"`).
+`behave` supports a [number of command line options](https://behave.readthedocs.io/en/stable/behave.html#command-line-arguments) - developers may adjust the default test execution behavior to optimize productivity, such as selectively executing features associated with a specific in-flight tag (`mvn clean test -Dhabushu.behaveOptions="--tags wip-feature"`).
 
 Default: None
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -646,6 +655,23 @@ Controls where the clean plugin will delete target artifacts.
 
 Default: `${project.basedir}/target`
```
colinpalmer-pro, line 657: Updated documentation from @cpointe-ibllanos

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/ProcessExecutor.java
```diff
@@ -187,7 +190,15 @@ public final void flush() {
```
colinpalmer-pro, line 190: Before we were throwing all the logs out on a failure in a logger.error message. This is an attempt to output the Python logs with the correct logger level. Open to discussions about a more robust solution.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/ProcessExecutor.java
```diff
@@ -187,7 +190,15 @@ public final void flush() {
 
         @Override
         protected void processLine(final String line, final int logLevel) {
-            logger.info(line);
+            if (line.toUpperCase().contains(ERROR_LOG) || line.toUpperCase().contains(CRITICAL_LOG)) {
```
epwilkins, line 193: A: Not too much of a difference, but I think `startsWith` over `contains` makes this logic less susceptible to edge cases.

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/ProcessExecutor.java
colinpalmer-pro: Yea actually thats what I started with but I noticed that some of the aiSSEMBLE logs have the following format: `2024/03/20 13:06:51 INFO Environment: Starting Test container services`, where the level isn't listed at the beginning. So I switched to `contains`

aaron-gary (approved):

PR #109, opened 2024-04-03
HAB-108 Behave Bdd Environment Variables

### LINKED ISSUE #108: Add ability to set environment variables for behave bdd tests
Add functionality to the habushu-maven-plugin to accept environment variables that are recognized during the behave bdd tests. Will be similar to environment variables in the maven-failsafe-plugin and the maven-surefire-plugin.

colinpalmer-pro:
Added behaveTestEnvironmentVariables parameter to be able to pass in environment variables for behave tests


README.md
```diff
@@ -304,6 +304,40 @@ Exclude any BDD scenario or feature file tagged with `@manual`.
...
+  <groupId>org.technologybrewery.habushu</groupId>
+  <artifactId>habushu-maven-plugin</artifactId>
+  <extensions>true</extensions>
+  <configuration>
+    <behaveTestEnvironmentVariables>
+      <ENV_VAR>VALUE</ENV_VAR>
+    </behaveTestEnvironmentVariables>
+  </configuration>
+</plugin>
+```
```
colinpalmer-pro, line 322: I: Added additional instructions for adding environment variables through the cli because it differs slightly than other configuration options. From my experience with this, if you want to be able to pass environment variables as `-Dhabushu...` without any `pom.xml` configuration, the `behaveTestEnvironmentVariables` parameter would need to be a string. We could then parse the string and convert to a map but entries in the pom (not using the cli) would need to follow the format: 
```
<behaveTestEnvironmentVariables>ENV_VAR1=value1;ENV_VAR2=value2</behaveTestEnvironmentVariables>
```

Since envrionment variables are usually set in the pom and follow the format:
```
<behaveTestEnvironmentVariables>
      <ENV_VAR>VALUE</ENV_VAR>
      <ENV_VAR2>VALUE2</ENV_VAR2>
</behaveTestEnvironmentVariables>
```
I chose to go with `behaveTestEnvironmentVariables` as a map (instead of a string to parse) and you still have the ability to pass in environment variables with `-D` option if you set up the pom correctly.

Issues/comments/concerns here let me know.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -133,6 +133,26 @@ public int executeAndLogOutput(List<String> arguments) throws MojoExecutionExcep
         return executor.executeAndRedirectOutput(logger);
     }
```
colinpalmer-pro, line 135: Overloaded `executeAndLogOutput()` to include `Map<String, String> environmentVariables` parameter

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -214,4 +234,11 @@ protected ProcessExecutor createPoetryExecutor(List<String> arguments) {
         fullCommandArgs.addAll(arguments);
         return new ProcessExecutor(workingDirectory, fullCommandArgs, Platform.guess(), null);
     }
+
```
colinpalmer-pro, line 235: Overloaded `createPoetryExecutor()` to include `Map<String, String> environmentVariables` parameter

d-ryan-ashcraft (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -74,6 +75,12 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
     @Parameter(property = "habushu.disableOutputCapture", defaultValue = "true")
     protected boolean disableOutputCapture;
 
+    /**
+     * Environment variables utilized in the behave bdd tests.
+     */
+    @Parameter(property = "habushu.behaveTestEnvironmentVariables")
+    protected Map<String, String> behaveTestEnvironmentVariables = null;
+
```
cpointe-ibllanos, line 83: S: Should update the documentation with this new config and a description of it.

README.md
cpointe-ibllanos: https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Mapping_Collection_Types

I think you went with the right approach -  this is the standard format for maps in poms and there's no inherent comma-separated string parsing like there is for lists.

cpointe-ibllanos (approved):

Issue #110, opened 2024-04-04
Add GitHub Action to automate Habushu release and OSSRH publishing activities


ekoniec1:
Create the appropriately restricted GitHub Action to automate publishing and releasing Habushu artifacts to Maven Central/OSSRH

colinpalmer-pro:
Can now run a Habushu Release with the "Release Habushu" Action. Closing issue.

PR #111, opened 2024-04-05
HAB-#110 Github Action of Habushu Release and OSSRH publishing - Adde…

colinpalmer-pro:
Added release.yaml to create new GitHub action for habushu release. The action can be kicked off manually or from GitHub CLI and takes two inputs (release version and next development version) and runs the release-habushu.sh script

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,50 @@
+# This workflow kicks off the release-habushu.sh script which creates a new version of habushu
+# and deploys artifacts to Maven Central.
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Release Habushu
+
+on:
+    workflow_dispatch:
+        inputs:
+            releaseVersion:
+                description: "Release version"
+                required: true
+            developmentVersion:
+                description: "Next development version"
+                required: true
+
+jobs:
+    release:
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Check out code
+              uses: actions/checkout@v3
+            - name: Setup Pyenv
+              uses: gabrielfalcao/pyenv-action@v16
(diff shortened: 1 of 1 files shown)


kdendtler (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,50 @@
...
+            - name: Set up JDK 11
+              uses: actions/setup-java@v3
+              with:
+                  java-version: '11'
+                  distribution: 'temurin'
+                  cache: maven
+                  server-id: 'ossrh'
+                  server-username: MAVEN_USERNAME
+                  server-password: MAVEN_PASSWORD
+                  gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
```
d-ryan-ashcraft, line 41: A: is there a reason this one is directly wired and the others are indirected as ENV variables?

d-ryan-ashcraft (commented):
How is it restricted so only maintainers can run it?

.github/workflows/release.yaml
colinpalmer-pro: I don't have a great answer here except that I'm following the format I saw the most and from some posts I read. In the GitHub docs for [Publishing Packages to the Maven Central Repository](https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven#publishing-packages-to-the-maven-central-repository), they utilize environment variables for server-username and server-password. 

And according to this example https://www.linkedin.com/pulse/uploading-artifacts-maven-central-github-actions-itemis, in the "Explanation" section he mentions that the `settings.xml` is built up from references to environment variables. So we need server-name, server-password and the gpg-passphrase as environment variables to populate the `settings.xml`. But we can just pass in the gpg-private-key from the secret because it's not utilized in the `settings.xml`

colinpalmer-pro:
> How is it restricted so only maintainers can run it?

I think default behavior is anyone with write access can run it. Let me see if there is a way to update to maintainer

PR #112, opened 2024-04-08
HAB-release-git-config-issue-#110

colinpalmer-pro:
- Updated scm to use https (instead of ssh) and authenticate with `GITHUB_TOKEN`
- Added Generic GitHub config information

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -42,9 +42,12 @@ jobs:
                   gpg-passphrase: GPG_PASSPHRASE
             - name: Release Habushu
               run: |
+                  git config --local user.email "habushu-noreply@technologybrewery.org"
+                  git config --local user.name "Github Actions"
                   chmod +x release-habushu.sh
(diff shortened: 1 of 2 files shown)


JeffreyRoss (approved):

.github/workflows/release.yaml
```diff
@@ -42,9 +42,12 @@ jobs:
                   gpg-passphrase: GPG_PASSPHRASE
             - name: Release Habushu
               run: |
+                  git config --local user.email "habushu-noreply@technologybrewery.org"
+                  git config --local user.name "Github Actions"
```
JeffreyRoss, line 46: A: I'm curious if these config settings like this will persist past the execution? _(if that makes sense)_

colinpalmer-pro, line 46: The settings are supposed to be ephemeral. I can see the "Post Checkout Code" step it runs "/usr/bin/git config --local --unset-all" so it should wipe them out.

PR #122, opened 2024-04-23
#119 Point Behave test suite to new formatter, kappa-maki, to address malgenerated test reports

### LINKED ISSUE #121: Behave generating malformed test reports when all tests skipped
When any Behave test contains no features or steps (presumably because they were all skipped due to a @manual tag), Behave's formatter will attempt to close the report with the character ']'. This causes an exception in the report generation plugin. The file needs to be empty so they are skipped when the reporting plugin aggregates all the JSON test results to generate a HTML report.

cpointe-ibllanos:


habushu-mixology-consumer/tests/features/generate-report-with-manual-tests.feature
cpointe-ibllanos, line 1: Introduced this test to confirm there are no issues formatting the test results when some features or steps are skipped.

habushu-mixology-consumer/tests/features/generate-report-with-manual-tests.feature
```diff
@@ -0,0 +1,9 @@
+@manual
```
jacksondelametter, line 1: Could we not automate this test? if all were doing is checking to see if an HTML report is in the target directory, this should be fairly easy to do.

d-ryan-ashcraft: I think the point is to have a test that doesn't generate anything to simulate missing tests in the wild?

d-ryan-ashcraft (approved):

habushu-mixology-consumer/tests/features/generate-report-with-manual-tests.feature
cpointe-ibllanos: @d-ryan-ashcraft is correct. It's a bit of a weird setup, but it's effectively a means of reproducing the issue. That said, since this is specifically testing the issue, I can see the argument for removing it.

jacksondelametter (approved):

Issue #124, opened 2024-05-02
Support Python linting to capture any issues and errors


liangyun123:
Introduce the functionality to flag issues within Python code so that users can address concerns before the Python package is built and released. This will allow projects to ensure deployed code is up to certain standards and capture any errors that unit testing do not address for.

epwilkins:
CI Build passed
![Screenshot 2024-05-07 at 4 18 29 PM](https://github.com/TechnologyBrewery/habushu/assets/80470879/5f9582f7-7d64-4aec-96fa-921851194f05)

PR #125, opened 2024-05-03
#124 Add support for Python linting to check for issues before packaging

liangyun123:


README.md
```diff
@@ -703,10 +703,28 @@ Default: `true`
...
 
-Controls whether skipped tests should be completely omitted from test reports rather than showing up as a skip / failure. This mimics the default behavior of Cucumber and will have no effect if outputCuucumberStyleTestReports is not set to `true`
+Controls whether skipped tests should be completely omitted from test reports rather than showing up as a skip / failure. This mimics the default behavior of Cucumber and will have no effect if outputCucumberStyleTestReports is not set to `true`
 
 Default: `true`
 
+#### pylintDisabledChecker ####
+
+Controls the disabled checkers for Pylint. 
+
```
d-ryan-ashcraft, line 796: S (Substantive): what do these letters mean and why don't we make them more user friendly?

d-ryan-ashcraft: Perhaps b/c of these options - but why not just have this be `lintEnabled=true/false`?

README.md
```diff
@@ -703,10 +703,28 @@ Default: `true`
...
+Default: `C,R,W`
+
+#### pylintEnabledChecker ####
+
+Controls the enabled checkers for pylint. 
+
+Default: `E,F`
+
+#### pylintExitZero ####
```
d-ryan-ashcraft: S: why not call this something easier to understand like `failOnLintErrors`?

README.md
```diff
@@ -703,10 +703,28 @@ Default: `true`
 
 #### omitSkippedTests ####
 
-Controls whether skipped tests should be completely omitted from test reports rather than showing up as a skip / failure. This mimics the default behavior of Cucumber and will have no effect if outputCuucumberStyleTestReports is not set to `true`
+Controls whether skipped tests should be completely omitted from test reports rather than showing up as a skip / failure. This mimics the default behavior of Cucumber and will have no effect if outputCucumberStyleTestReports is not set to `true`
 
 Default: `true`
 
+#### pylintDisabledChecker ####
```
d-ryan-ashcraft: S (Substantive): would be better if we took the underlying library name out and just called it linting.  I realize we've made this mistake elsewhere, but would prefer not to add on more to that now.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
```diff
@@ -0,0 +1,80 @@
...
+import org.technologybrewery.habushu.exec.PoetryCommandHelper;
+
+/**
+ * Leverages the pylint package to validate both source and test Python
+ * directories using Poetry's run command.
+ */
+@Mojo(name = "validate-python-code", defaultPhase = LifecyclePhase.PROCESS_CLASSES, requiresDependencyResolution = ResolutionScope.COMPILE)
+public class ValidatePythonCodeMojo extends AbstractHabushuMojo {
+
+    protected static final String PYLINT_PACKAGE = "pylint";
```
d-ryan-ashcraft: S: s/b "LINT_PACKAGE" so we could, in theory change it more easily.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
```diff
@@ -0,0 +1,80 @@
...
+     * build will continue. 
+     */
+    @Parameter(property = "habushu.pylintExitZero", required = false, defaultValue = "false")
+    private boolean pylintExitZero;
+
+    @Override
+    public void doExecute() throws MojoExecutionException {
+        List<String> directoriesToValidate = new ArrayList<>();
+        if (this.sourceDirectory.exists()) {
+            directoriesToValidate.add(getCanonicalPathForFile(sourceDirectory));
```
d-ryan-ashcraft: S: will this cover test clases?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
liangyun123: Addressed, thanks Ryan for the suggestion!

README.md
liangyun123: Addressed, thanks Ryan for the suggestion!

README.md
liangyun123: The pylintDisabledChecker (now referenced as lintDisabledChecker) is used to configure the checkers to disable on pylint. Likewise, the pylineEnabledChcker (now referenced as lintEnabledChecker) is used to configure the checkers to enable on pylint (checkers referring to https://pylint.readthedocs.io/en/latest/user_guide/checkers/features.html). Hence, both configurations are not used for enabling/disabling the linting package, but to configure which checkers to enable/disable for linting. 

But to your first comment, yup, I can add a small description to the default values, as the characters are not obvious unless reader has used Pylint before.

README.md
liangyun123: Changed, thanks Ryan for the suggestion!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
liangyun123: I'm opting out linting on the test classes because the tests are already executed by Python, hence Python would error out when the behave tests are executed.

d-ryan-ashcraft: But wouldn't this possibly highlight a bad test?

liangyun123: Yup, it would, lint would be able to identify bad code before tests are executed by Python. My only thoughts are that linting on test modules should have separate configurations from the source modules, so the Habushu configurations will need to be expanded. Expanding on this thought: there will be different requirements for checks in the source and test modules. For example, one of the errors I found when linting on tests is the overwriting function error, since a common function name for test steps is `step_impl`, but lint assumes you are overwriting the function with each step. Hence, we will need to disable the checker for overwriting function in the test or rename tests to be unique, but we do not want to disable this on the source module. 

But ultimately I think it's reasonable to lint on both source and test modules to give an earlier warning to users before the tests are executed.

d-ryan-ashcraft: I'm pretty squarely in the camp that test code is first class code and should be treated just like source code.  Perhaps we should challenge some of the assumptions around overwriting et, al.  We might be able to find some useful tweaks that allow the same linting while also meeting our intent.

README.md
```diff
@@ -703,10 +703,42 @@ Default: `true`
 
 #### omitSkippedTests ####
 
-Controls whether skipped tests should be completely omitted from test reports rather than showing up as a skip / failure. This mimics the default behavior of Cucumber and will have no effect if outputCuucumberStyleTestReports is not set to `true`
+Controls whether skipped tests should be completely omitted from test reports rather than showing up as a skip / failure. This mimics the default behavior of Cucumber and will have no effect if outputCucumberStyleTestReports is not set to `true`
```
liangyun123, line 783: Small typo I noticed when updating README.md.

habushu-mixology-consumer/pom.xml
```diff
@@ -39,6 +39,7 @@
                             <operatorAndVersion>${project.version}</operatorAndVersion>
                         </packageDefinition>
                     </managedDependencies>
+                    <testLintDisabledChecker>C,R,W,E0102</testLintDisabledChecker>
```
liangyun123: E0102 is Pylint's code for redefined-function error. Since the test steps' functions are named `step_impl`, pylint assumes that the functions are being redefined. Hence disabling this checker so that build continues.

d-ryan-ashcraft: I think this comment seems targeted for a specific project - recommend noting the general beenfit of this vice a project-specific use case (Seems like this could still be useful outside whatever use can you are thinking of specifically).

d-ryan-ashcraft (approved):

README.md
epwilkins: We probably do want an on/off control for linting in general

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.technologybrewery.habushu.exec.PoetryCommandHelper;
+
+/**
+ * Leverages the lint package to validate both source and test Python
+ * directories using Poetry's run command.
+ */
+@Mojo(name = "validate-python-code", defaultPhase = LifecyclePhase.PROCESS_CLASSES, requiresDependencyResolution = ResolutionScope.COMPILE)
```
epwilkins: A: Might make more sense for this to be in the `process-sources` phase by default. I'm thinking if we add python compilation, that would be attached to the `compile` phase which happens _before_ process classes.  Logically it makes more sense to lint before you compile.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
```diff
@@ -0,0 +1,135 @@
...
+    protected static final String LINT_PACKAGE = "pylint";
+
+    /**
+     * Specifies disabled checkers for lint on source module.
+     */
+    @Parameter(property = "habushu.sourceLintDisabledChecker", required = false, defaultValue = "C,R,W")
+    private String sourceLintDisabledChecker;
+
+    /**
+     * Specifies disabled checkers for lint on test module.
```
epwilkins: S: I think we should split this into two mojos, one for the source code during `process-sources`/`process-classes` and one for the test code during `process-test-sources`/`process-test-classes`.  Since the code will have a ton of overlap we can just have an abstract that houses the majority of the logic, similar to how the jar plugin does jar vs test-jar.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
```diff
@@ -0,0 +1,135 @@
...
+    public void doExecute() throws MojoExecutionException {
+        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+
+        List<String> sourceDirectoriesToValidate = new ArrayList<>();
+        if (this.sourceDirectory.exists()) {
+            sourceDirectoriesToValidate.add(getCanonicalPathForFile(sourceDirectory));
+        }
+
+        if (sourceDirectoriesToValidate.isEmpty()) {
+            getLog().warn(String.format("Configured source (%s) directories does not exist - skipping...",
```
epwilkins: A:
```suggestion
            getLog().warn(String.format("Configured source (%s) directory does not exist - skipping...",
```

epwilkins: Also why would this not just be in the `else` branch of the if check above? Seems roundabout to check if we added it to a list to know whether the `exists` check was true or false.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
```diff
@@ -0,0 +1,135 @@
...
+        if (StringUtils.isNotEmpty(sourceLintDisabledChecker)) {
+            executeLintArgsSource.addAll(Arrays.asList("--disable", sourceLintDisabledChecker));
+        }
+
+        if (StringUtils.isNotEmpty(sourceLintEnabledChecker)) {
+            executeLintArgsSource.addAll(Arrays.asList("--enable", sourceLintEnabledChecker));
+        }
+
+        if (sourceFailOnLintErrors) {
+            executeLintArgsSource.add("--exit-zero");
```
epwilkins: C: Is this not backwards?  If I want to fail on errors, then I'll want a non-zero exit code in the case of errors.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
```diff
@@ -0,0 +1,135 @@
...
+
+
+
+    @Override
+    public void doExecute() throws MojoExecutionException {
+        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+
+        List<String> sourceDirectoriesToValidate = new ArrayList<>();
+        if (this.sourceDirectory.exists()) {
+            sourceDirectoriesToValidate.add(getCanonicalPathForFile(sourceDirectory));
```
epwilkins: S: Two questions: 1) what's the benefit of adding this to a list if we only ever add one item, and 2) if the source directory does _not_ exist, doesn't this code still run the poetry pylint command but with an invalid configuration (i.e. no directory to lint)?  Even if the latter is untrue and pylint makes sense of not having a directory, it seems to clash with the log statement further down that claims we're skipping linting because the source directory doesn't exist.

habushu-maven-plugin/src/main/resources/META-INF/plexus/components.xml
```diff
@@ -22,7 +22,10 @@
                                 org.technologybrewery.habushu:habushu-maven-plugin:${project.version}:retrieve-wheels
                             </validate>							
                             <compile>org.technologybrewery.habushu:habushu-maven-plugin:${project.version}:install-dependencies</compile>
-                            <process-classes>org.technologybrewery.habushu:habushu-maven-plugin:${project.version}:format-python</process-classes>
+                            <process-classes>
+                                org.technologybrewery.habushu:habushu-maven-plugin:${project.version}:format-python,
+                                org.technologybrewery.habushu:habushu-maven-plugin:${project.version}:validate-python-code
```
epwilkins: So now that I'm further along I see that maybe my comment previously about moving the phase doesn't make sense.  The primary concern about if we add python code compilation looks like it's already a concern, since the compile phase currently installs dependencies and formatting happens _after_ the compile phase.  So I think what you have is consistent with the current state.

habushu-mixology-consumer/pom.xml
epwilkins: This looks like it's specifically referring to the `step_impl` glue code for the Behave BDD scenarios of Mixology.  But since Habushu pushes Behave as the preferred testing framework, and Behave encourages the use of a redefined `step_impl` function, we should definitely make this the _default_ configuration.

README.md
liangyun123: Added the configurations: lintSource and lintTest, that controls linting for the source and test modules respectively. Thank you for the suggestion.

habushu-mixology-consumer/pom.xml
liangyun123: Reconfigured the default disabled checkers for test module to include E0102. Thank you for the suggestion!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
liangyun123: This pattern was recognized from the FormatPythonMojo.java. I've made the changes to clean up the code, but we should flow another PR to address FormatPythonMojo as well.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
liangyun123: Yup, it's backward, great catch! Variable's name was slightly altered but code logic was never addressed. Thank you for the suggestion!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonCodeMojo.java
liangyun123: Similar to the roundabout code above, this pattern was recognized from FormatPythonMojo.java. To address your questions 1) There are no benefits, I've reformatted the code to not use a list, 2) added conditionals to make sure when source/test modules do not exist, pylint does not run. I've cleaned up the code for this, but I think a follow-on PR is required to address this pattern in FormatPythonMojo.java. Thanks for the suggestion.

epwilkins (approved):

PR #133, opened 2024-05-10
#119 :zap: :green_heart: improve CI performance

d-ryan-ashcraft:


.github/workflows/maven.yaml
```diff
@@ -20,22 +20,35 @@ jobs:
...
 
         steps:
-            - uses: actions/checkout@v3
-            - name: Setup Pyenv
-              uses: gabrielfalcao/pyenv-action@v16
+            - uses: actions/checkout@v4
+            - name: Install Python # use direct install rather than pyenv for CI for large speed improvement
+              uses: actions/setup-python@v5
+              with:
+                python-version: '3.11.4'
```
JeffreyRoss, line 27: Q: Is it typical to hardcode the version in this way?

JeffreyRoss (approved):

.github/workflows/maven.yaml
d-ryan-ashcraft: Good question.  Typically, we have it in Maven as a property and let pyenv install it.  While great and fast when using previously used versions locally or switching between local versions, it's very slow on CI.  So here, we are making the trade off of having to specify it a second time to save ~2m per build b/c this is a much more efficient GitHub Action installation process (as we get a new runner each build, so there isn't a good way to store a prior version).

Issue #140, opened 2024-05-16
Update habushu to use `virtualenvs.in-project` by default


d-ryan-ashcraft:
## Overview

When debugging Poetry-based Python projects, you need to select a Python Interpreter to get the full value of the debugger.  This requires chasing down the virtual environment for the module in question.  With the structure of the default Poetry cache, this can be a bit more involved than would be optimal.  

For example, on MacBooks, you can find a specific virtual environments in a path like the following:
```
/Users/ichiro/Library/Caches/pypoetry/virtualenvs/habushu-mixology-consumer-V-CVsbBP-py3.11
```

At times, there can be multiple virtual environments per projects, as well, adding another step for a developer to take to figure out which one is correct.  

Switching to `virtualenvs.in-project` as the default choice would make this process much easier by placing the virtual environment directly in the project's root under the `.venv` folder.  This would greatly simplify finding it folder and likely would eliminate the situation where we have duplicate virtual environments (not currently clear why that is happening, but have seen it personally and heard from others about it too).  

## Definition of Done
- [x] Add a `habushu.useInProjectVirtualEnvironment` flag and default it to `true`
- [x]  When set to `true`, it would trigger the following sequence of events:
   * During `clean`, if the virtual environment returned by Poetry does not match the local `.venv` folder, it will delete the alternative so `.venv` can become the new virtual environment (Poetry will not create a virtual environment "in project" is an exisitn virtual environment exists elsewhere
   * Set `virtualenvs.in-project` for the project in the project's `poetry.toml` file
- [x]  Update documentation for new flag and behavior noted above

## Test Steps
- Build the latest version of the `habushu-maven-plugin`:
```
mvn clean install -Pbootstrap
```
- Add the configuration following configuration to `habushu-mixology/pom.xml` to ensure your first build of the `habushu-mixology` uses the traditional Poetry Cache directory approach by adding: 
```
<plugin>
    <groupId>org.technologybrewery.habushu</groupId>
    <artifactId>habushu-maven-plugin</artifactId>
    <extensions>true</extensions>
    <configuration>
        ...
       <!-- ADD THE FOLLOWING LINE: -->        
       <useInProjectVirtualEnvironment>false</useInProjectVirtualEnvironment>
    </configuration>   
</plugin>
```
- Run `mvn clean compile -pl :habushu-mixology -Dmaven.build.cache.skipCache=true`
- Update `useInProjectVirtualEnvironment` to be `true`
[...]

PR #141, opened 2024-05-17
#140 :sparkles: support in project virtual environments

d-ryan-ashcraft:


habushu-maven-plugin/src/main/resources/META-INF/plexus/components.xml
```diff
@@ -14,8 +14,8 @@
                         <id>default</id>
                         <phases>
                             <initialize>
```
d-ryan-ashcraft, line 16: I: just cleaned up the order here as we should make sure poetry is set up before initializing it

.mvn/extensions.xml
```diff
@@ -2,6 +2,6 @@
     <extension>
         <groupId>org.apache.maven.extensions</groupId>
         <artifactId>maven-build-cache-extension</artifactId>
```
d-ryan-ashcraft, line 4: I (Informational): just bringing this up to date

jocobtt (approved):
Looks good to me!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -107,42 +119,64 @@ public void execute() throws MojoExecutionException {
...
+        File venv = new File(inVirtualEnvironmentPath);
+
+        if (this.useInProjectVirtualEnvironment) {
+            getLog().debug("`in-project` virtual environment configured for this project");
+            if (StringUtils.isNotBlank(virtualEnvFullPath) && !inVirtualEnvironmentPath.equals(virtualEnvFullPath)) {
+                getLog().warn("'in-project' virtual environment is configured, but an external virtual environment was found!");
+                getLog().warn("Deleting external virtual environment: " + virtualEnvFullPath);
+                deleteVirtualEnv = true;
             }
+        } else {
```
jacksondelametter, line 138: Could this just be an else if?

d-ryan-ashcraft: It definitely could be.  I moved it to just try to make the logic a bit more straightforward to read.

Issue #142, opened 2024-05-20
Add retry logic for publishing to PyPI repositories


d-ryan-ashcraft:
There is no retry logic inherent to PyPI publishing.  Given the typical single project upload pattern used with PyPI, this isn't a huge issue.  On longer monorepo builds, however, less reliable repositories like `test.pypi.org` can timeout.  Adding some basic retry logic would be nice to provide a little more resiliency to these builds.

## Definition of Done
- [x] Add `pypiPushRetries` to `PublishToPyPiRepoMojo` with a default value of three
- [x] Update documentation for the project, including that this can be disabled by setting retries to zero

## Test Steps
- Build the latest version of `habushu-maven-plugin`:
```
mvn clean install -Pbootstrap
```
- Run the targeted `pypiRetries` tagged tests (these were in the last build, but this command will isolate them for easier reading):
```
mvn clean test -Pbootstrap -Dmaven.build.cache.skipCache=true -Dcucumber.filter.tags="@pypiRetries"
```
- Look at the test output to verify that the number of retries logically make sense (NOTE: these are faux retries, but use the same mechanism that the actual deploy call does in the wild)
- Validate that deploy still works with new retry logic (requires Nexus push permissions):
```
mvn clean deploy -Dmaven.build.cache.skipCache=true -Pci
```
- Simulate a network outage 
  - NOTE: we could add a mock server for this, but generally find this to be below the bar needed to warrant the maintenance, complexity, and false-positive nature that is inherent to such tests.  Feel free to disagree, but that's how we are doing it here.
  - Disconnect your network and run the following command:
```
mvn clean deploy -pl :habushu-mixology -Dmaven.build.cache.skipCache=true -Pci
```
- When your see the following error for the **second** time, reenable your network:
```
[ERROR] Connection Error: We were unable to connect to the repository, ensure the url is correct and can be reached.
```
- Validate that the third attempt succeeds to push to the SNAPSHOT repository

## Sample Output of a Retry
For reference, retry output from the last test step will look something like the following:
```
[INFO] --- habushu:2.15.0-SNAPSHOT:publish-to-pypi-repo (default-publish-to-pypi-repo) @ habushu-mixology ---
[INFO] Executing Poetry command: poetry version -s
[INFO] Setting version of Poetry package to publish to 2.15.0.dev1716369907
[INFO] Executing Poetry command: poetry version 2.15.0.dev1716369907
[...]

PR #143, opened 2024-05-22
#142 :sparkles: add retry ability into push for smoother monorepo rel…

d-ryan-ashcraft:
…eases

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
@@ -5,7 +5,17 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 
+import com.google.common.base.Predicates;
+import io.github.itning.retry.RetryException;
+import io.github.itning.retry.Retryer;
+import io.github.itning.retry.RetryerBuilder;
+import io.github.itning.retry.strategy.stop.StopStrategies;
+import io.github.itning.retry.strategy.stop.StopStrategy;
+import io.github.itning.retry.strategy.wait.WaitStrategies;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
@@ -45,6 +55,8 @@ import org.technologybrewery.habushu.exec.PoetryCommandHelper;
  */
 @Mojo(name = "publish-to-pypi-repo", defaultPhase = LifecyclePhase.DEPLOY)
 public class PublishToPyPiRepoMojo extends AbstractHabushuMojo {
+
+    private static final String VERSION = "version";
 
     /**
      * {@link DateTimeFormatter} compliant pattern that configures the numeric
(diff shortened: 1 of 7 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
```diff
@@ -90,6 +102,27 @@ public class PublishToPyPiRepoMojo extends AbstractHabushuMojo {
...
 
+    /**
+     * Specifies the number of times a push to the configured PyPI repository will be attempted before stopping (inclusive
+     * of the initial attempt). While this defaults to three and is fully configurable, it can be set to zero to never
+     * retry or set to any negative number for unlimited retries.  Unlimited retries will follow a fibonacci backoff
+     * interval.
+     */
+    @Parameter(property = "habushu.pyPiPushRetries", defaultValue = "3")
+    protected int pypiPushRetries = 3;
+
```
d-ryan-ashcraft, line 113: I (Informational): purposely not putting these in the main readme.  It has gotten too complicated and is now less useful.  A ticket will flow behind this to simplify it overall and point to these types of supplementary options at the same time.  (TL;DR - we just copy the the comment into the README anyway, so the new process should just point to where additional options can be found while focusing on the big picture items in the README itself)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
```diff
@@ -90,6 +102,27 @@ public class PublishToPyPiRepoMojo extends AbstractHabushuMojo {
...
+    /**
+     * A multiplier to apply to the retry wait time (e.g., round of fibonacci).
+     */
+    @Parameter(property = "habushu.pypiPushRetryMultiplier", defaultValue = "15000")
+    protected long pypiPushRetryMultiplier;
+
+    /**
+     * Maximum time to wait for a retry, even if the backoff algorithm is above this threshold.
+     */
+    @Parameter(property = "habushu.pypiPushRetryMaxTimeout", defaultValue = "3")
```
epwilkins, line 123: A: Might be worth specifying what unit is used for this field

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
```diff
@@ -90,6 +102,27 @@ public class PublishToPyPiRepoMojo extends AbstractHabushuMojo {
...
+     * retry or set to any negative number for unlimited retries.  Unlimited retries will follow a fibonacci backoff
+     * interval.
+     */
+    @Parameter(property = "habushu.pyPiPushRetries", defaultValue = "3")
+    protected int pypiPushRetries = 3;
+
+    /**
+     * A multiplier to apply to the retry wait time (e.g., round of fibonacci).
+     */
+    @Parameter(property = "habushu.pypiPushRetryMultiplier", defaultValue = "15000")
```
epwilkins, line 117: A: This one is way more obvious based on the default, but again might be worth saying the unit is `ms`

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
d-ryan-ashcraft: Good catch - ✅ FIXED.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
d-ryan-ashcraft: Good catch - ✅ FIXED.

Issue #147, opened 2024-06-18
Maven build cache not recognizing new SNAPSHOT plugin jars to invalidate mixology modules


d-ryan-ashcraft:
When working with Habushu, the Maven Build Cache can be as much of a challenge we help with how it is currently configured.  In order to get either `habushu-mixology` of `habushu-mixology-consumer` to rebuild, you currently have to pass in flag to skip the build cache (e.g., `-Dmaven.build.cache.skipCache=true`).  It would be great if that process was not necessary and it would happen sensibly and automatically.

## Test Steps
- Build the plugin:
```
mvn clean install -Pbootstrap
```
- Build the mixology modules:
```
mvn clean install
```
- Rebuild the mixology modules and validate they hit the cache and do not rebuild fully:
```
mvn clean install
```
- Make a change to the `habushu-maven-plugin` source code
- Build the plugin:
```
mvn clean install -Pbootstrap
```
- Build the mixology modules:
```
mvn clean install
```
- Validate that they MISS the cache and DO rebuild fully
- Rebuild the mixology modules and validate they hit the cache and do not rebuild fully:
```
mvn clean install
```

PR #148, opened 2024-06-18
#147 :wrench: improve cache invalidation for plugin updates

d-ryan-ashcraft:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -6,7 +6,6 @@
 import com.vdurmont.semver4j.Semver;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
```
d-ryan-ashcraft, line 8: I (Informational): I cleaned up some sonarlint items as a means to test invalidating the cache.  So conveying those small tweaks into the PR as well.

csun-cpointe (approved):

Issue #146, opened 2024-06-14
Enable leveraging Maven Session to stage Habushu-type dependency source files to the module directory


Cho-William:
## Overview
Currently, Habushu supports retrieving wheels from a custom cache to enable modules to stage locally developed wheels. However, this approach is brittle (when handling path-based dependencies) and requires the user to list the dependencies twice in their module (once for the retrieve wheels goal, and once as a maven dependency). 

Instead, by emulating the [maven-dependency-plugin:unpack-dependencies](https://maven.apache.org/plugins/maven-dependency-plugin/unpack-dependencies-mojo.html) goal, Habushu would be able to retrieve monorepo dependency source code AND retrieve any transitive path-based dependencies. This will remove significant burden from the user, and enable a more resilient pattern for making locally developed monorepo dependencies available to other modules in the monorepo.

## Definition of Done
- [x] Implement test code for the BDD scenarios (below)
- [x] Implement the scenarios
- [x] Add documentation for the new goal
- [x] Update the release notes (via PR blurb)
- [x] Deprecate `cache-wheels` and `retrieve-wheels` goals

## Implementation Details
- Check the pom for any Habushu-type dependencies and recognize them as needed monorepo dependencies
- Leverage the Maven session to copy monorepo dependency source files to the docker build context
- Examine any transitive path-based monorepo dependencies and also stage those source files (while maintaining the existing relative structure)
  - Identifying any transitive path-based dependencies will be facilitated by looking the pom/maven session (as in NOT examining the `pyproject.toml` files)
## BDD Scenarios
```feature
# to be implemented for this issue
Scenario: One Habushu-type dependency is specified with the default configuration
  Given a single Habushu-type dependency
  When the stage-dependencies goal is executed
  Then the source code for the dependency is staged in the build directory
  And all transitive Habushu-type dependencies are staged in the build directory 
  And the relative directory structure is preserved

# future scenarios
Scenario: Multiple Habushu-type dependencies are specified with the aggregate configuration

Scenario: Multiple Habushu-type dependencies are specified with the default configuration
```

## Test Steps
* OTS Only:
  * Pull this forked [branch](https://github.com/Cho-William/habushu/tree/146-stage-dependencies) and run `mvn clean install`
* Final Test Only:
  * Pull the latest dev branch of the technology brewery habushu repo
  * run the following in the root `mvn clean install -Pbootstrap -Dmaven.build.cache.enabled=false`
* Download the attached 
[...]

PR #150, opened 2024-06-21
#146 copying transitive dependencies functionality for containerize-d…

Cho-William:
…ependencies goal

**Release Notes Addition Request**
A new non-default execution goal `containerize-dependencies` has been added the habushu-maven-plugin. This execution is helpful for establishing a virtual environment (of a containerized setting) housing monorepo python dependencies in a more resilient pattern. The goal examines the project's habushu-type dependencies declared in the building project's `pom.xml` and copies relevant source files to the build directory. Transitive habushu-type dependencies are also copied over, with the original file structure preserved to maintain continuity for any path-based dependencies specified throughout the monorepo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,43 @@ For example, developers may use this feature to bind a Habushu module's `compile
 	</plugin>
 ```
 
+### Leveraging the containerize-dependencies Goal to Prepare a Containerized Virtual Environment ###
+If the execution is specified (does not run by default), the `containerize-dependencies` goal will collect a single `habushu` dependency specified the project's `pom.xml`, including all transitive habushu-packaged dependencies. After collecting the set of necessary dependencies, Habushu will copy the project files to the build directory under `venv-support`, while preserving the original structure of the dependency modules to ensure that any path-based dependencies can be leveraged as-is. This directory can then be copied onto a Docker container and used to create a virtual environment capable of running the target Habushu project.
+
+```xml
+	<plugin>
+		<groupId>org.technologybrewery.habushu</groupId>
+		<artifactId>habushu-maven-plugin</artifactId>
+		<configuration>
+		...
+		</configuration>
+		<executions>
+			<execution>
+				<id>containerize-deps</id>
+				<phase>validate</phase>
+				<goals>
+					<goal>containerize-dependencies</goal>
+				</goals>
+			</execution>
+		</executions>
+	</plugin>
+```
+
+The plugin will only examine dependencies that are of the type `habushu` in the dependencies block of the `pom.xml` file. 
+
+```xml
+    <dependencies>
+        <dependency>
+            <groupId>your-group-id</groupId>
+            <artifactId>your-artifact-id</artifactId>
+            <version>your-version</version>
+            <type>habushu</type>
+        </dependency>
+    </dependencies>
+```
+
+Any transitive monorepo dependencie shoud also use this convention to ensure they are captured by the plugin.
+
 ### Leveraging Maven Build Cache for Faster Builds ###
 
 Habushu enables support for faster builds via
@@ -682,13 +719,15 @@ Folder in which Python test files are located - should align with Poetry's proje
 
 Default: `${project.basedir}/tests`
 
-#### cacheWheels ####
+#### cacheWheels (deprecated) ####
+The `cache-wheels` goal has been `deprecated`, please see the `containerize-dependencies` goal instead.
 
 Enables or Disables the copying of wheels into Poetry cache.
 
 Default: `false`
 
-#### wheelDependencies ####
+#### wheelDependencies (deprecated) ####
+The `retrieve-wheels` goal has been `deprecated`, please see the `containerize-dependencies` goal instead.
 
 Optional set of wheel dependencies to retrieve from poetry cache. This allows previously cached external 
 wheel dependencies to be copied into a given target directory if it exists in poetry cache. This logic 
@@ -771,6 +810,11 @@ normal circumstances.
 
 Default: `${project.basedir}/target/habushu.placeholder.txt`
 
+#### workingDirectory ####
+
+Controls where the clean plugin will determine where Poetry projects are located - should always be the basedir of the encapsulating Maven project.
+
+Default: `${project.basedir}`
 
 #### distDirectory ####
 
@@ -847,6 +891,42 @@ Controls whether the build will continue if lint identifies code that violate ch
 
 Default: `true`
 
+#### workingDirectoryRelativeToBasedir ####
+
+Controls the working directory relative to the `${project.basedir}` when examining monorepo dependency modules for copying source files with the `containerize-dependencies` goal. The working directory should be the `${project.basedir}`, hence the `null` default.
+
+Note: this parameter should be consistent with the `workingDirectory` configuration, such that both parameters point to the same expected working directory. For example, if the `workingDirectory` was overriden to `${project.basedir}/custom-workdir`, then `workingDirectoryRelativeToBasedir` should be set to `/custom-workdir`. Setting this parameter should only be necessary if a non-default `workingDirectory` configuration was set.
+
+Default: `null`
+
+#### distDirectoryRelativeToBasedir ####
+
+Controls the expected dist directory relative to the `${project.basedir}` when examining monorepo dependency modules for copying source files with the `containerize-dependencies` goal.
+
+Note: this parameter should be consistent with the `distDirectory` configuration. For example, if the `distDirectory` was set to `${project.basedir}/custom-dist`, then `distDirectoryRelativeToBasedir` should be set to `/custom-dist`. Setting this parameter should only be necessary if a non-default `distDirectory` configuration was set.
+
+Default: `/dist`
+
+#### targetDirectoryRelativeToBasedir ####
+
+Controls the expected target directory relative to the `${project.build.directory}` when examining monorepo dependency modules for copying source files with the `containerize-dependencies` goal.
+
+Note: this parameter should be consistent with the `targetDirectory` configuration. For example, if the `targetDirectory` was overriden to `${project.basedir}/custom-target`, then `targetDirectoryRelativeToBasedir` should be set to `/custom-target`. Setting this parameter should only be necessary if a non-default `targetDirectory` configuration was set.
+
+Default: `/target`
+
+#### anchorSourceDirectory ####
+
+Controls the directory of where containerization files will be copied from with the `containerize-dependencies` goal. If set, this value should be a directory path that houses all the necessary monorepo dependency Habushu modules (including transitive monorepo dependency modules). Additionally, if set, this path should either be an absolute path or a path relative to the `${project.basedir}` of the `containerize-dependencies` goal's execution. Typically, it is not recommended to set this parameter, as the default directory will be sufficient for monorepo use-cases.
+
+Default: the build's execution root directory
+
+#### containerizeSupportDirectory ####
+
+Controls the location of where containerization files will be copied to as part of the `containerize-dependencies` goal.
+
+Default: `${project.build.directory}/containerize-support`
+
 ## The Habushu Build Lifecycle ##
 
 Habushu applies a [custom Maven lifecycle that binds Poetry-based DevSecOps workflow commands](https://fermenter.atlassian.net/wiki/spaces/HAB/pages/2056749057/Dependency+Management+and+Build+Automation+through+Poetry+and+Maven) to the following phases:
@@ -873,11 +953,10 @@ Uses [behave](https://github.com/behave/behave) to execute BDD scenarios that ar
 
 ##### package #####
 
-Builds the `sdist` and `wheel` archives of this project using `poetry build`. It also generates a `requirements.txt` file which is useful when installing the package in a Docker container where you may want to install the dependencies in a specific Docker layer to optimize caching.
+Builds the `sdist` and `wheel` archives of this project using `poetry build`. It also generates a `requirements.txt` file which is useful when installing the package in a Docker container where you may want to install the dependencies in a specific Docker layer to optimize caching. If the `containerize-dependencies` execution is enabled, supporting monorepo dependency source files will be staged.
 
 ##### install #####
-Publishes the `pom.xml` for the module into your local Maven Repository (`~/.m2/repository`). If the **cacheWheels** configuration is set to True, the `wheel` archive will be copied to the poetry cache directory (`~/{poetry-cache-dir}/cache/repositories/wheels/{artifact-id}/`). The **cacheWheels** configuration default behavior is not to cache the `wheel` archive. If the **wheelDependencies** list is set, each specified wheel dependency will be 
-retrieve and placed into the given target directory.
+Publishes the `pom.xml` for the module into your local Maven Repository (`~/.m2/repository`).
 
 ##### deploy #####
 
--- a/habushu-maven-plugin/pom.xml
+++ b/habushu-maven-plugin/pom.xml
@@ -153,6 +153,24 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.13.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-compat</artifactId>
+            <version>${version.maven}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven.plugin-testing</groupId>
+            <artifactId>maven-plugin-testing-harness</artifactId>
+            <version>4.0.0-alpha-2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.junit.platform</groupId>
             <artifactId>junit-platform-suite</artifactId>
             <scope>test</scope>
--- a/habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
+++ b/habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
@@ -1,0 +1,12 @@
+@containerizeDependencies
+Feature: Test staging source files of specified dependencies and transitive path-based dependencies
+
+  Scenario: One Habushu-type dependency is specified with the default configuration
+    Given a single dependency with packaging type habushu
+    When the containerize-dependencies goal is executed
+    Then the source files of the dependency and transitive Habushu-type dependencies are staged for containerization
+
+  Scenario: No Habushu-type dependencies are specified with the default configuration
+    Given no Habushu-type dependencies
+    When the containerize-dependencies goal is executed
+    Then no source files are staged in the build directory
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CacheWheelsMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CacheWheelsMojo.java
@@ -20,7 +20,9 @@ import org.technologybrewery.habushu.util.HabushuUtil;
  *                          wheel files in poetry.
  * 
  * @throws HabushuException
+ * @deprecated This mojo is deprecated, use the containerize-dependencies goal instead
  */
+@Deprecated
 @Mojo(name = "cache-wheels", defaultPhase = LifecyclePhase.INSTALL)
 public class CacheWheelsMojo extends AbstractHabushuMojo {
     /**
@@ -31,6 +33,8 @@ public class CacheWheelsMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
+        getLog().warn("The cache-wheels goal has been deperecated," +
+                "please see the containerize-dependencies goal instead." );
         if (cacheWheels) {
             cacheWheels();
         }
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
@@ -14,7 +14,6 @@ import org.technologybrewery.habushu.util.HabushuUtil;
 import java.io.File;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -122,8 +121,10 @@ public class CleanHabushuMojo extends CleanMojo {
         List<Fileset> filesetsToDelete = new ArrayList<>();
         boolean removeVenvManually = false;
 
-        String virtualEnvFullPath = findCurrentVirtualEnvironmentFullPath();
-        virtualEnvFullPath = getCleanVirtualEnvironmentPath(virtualEnvFullPath);
+        String virtualEnvFullPath = HabushuUtil.findCurrentVirtualEnvironmentFullPath(
+                pythonVersion, usePyenv, patchInstallScript,
+                workingDirectory, rewriteLocalPathDepsInArchives, getLog());
+        virtualEnvFullPath = HabushuUtil.getCleanVirtualEnvironmentPath(virtualEnvFullPath);
 
         String inVirtualEnvironmentPath = HabushuUtil.getInProjectVirtualEnvironmentPath(this.workingDirectory);
         File venv = new File(inVirtualEnvironmentPath);
@@ -188,33 +189,7 @@ public class CleanHabushuMojo extends CleanMojo {
         super.execute();
     }
 
-    /**
-     * Removes (Activated) from path in 1.3.x and higher versions.
-     *
-     * @param virtualEnvFullPath path to clean
-     * @return cleaned path
-     */
-    private static String getCleanVirtualEnvironmentPath(String virtualEnvFullPath) {
-        return StringUtils.replace(virtualEnvFullPath, " (Activated)", StringUtils.EMPTY);
 
-    }
-
-    private String findCurrentVirtualEnvironmentFullPath() throws MojoExecutionException {
-        String virtualEnvFullPath = null;
-        PyenvAndPoetrySetup configureTools = new PyenvAndPoetrySetup(pythonVersion, usePyenv,
-                patchInstallScript, workingDirectory, rewriteLocalPathDepsInArchives, getLog());
-        configureTools.execute();
-
-        try {
-            PoetryCommandHelper poetryHelper = new PoetryCommandHelper(this.workingDirectory);
-            virtualEnvFullPath = poetryHelper.execute(Arrays.asList("env", "list", "--full-path"));
-        } catch (RuntimeException e) {
-            getLog().debug("Could not retrieve Poetry-managed virtual environment path - it likely does not exist",
-                    e);
-        }
-
-        return virtualEnvFullPath;
-    }
 
     /**
      * Creates a new {@link Fileset} that may be used to identify a set of files
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
@@ -1,0 +1,267 @@
+package org.technologybrewery.habushu;
+
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.shared.model.fileset.FileSet;
+import org.apache.maven.shared.model.fileset.util.FileSetManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.habushu.util.HabushuUtil;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Stages the source files of a monorepo dependency
+ * to the target directory along with the source files of
+ * any transitive path-based dependencies.
+ */
+@Mojo(name = "containerize-dependencies", defaultPhase = LifecyclePhase.PACKAGE)
+public class ContainerizeDepsMojo extends AbstractHabushuMojo {
+
+    private static final Logger logger = LoggerFactory.getLogger(ContainerizeDepsMojo.class);
+
+    @Component
+    protected MavenSession session;
+
+    /**
+     * The desired version of Python to use.
+     */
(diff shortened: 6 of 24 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Stages the source files of a monorepo dependency
+ * to the target directory along with the source file of
+ * any transitive path-based dependencies.
+ */
+
+@Mojo(name = "containerize-dependencies", defaultPhase = LifecyclePhase.VALIDATE)
```
d-ryan-ashcraft: S: seems like this should be in the `package` phase instead since that would put it just-in-time for usage and seems more functionally aligned with what is happening?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+public class ContainerizeDependenciesMojo extends AbstractHabushuMojo {
+
+    private static final Logger logger = LoggerFactory.getLogger(ContainerizeDependenciesMojo.class);
+
+    @Component
+    protected MavenSession session;
+
+    protected File anchorSourceDirectory;
```
d-ryan-ashcraft: S: seems like it might be useful for these `anchor*` variables to be plugin parameters so they could be configured via the plugin in non-standard cases

README.md
```diff
@@ -162,6 +162,44 @@ For example, developers may use this feature to bind a Habushu module's `compile
 	</plugin>
 ```
 
+### Leveraging the containerize-dependencies Goal to Prepare a Containerized Virtual Environment ###
+If the execution is specified (does not run by default), the `containerize-dependencies` goal will collect the monorepo python dependency specified in the dependencies block of the project's `pom.xml`, including all transitive dependencies. After collecting the set of necessary dependencies, Habushu will copy the source files to the build directory, while preserving the original structure of the dependency modules to ensure that any path-based dependencies can be leveraged as-is.
```
carter-cundiff: `will collect the monorepo python dependency specified` -> ` will collect any monorepo python dependencies specified`

README.md
```diff
@@ -162,6 +162,44 @@ For example, developers may use this feature to bind a Habushu module's `compile
...
+        <dependency>
+          <groupId>your-group-id</groupId>
+          <artifactId>your-artifact-id</artifactId>
+          <version>your-version</version>
+          <type>habushu</type>
+        </dependency>
+    </dependencies>
+```
+
+Any transitive monorepo dependency `pom.xml` files shoud also use this convention to ensure they are captured by the plugin.
```
carter-cundiff: Any transitive monorepo dependency `pom.xml` files -> Any transitive monorepo dependencies

README.md
epwilkins: A:  Presumably the latter portion of this instruction becomes simplified after the Dockerfile changes are introduces to focus less on the implementation and more on the outcome.
```suggestion
If the execution is specified (does not run by default), the `containerize-dependencies` goal will collect a single `habushu` dependency specified the project's `pom.xml`, including all transitive  `habushu` dependencies. After collecting the set of necessary dependencies, Habushu will copy the project files to the build directory under `venv-support`, while preserving the original structure of the dependency modules to ensure that any path-based dependencies can be leveraged as-is. This directory can then be copied onto a Docker container and used to create a virtual environment capable of running the target Habushu project.
```

epwilkins: I think v1 of this feature is only intended to support a single Habushu project being containerized, so even if the current code _could_ copy multiple, the second half of this feature (currently in flight) will depend on there being only one.

README.md
```diff
@@ -162,6 +162,44 @@ For example, developers may use this feature to bind a Habushu module's `compile
...
 ```
 
+### Leveraging the containerize-dependencies Goal to Prepare a Containerized Virtual Environment ###
+If the execution is specified (does not run by default), the `containerize-dependencies` goal will collect the monorepo python dependency specified in the dependencies block of the project's `pom.xml`, including all transitive dependencies. After collecting the set of necessary dependencies, Habushu will copy the source files to the build directory, while preserving the original structure of the dependency modules to ensure that any path-based dependencies can be leveraged as-is.
+
+```xml
+	<plugin>
+		<groupId>org.technologybrewery.habushu</groupId>
+		<artifactId>habushu-maven-plugin</artifactId>
+		<extensions>true</extensions>
```
epwilkins: S: I don't think this is necessary for the specific goal at hand. This would be needed for a `habushu` packaged project, but not one simply leveraging this feature.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+        return fileSet;
+    }
+
+    /**
+     * Checks listed habushu-type dependencies against the set of projects included in the Maven build's session
+     * @return the corresponding Maven projects that match the habushu-type dependencies
+     */
+    protected Set<MavenProject> getHabushuProjects() {
+        HashSet<MavenProject> habushuProjects = new HashSet<>();
+        collectHabushuDependenciesAsProjects(project, habushuProjects);
```
carter-cundiff, line 115: Can drop the `project` parameter here and just reference it as `this.project` in the `collectHabushuDependenciesAsProjects` implementation.

habushu-maven-plugin/pom.xml
```diff
@@ -191,6 +209,9 @@
                             cucumber.junit-platform.naming-strategy=long
                         </configurationParameters>
                     </properties>
+                    <excludes>
+                        <exclude>**/ContainerizeDependenciesMojoTestCase.java</exclude>
```
epwilkins: Q: Why is this necessary for this class, but not the extremely similar `MojoTestCaseWrapper` in fermenter?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+        return habushuProjects;
+    }
+
+    /**
+     * Collects the projects of the habushu-type dependencies and adds them to the given project set
```
carter-cundiff: Would it be correct to say `Collects the projects with habushu-type dependencies`? Current wording is sort of confusing to me.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param habushuProjects the set to append to with matching projects
+     */
+    private void collectHabushuDependenciesAsProjects(MavenProject currentProject, Set<MavenProject> habushuProjects) {
+        Set<String> habushuDeps = currentProject.getDependencies().stream()
+                .filter(d -> HABUSHU.equals(d.getType()))
+                .map(ContainerizeDependenciesMojo::toGav)
+                .collect(Collectors.toSet());
+        for (MavenProject project : getSession().getProjects()) {
+            if (habushuDeps.contains(toGav(project))) {
+                logger.info("Found habushu-type dependency's project {}.", project);
```
carter-cundiff: `Found habushu-type dependency's project {}.` -> `Found habushu-type dependency's within project {}.`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: that's a better wording, can update

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+    protected void doExecute() throws MojoExecutionException, MojoFailureException {
+        if (this.anchorSourceDirectory == null) {
+            this.anchorSourceDirectory = new File(session.getExecutionRootDirectory());
+        }
+
+        Set<MavenProject> habushuProjects = getHabushuProjects();
+        try {
+            copyRelevant(habushuProjects);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
d-ryan-ashcraft: A (Administrative): It is generally bad practice to throw `RuntimeException`, but we have `HabushuException` for just this purpose.  

https://rules.sonarsource.com/java/tag/error-handling/RSPEC-112/?search=RuntimeException

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Override
+    protected void doExecute() throws MojoExecutionException, MojoFailureException {
+        if (this.anchorSourceDirectory == null) {
+            this.anchorSourceDirectory = new File(session.getExecutionRootDirectory());
+        }
+
+        Set<MavenProject> habushuProjects = getHabushuProjects();
+        try {
+            copyRelevant(habushuProjects);
```
epwilkins: A: Name could be better. We're not doing any check for relevancy, right? Unless I'm misunderstanding what we mean by "relevant".  Maybe just `copySourceCode`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException {
+        doExecute();
+    }
+
+    @Override
+    protected void doExecute() throws MojoExecutionException, MojoFailureException {
+        if (this.anchorSourceDirectory == null) {
+            this.anchorSourceDirectory = new File(session.getExecutionRootDirectory());
```
epwilkins: A: Why make this a field on the class? Same with `anchorOutputDirectory`.  I would think we could just pass it into the method call.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Copies the relevant source files by leveraging {@link FileSet}s to filter appropriately.
+     * @param habushuProjects corresponding projects of the pom's habushu-type dependencies
+     * @throws IOException
+     */
+    private void copyRelevant(Set<MavenProject> habushuProjects) throws IOException {
+        this.anchorOutputDirectory = new File(project.getBuild().getDirectory()
+                + "/venv-support/" + this.anchorSourceDirectory.getName());
+        Path srcRoot = this.anchorSourceDirectory.toPath();
+        Path destRoot = this.anchorOutputDirectory.toPath();
```
epwilkins: A: If we're only using these `Path` objects, seems like we should just construct them directly instead of constructing a `File` first and converting to `Path`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
epwilkins: A: Might be nice to detect if multiple Habushu dependencies are directly listed. Either log a message and indicate which dependency we're choosing as the "main" dependency for which to build the virtual env, or throw an exception.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+    private static String toGav(MavenProject project) {
+        return project.getGroupId() + ":" + project.getArtifactId() + ":" + project.getVersion();
+    }
+
+    protected MavenSession getSession() {
+        return this.session;
+    }
+
+    protected void setAnchorSourceDirectory(File anchorSourceDirectory) {
```
epwilkins: Q: Is this required since we're setting the base directory in the tests??

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Copies the relevant source files by leveraging {@link FileSet}s to filter appropriately.
+     * @param habushuProjects corresponding projects of the pom's habushu-type dependencies
+     * @throws IOException
+     */
+    private void copyRelevant(Set<MavenProject> habushuProjects) throws IOException {
+        this.anchorOutputDirectory = new File(project.getBuild().getDirectory()
+                + "/venv-support/" + this.anchorSourceDirectory.getName());
```
epwilkins: A: Seems pointless to include the source directory root here as part of the output path. We could just put the files directly under `venv-support` since it's already acting as the "container" for the venv source files.  That would then allow us to move this logic to the `getAnchorOutputDirectory` method.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+            Path relativeProjectPath = srcRoot.relativize(projectPath);
+            FileSet fileSet = getDefaultFileSet();
+            fileSet.setDirectory(projectPath.toString());
+            dependencyFileSets.put(relativeProjectPath, fileSet);
+        }
+
+        FileSetManager fileSetManager = new FileSetManager();
+        for (Path project : dependencyFileSets.keySet()) {
+            FileSet fileSet = dependencyFileSets.get(project);
+            logger.info("Copying {} files from {}",
```
d-ryan-ashcraft: A: Maybe something like `Staging monorepo dependency files {} from {}` would be a bit more descriptive?

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+                Path relativePath = project.resolve(includedFile);
+                Files.createDirectories(destRoot.resolve(relativePath).getParent());
+                Files.copy(srcRoot.resolve(relativePath), destRoot.resolve(relativePath));
+            }
+        }
+    }
+
+    private FileSet getDefaultFileSet() {
+        FileSet fileSet = new FileSet();
+        fileSet.addExclude("target/**");
```
d-ryan-ashcraft: S: `target`, `dist`, and `.venv` are all possible to override or be different.  We need to dynamically get the values.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+        fileSet.addExclude("dist/**");
+        return fileSet;
+    }
+
+    /**
+     * Checks listed habushu-type dependencies against the set of projects included in the Maven build's session
+     * @return the corresponding Maven projects that match the habushu-type dependencies
+     */
+    protected Set<MavenProject> getHabushuProjects() {
+        HashSet<MavenProject> habushuProjects = new HashSet<>();
```
d-ryan-ashcraft: A: very minor, but should be `Set<MavenProject> habushuProjects = new HashSet<>()`.  Always want it to be the interface on the LHS unless there is specific implementation class functionality you need to use.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
```diff
@@ -0,0 +1,156 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Copies the relevant source files by leveraging {@link FileSet}s to filter appropriately.
+     * @param habushuProjects corresponding projects of the pom's habushu-type dependencies
+     * @throws IOException
+     */
+    private void copyRelevant(Set<MavenProject> habushuProjects) throws IOException {
```
d-ryan-ashcraft: S: in general, not sure I follow what is `private` vs. `protected`.  Recommend making these protected to allow for extension unless there is a good reason not to.

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/ContainerizeDependenciesMojoTestCase.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Overrides super's {@link #newMavenSession(MavenProject)} to delegate to
+     * the new {@link #newDefaultMavenSession()} introduced in {@link ContainerizeDependenciesMojoTestCase},
+     * which sets the defaults that are normally expected by Maven.
+     */
+    @Override
+    protected MavenSession newMavenSession(MavenProject project) {
+        MavenSession session = newDefaultMavenSession();
+        ArrayList<MavenProject> mavenProjects = new ArrayList<>();
```
d-ryan-ashcraft: A: see prior comment

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/ContainerizeDependenciesMojoTestCase.java
```diff
@@ -0,0 +1,147 @@
...
+
+        try {
+            ProjectBuilder projectBuilder = lookup(ProjectBuilder.class);
+
+            for(File mavenProjectFile: this.mavenProjectFiles) {
+                mavenProjects.add(projectBuilder.build(mavenProjectFile, session.getProjectBuildingRequest()).getProject());
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
d-ryan-ashcraft: A: see prior comment

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
carter-cundiff: Or should it be `Found project for habushu-type dependency`?

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
```diff
@@ -0,0 +1,12 @@
+@containerizeDependencies
+Feature: Test staging source files of specified dependencies and transitive path-based dependencies
+
+  Scenario: One Habushu-type dependency is specified with the default configuration
+    Given a single Habushu-type dependency
```
d-ryan-ashcraft: S: I think language here is very important to make it crisp.  I think we are saying:
```gherkin
Given a single dependency with a packaging type of habushu
...
Then the source files of the dependency and its associated transitive Habushu packaged dependencies are staged for containerization
```

This directly ties it to the both the original pom where it is `<packaging>habushu</packaging>` as well as the dependency where it is `<type>habushu</type>`.

README.md
d-ryan-ashcraft: S (Substantive):`including all transitive dependencies` --> `including all Habushu-packaged transitive dependencies`

habushu-maven-plugin/pom.xml
```diff
@@ -191,6 +209,9 @@
                             cucumber.junit-platform.naming-strategy=long
                         </configurationParameters>
                     </properties>
+                    <excludes>
```
d-ryan-ashcraft: S: I believe if you rename this class to not end with `TestCase`, you can eliminate this.  To me, that's probably worth the simplification.

d-ryan-ashcraft (commented):
Looks good overall!  Left some items that I think we should consider prior to merging.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
carter-cundiff: This was originally done for testing to bypass our method of determining the project root directory which recursively called `project.getParent()`. The `project.getParent()` doesn't work with the maven plugin testing harness as the parents cannot be set. However, now that we're using `session.getExecutionRootDirectory()` for determining the root directory, it's possible this bypass is no longer needed.

carter-cundiff: Also possible it might still be necessary though depending on the output of `session.getExecutionRootDirectory()`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: I was supposed to fix this, thank you for the suggest

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: this was an oopsie on me, can fix

README.md
Cho-William: What Emily said was my thinking. Once we implement the other cases, this README will be refined more

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: second one is better to me, updated

README.md
Cho-William: incorporated, thank you

README.md
Cho-William: that's correct, removed

habushu-maven-plugin/pom.xml
Cho-William: per Ryan's comment, it's due to the naming of the class. this has been fixed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: The anchor attributes were important for setting up the tests. Tried pulling it out, but ended up breaking the tests. For the sake of time, I'm going to keep this as is, but when we clean up the test case in the follow-on, I think we can implement what you're suggesting

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: agreed, refactored

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: addressed, see changes throughout the mojo and test steps

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: I have it as throwing a Habushu exception if the direct dependencies has more than one habushu-type dependency.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: Per previous comment, this is necessary for setting up the tests. Have made a note to look into this for follow-on.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: This `venv-support` was nice to have, because we will eventually be adding the aggregator poetry project under it. Going to keep this for now, but will revisit this for the follow-on.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: agreed, fixed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: You're actually touching on something we intend to implement. This will be a configurable option in the follow-on.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: Agreed, updated

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: That will be covered in a follow-on to enable the user to pass in via configs. We just have it this way as the default, or else using it in the baseline will take a very long time, due to `.venv` in particular taking a while to copy

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: Absolutely, thank you for the catch

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
Cho-William: updated, thank you for the suggest!

habushu-maven-plugin/pom.xml
Cho-William: could not figure out for the life of me why this was getting triggered. that worked, thank you!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
d-ryan-ashcraft: Not sure I follow or agree.  You can get the `target`/`dist` directory by adding the appropriate member variable to the Mojo (worth noting that I'm not sure target should be included either):
```java
    /**
     * Directory in which Poetry places generated source and wheel archive
     * distributions.
     */
    @Parameter(defaultValue = "${project.basedir}/dist", readonly = true, required = true)
    protected File distDirectory;

    /**
     * Directory in which Maven places build-time artifacts - should NOT include dist items.
     */
    @Parameter(defaultValue = "${project.basedir}/target", readonly = true, required = true)
    protected File targetDirectory;

```

Similarly, we can get the virtual environment with an existing call if you look at `CleanHabushuMojo.findCurrentVirtualEnvironmentFullPath()`

These seem like low hanging fruit and important.  I'm a non-concur on pushing them to the next ticket without some additional rationale.

README.md
Cho-William: fixed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
d-ryan-ashcraft: This is also a super easy change - if you just add something similar to the following:
```java
@Parameter(defaultValue = "${project.basedir}/venv-support", readonly = true, required = true)
```

As such, we should do it here versus kick it to the next ticket IMO

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: Suggestion has been implemented. Please note that `findCurrentVirtualEnvironmentFullPath()` and `getCleanVirtualEnvironmentPath()` have been refactored to live in `HabushuUtil`. `findCurrentVirtualEnvironmentFullPath()` now accepts (as arguments) the various parameters it got from the `CleanHabushuMojo`. Necessary parameters have also been added to the `ContainerizeDepsMojo` as well so that it can call on `findCurrentVirtualEnvironmentFullPath()` appropriately.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: The `venv-support` directory has been refactored to `containerization-support`, and is now a configurable parameter. This effectively parameterizes the `anchorOutputDirectory` attribute.  

`anchorSourceDirectory` is also now a configurable parameter. Note that `session.getExecutionRootDirectory()` is not callable for the parameter's default value in the `@Parameter()` annotation, so it is populated with the default value at the start of the mojo's execution.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: Update: these are now configurable parameters

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDependenciesMojo.java
Cho-William: Update: these were changed back to `File` objects because mojos cannot create `Path` objects via the `@Parameter()` annotation. (This is a result of making the anchor attributes configurable mojo parameters per other feedback)

Cho-William:
PR feedback has been addressed, along with updated documentation to reflect the available configurations.

d-ryan-ashcraft (approved):

README.md
```diff
@@ -847,6 +891,42 @@ Controls whether the build will continue if lint identifies code that violate ch
...
+
+#### anchorSourceDirectory ####
+
+Controls the directory of where containerization files will be copied from with the `containerize-dependencies` goal. If set, this value should be a directory path that houses all the necessary monorepo dependency Habushu modules (including transitive monorepo dependency modules). Additionally, if set, this path should either be an absolute path or a path relative to the `${project.basedir}` of the `containerize-dependencies` goal's execution. Typically, it is not recommended to set this parameter, as the default directory will be sufficient for monorepo use-cases.
+
+Default: the build's execution root directory
+
+#### containerizeSupportDirectory ####
+
+Controls the location of where containerization files will be placed to with the `containerize-dependencies` goal.
```
carter-cundiff: will be placed to with the `containerize-dependencies` goal. -> will be copied to as part of the `containerize-dependencies` goal.

Issue #161, opened 2024-06-25
Migrate `dev-dependencies` to Poetry-preferred `group.dev.dependencies` automatically


d-ryan-ashcraft:
Many modules managed by Habushu now have a combination of the following groups:
* `tool.poetry.dev-dependenies` (legacy approach prior to Poetry `1.2.0`)
* `tool.poetry.group.dev.dependencies` (preferred approach from Poetry `1.2.0`+)

It would be helpful to tidy this up with a migration and update Habushu to [use the preferred approach](https://python-poetry.org/docs/managing-dependencies/) when it does work with these dependencies.

> A note about defining a dev dependencies group
> 
> The proper way to define a dev dependencies group since Poetry 1.2.0 is the following:
>
> ``` [tool.poetry.group.dev.dependencies]```
> ...
> 
>This group notation is preferred since Poetry 1.2.0 and not usable in earlier versions. For backwards compatibility with older versions of Poetry, any dependency declared in the dev-dependencies section will automatically be added to the dev group. So the above and following notations are equivalent:
> 
> # Poetry pre-1.2.x style, understood by Poetry 1.0–1.2
> ```[tool.poetry.dev-dependencies]```
> ...
> Poetry will slowly transition away from the dev-dependencies notation which will soon be deprecated, so it’s advised to migrate your existing development dependencies to the new group notation.

## Definition of Done
- [x] Add a new migration to preform the migration above
- [x] Refactor existing group migration logic to be more generic and easily reused
- [x] Update references in codebase and documentation

PR #162, opened 2024-06-25
#161 :sparkles: migration dev-dependencies to group.dev.dependencies …

d-ryan-ashcraft:
…automatically

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/AbstractTomlGroupMigration.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/AbstractTomlGroupMigration.java
@@ -1,0 +1,157 @@
+package org.technologybrewery.habushu.migration;
+
+import com.electronwill.nightconfig.core.CommentedConfig;
+import com.electronwill.nightconfig.core.Config;
+import com.electronwill.nightconfig.core.file.FileConfig;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.baton.AbstractMigration;
+import org.technologybrewery.baton.BatonException;
+import org.technologybrewery.habushu.HabushuException;
+import org.technologybrewery.habushu.util.TomlReplacementTuple;
+import org.technologybrewery.habushu.util.TomlUtils;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Common logic to migrate TOML file entries from one group to another.
+ */
+public abstract class AbstractTomlGroupMigration extends AbstractMigration {
+
+    public static final Logger logger = LoggerFactory.getLogger(AbstractTomlGroupMigration.class);
+
+    protected Map<String, TomlReplacementTuple> replacements = new HashMap<>();
+
+    private boolean hasExistingNewGroup = false;
+
+    protected abstract String getLegacyGroupName();
+
+    protected abstract String getNewGroupName();
+
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        replacements.clear();
+        boolean shouldExecute = false;
+        try (FileConfig tomlFileConfig = FileConfig.of(file)) {
+            tomlFileConfig.load();
+
+            String legacyGroup = getLegacyGroupName();
+            Optional<Config> legacyGroupEntries = tomlFileConfig.getOptional(legacyGroup);
+            if (legacyGroupEntries.isPresent()) {
+                Config foundGroupEntries = legacyGroupEntries.get();
+                Map<String, Object> groupEntryMap = foundGroupEntries.valueMap();
+
+                for (Map.Entry<String, Object> groupEntry : groupEntryMap.entrySet()) {
+                    String groupEntryName = groupEntry.getKey();
+                    Object groupEntryRhs = groupEntry.getValue();
+                    String groupEntryRshAsString = null;
+                    if (groupEntryRhs instanceof CommentedConfig) {
+                        groupEntryRshAsString = TomlUtils.convertCommentedConfigToToml((CommentedConfig) groupEntryRhs);
+                    } else {
+                        groupEntryRshAsString = (String) groupEntryRhs;
+                    }
+                    logger.info("Found [{}] group entry to migrate! ({} = {})", legacyGroup, groupEntryName, groupEntryRshAsString);
+                    TomlReplacementTuple replacementTuple = new TomlReplacementTuple(groupEntryName, groupEntryRshAsString, "");
+                    replacements.put(groupEntryName, replacementTuple);
+                }
+
+                String newGroupName = getNewGroupName();
+                Optional<Config> newGroupEntries = tomlFileConfig.getOptional(newGroupName);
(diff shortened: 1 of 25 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/AbstractTomlGroupMigration.java
```diff
@@ -0,0 +1,157 @@
+package org.technologybrewery.habushu.migration;
```
d-ryan-ashcraft, line 1: I (Informational): refactored existing group migration to an abstract class to make it easy to add multiple

habushu-maven-plugin/src/test/resources/migration/dev-dependencies/with-no-dev-dependenies-group.toml
```diff
@@ -0,0 +1,17 @@
+[tool.poetry]
```
Cho-William: this file's spelling seems to be missing a "c"

habushu-maven-plugin/src/main/resources/migrations.json
```diff
@@ -1,26 +1,41 @@
...
+        "fileSets": [
+          {
+            "includes": [
+              "pyproject.toml"
+            ]
+          }
+        ]
+      },
+      {
+        "name": "dev-dependencies-preferred-poetry-approach-group-migration",
```
Cho-William, line 29: A: I personally think this reads better, but no worries if we think the original naming is better 

```suggestion
        "name": "preferred-poetry-approach-dev-dependencies-group-migration",
```

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/DevDependenciesGroupMigrationSteps.java
```diff
@@ -0,0 +1,85 @@
...
+    @Then("{int} pyproject.toml dependencies exist in the tool.poetry.dev-dependencies group")
+    public void pyproject_toml_dependencies_exist_in_the_tool_poetry_dev_dependencies_group(Integer expectedDevDashDependenciesGroupDependencies) {
+        verifyExecutionOccurred();
+        try (FileConfig tomlFileConfig = FileConfig.of(pyProjectToml)) {
+            tomlFileConfig.load();
+
+            Optional<Config> toolPoetryDevDependencies = tomlFileConfig.getOptional("tool.poetry.dev-dependencies");
+            int numberOfLegacyDependencies = getNumberOfChildren(toolPoetryDevDependencies);
+            assertEquals(expectedDevDashDependenciesGroupDependencies, numberOfLegacyDependencies,
+                    expectedDevDashDependenciesGroupDependencies + " dev-dependencies group should not exist!");
```
Cho-William: A: The string message seems a bit off? Given that we are expecting `expectedDevDashDependenciesGroupDependencies` to be zero, shouldn't this read as:
```suggestion
                    expectedDevDashDependenciesGroupDependencies + " dev-dependencies should exist!");
```

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/DevDependenciesGroupMigrationSteps.java
```diff
@@ -0,0 +1,85 @@
...
+    @Then("{int} pyproject.toml dependencies exist in the tool.poetry.group.dev.dependencies group")
+    public void pyproject_toml_dependencies_exist_in_the_tool_poetry_group_dev_dependencies_group(Integer expectedGroupDevDependencies) {
+        verifyExecutionOccurred();
+        try (FileConfig tomlFileConfig = FileConfig.of(pyProjectToml)) {
+            tomlFileConfig.load();
+
+            Optional<Config> toolPoetryDependencies = tomlFileConfig.getOptional(TomlUtils.TOOL_POETRY_DEV_DEPENDENCIES);
+            int numberOfGroupDevDependencies = getNumberOfChildren(toolPoetryDependencies);
+            assertEquals(expectedGroupDevDependencies, numberOfGroupDevDependencies,
+                    expectedGroupDevDependencies + " group.dev.depependenies should remain in main group!");
```
Cho-William, line 64: See above comment.

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/DevDependenciesGroupMigrationSteps.java
```diff
@@ -0,0 +1,85 @@
...
+
+    protected int getNumberOfChildren(Optional<Config> tomlElement) {
+        int numberOfLegacyDependencies = 0;
+        if (tomlElement.isPresent()) {
+            Config foundDependencies = tomlElement.get();
+            Map<String, Object> dependencyMap = foundDependencies.valueMap();
+
+            for (Map.Entry<String, Object> dependency : dependencyMap.entrySet()) {
+                Object packageRhs = dependency.getValue();
+                numberOfLegacyDependencies++;
```
Cho-William: Q: is there a reason why we can't just get the `.size()` of the `entrySet`?

Cho-William (approved):
All feedback was very minor, LGTM!

epwilkins (approved):

habushu-maven-plugin/src/test/resources/migration/dev-dependencies/with-no-dev-dependenies-group.toml
d-ryan-ashcraft: ✅ Good catch - fixed!

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/DevDependenciesGroupMigrationSteps.java
d-ryan-ashcraft: I borrowed this from another class - I think intent is that this check is always looking for the passed number to not be there.  Updated to ` is an unexpected number of dev-dependencies`.  it happens to be zero in this case, but we could add more tests and that would no longer be the case.

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/DevDependenciesGroupMigrationSteps.java
d-ryan-ashcraft: I think this one is good as is

habushu-maven-plugin/src/main/resources/migrations.json
d-ryan-ashcraft: TomAto-tOmato - gonna leave it the way it is

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/DevDependenciesGroupMigrationSteps.java
Cho-William: agreed, was reading too fast

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/DevDependenciesGroupMigrationSteps.java
d-ryan-ashcraft: ✅ Good catch - refactored this from where it was using the values, but didn't simplify after doing that.  FIXED.

Issue #149, opened 2024-06-20
As a user, I want Habushu to create a virtual environment that can run my python application within my docker container so I can easily containerize single habushu module


csun-cpointe:
## Description
Follow on the Issue #146 , which all the single habushu module and its dependencies are copied to the build target directory, the Habushu will then work on the Dockerfile to resolve and export ONE virtual environment for the ONE monorepo dependency’s poetry project.

## DOD
- [x] Insert the logic to the Dockerfile to build and export the virtual environment.
   - [x] Optimize the build speed (potentially use cache)
   - [x] Update Habushu ReadMe 
   - [x] In the PR description, we should include a draft release note

## BDD Scenario:
```
Scenario: One Habushu-type dependency is specified with Dockerfile updates enabled
Given a single Habushu-type dependency
And a dockerfile to update
When the stage-dependencies goal is executed
Then all the source code to build the dependency is staged in the build directory
And Dockerfile is updated to build a virtual environment for the dependency
```

## Test Steps:

1. OTS Only:
Pull this forked [branch](https://github.com/Cho-William/habushu/tree/149-containerize-dockerfile-update) and run mvn clean install
1. Final Test Only:
Pull the latest dev branch of the technology brewery habushu repo
1. Run the following in the root `mvn clean install -Pbootstrap` 
1. Download the attached project  
[example-monorepo.zip](https://github.com/user-attachments/files/15996597/example-monorepo.zip) and unzip it

1. Create a Dockerfile under `example-monorepo/extensions/extensions-monorepo-dep-consuming-application/src/main/docker/Dockerfile` directory with below content
```
#HABUSHU_BUILDER_STAGE

FROM redhat/ubi9-minimal:latest
RUN microdnf install -y python3.11

#HABUSHU_FINAL_STAGE

ENTRYPOINT ["/opt/venv/bin/python3.11", "-m", "pets.main", "--enable_docs_url", "True"]
```

6. Run `mvn clean install` at the root of the unzipped project
7. Verify `example-monorepo/extensions/extensions-monorepo-dep-consuming-application/src/main/docker/Dockerfile` has been updated.

8. Run `mvn clean install` again at the root directory and verify the content is still the same
9. Run `cd extensions/extensions-monorepo-dep-consuming-application`
10. Run `docker build -f src/main/docker/Dockerfile . -t example.docker`
11. Run `docker run --rm --name example.docker example.docker`
12. Verify the dependency wheel file is created
```
I have 3 pets:
[...]

PR #163, opened 2024-06-25
#149 implement update dockerfile to create virtual environment

csun-cpointe:
Release Notes Addition Request:

To use the `containerize-dependencies` execution goal, please refer to [Leveraging the containerize-dependencies Goal to Prepare a Containerized Virtual Environment](https://github.com/TechnologyBrewery/habushu/blob/dev/README.md#leveraging-the-containerize-dependencies-goal-to-prepare-a-containerized-virtual-environment) in the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ In Okinawa, habushu (pronounced HA-BU-SHU) is a sake that is made with venomous 
 
 ## Why Do You Need Habushu? ##
 
-Habushu is implemented as a series of [Maven](https://maven.apache.org/) plugins that tie together existing tooling in an opinionated fashion similar to how Maven structures Java projects.  
+Habushu is implemented as a series of [Maven](https://maven.apache.org/) plugins that tie together existing tooling in an opinionated fashion similar to how Maven structures Java projects.
 
-**By taking care of manual steps and bringing a predictable order of execution and core naming conventions, Habushu increases time spent on impactful work rather than every developer or data scientist building out scripts that meet their personal preferences.**  
+**By taking care of manual steps and bringing a predictable order of execution and core naming conventions, Habushu increases time spent on impactful work rather than every developer or data scientist building out scripts that meet their personal preferences.**
 
 No one person will agree with all the opinions implemented by Habushu. The value in being able to run entire builds from a single `mvn clean install` command regardless of your prior experience with the projects adds substantial value - both locally and in DevSecOps scenarios.
 
@@ -20,7 +20,7 @@ More information about how Habushu can be used and benefit your project can be f
 In order to use Habushu, the following prerequisites must be installed:
 
 * Maven 3.6+
-  * Maven 3.9+ for use with Maven Build Cache
+* Maven 3.9+ for use with Maven Build Cache
 * Java 11+
 * [Poetry 1.5+](https://python-poetry.org/)
 * [Pyenv](https://github.com/pyenv/pyenv)
@@ -31,7 +31,7 @@ Additionally, Habushu may install and manage:
 
 ## Usage ##
 
-Habushu automates a consistent and predictable build lifecycle by delegating *nearly all* commands related to dependency management, virtual environment activation, and package publishing to [Poetry](https://python-poetry.org/).  As a result, Habushu projects are Poetry projects and are expected to align with the conventions, structure, and configurations utilized by Poetry projects with the `src/` packaging layout.  
+Habushu automates a consistent and predictable build lifecycle by delegating *nearly all* commands related to dependency management, virtual environment activation, and package publishing to [Poetry](https://python-poetry.org/).  As a result, Habushu projects are Poetry projects and are expected to align with the conventions, structure, and configurations utilized by Poetry projects with the `src/` packaging layout.
 
 A Poetry project using the `src/` packaging layout only needs an appropriately configured `pom.xml` within the root level of the project to instrumented through Habushu and participate in a Maven build lifecycle.  The following depicts the required folder structure within an example Habushu module named `spam-ham-eggs`, including the placement of the required `pom.xml` and `pyproject.toml` configurations and utilization of [behave](https://behave.readthedocs.io/en/stable/index.html) for automated testing:
 
@@ -61,14 +61,14 @@ $ ls spam-ham-eggs
 README.rst     pyproject.toml src            tests
 ```
 
-If migrating an existing Python package, consider using `poetry init` and use the interactive guide to create the desired `pyproject.toml` configuration with the appropriate dependencies.  
+If migrating an existing Python package, consider using `poetry init` and use the interactive guide to create the desired `pyproject.toml` configuration with the appropriate dependencies.
 
 If migrating an earlier release of Habushu, follow the same process, but note the following required changes:
 
 * Dependencies specified in `requirements.txt` must be specified in `pyproject.toml` - either use `poetry add` or add them interactively via `poetry init`
 * Python source and test files must be migrated into the folder structure described above, which aligns with the standard `src/` packaging layout.  Assuming that the package name is `spam_ham_eggs`, `src/main/python/*` from the existing Habushu project must be moved into `src/spam_ham_eggs` and `src/test/python/*` from the existing Habushu project must be moved into `tests`
 * Previously, Habushu 1.x modules depended on each other via Maven `<dependency>` declarations.  This approach is deprecated as Habushu 2.x+ expects that other Habushu modules are published to PyPI repositories and consumed as Python packages using Poetry's built-in dependency management capabilties.  For Habushu module dependencies within the same Maven multi-module build hierarchy, consider using editable development installs:
-  
+
     ```toml
     # pyproject.toml
     [tool.poetry.dependencies]
@@ -86,28 +86,28 @@ Set the `<packaging>` type of your module's `pom.xml` to `habushu`:
 
 Add the following plugin definition to your module's `pom.xml` `<build>` section:
 ```xml
-	<plugin>
-		<groupId>org.technologybrewery.habushu</groupId>
-		<artifactId>habushu-maven-plugin</artifactId>
-		<version>2.0.0</version>
-		<extensions>true</extensions>
-	</plugin>
+<plugin>
+    <groupId>org.technologybrewery.habushu</groupId>
+    <artifactId>habushu-maven-plugin</artifactId>
+    <version>2.0.0</version>
+    <extensions>true</extensions>
+</plugin>
 ```
 
-If publishing packages to or consuming dependencies from a private PyPI repository that requires authentication, add your repository credentials to your Maven's `settings.xml` file, usually located under your `~/.m2` folder. See the Configuration section below on how to 
+If publishing packages to or consuming dependencies from a private PyPI repository that requires authentication, add your repository credentials to your Maven's `settings.xml` file, usually located under your `~/.m2` folder. See the Configuration section below on how to
 
 ```xml
-    <server>
-        <!-- ID of the PyPI repository - this ID will be used to reference this repository in
-             the habushu-maven-plugin configuration -->
-        <id>private-pypi-repo</id>
-        
-        <!-- Username of the account by which to access the PyPI repository -->
-        <username>pypi-repo-username</username>
-        
-        <!-- Password of account by which to access the PyPI repository; should be encrypted as per Maven best practices  -->
-        <password>{encrypted-pypi-repo-password}</password>
-    </server>
+<server>
+    <!-- ID of the PyPI repository - this ID will be used to reference this repository in
+         the habushu-maven-plugin configuration -->
+    <id>private-pypi-repo</id>
+
+    <!-- Username of the account by which to access the PyPI repository -->
+    <username>pypi-repo-username</username>
+
+    <!-- Password of account by which to access the PyPI repository; should be encrypted as per Maven best practices  -->
+    <password>{encrypted-pypi-repo-password}</password>
+</server>
 ```
 
 ### Adding Tests ###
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -184,6 +184,24 @@ If the execution is specified (does not run by default), the `containerize-depen
 	</plugin>
 ```
 
+
+To update the Dockerfile with containerize dependencies logic, you need to specify the Dockerfile location by adding the `dockerfile` tag under `configuration` and, also, to insert both the `#HABUSHU_BUILDER_STAGE` and `#HABUSHU_FINAL_STAGE` tags in the Dockerfile in the preferred location.
```
Cho-William:
```suggestion
To update the `Dockerfile` with logic to leverage the monorepo dependencies, specify the `Dockerfile` location by adding the `dockerfile` tag under `configuration` and to insert both the `#HABUSHU_BUILDER_STAGE` and `#HABUSHU_FINAL_STAGE` tags in the `Dockerfile` in the preferred location. 
```

README.md
```diff
@@ -184,6 +184,24 @@ If the execution is specified (does not run by default), the `containerize-depen
 	</plugin>
 ```
 
+
+To update the Dockerfile with containerize dependencies logic, you need to specify the Dockerfile location by adding the `dockerfile` tag under `configuration` and, also, to insert both the `#HABUSHU_BUILDER_STAGE` and `#HABUSHU_FINAL_STAGE` tags in the Dockerfile in the preferred location. 
+Note: Miss adding the tags in the Dockerfile or configuring the dockerfile location will result the Dockerfile not being updated.
```
Cho-William:
```suggestion
Note: Not adding the tags in the `Dockerfile` or configuring the `dockerfile` location configuration will result in the Dockerfile not being updated.
```

README.md
```diff
@@ -184,6 +184,24 @@ If the execution is specified (does not run by default), the `containerize-depen
...
+
+`#HABUSHU_BUILDER_STAGE` tag will be replaced by the habushu builder stage logic to, first, copy the generated dependencies files onto a Docker container and then to create the virtual environment for building the single habushu module.
+
+`#HABUSHU_FINAL_STAGE` tag will be replaced by the habushu final stage logic to copy over the built wheel file to the final stage.
```
Cho-William:
```suggestion
The `#HABUSHU_BUILDER_STAGE` tag will be replaced with the habushu builder stage logic to, first, copy the generated dependencies files onto a Docker container and then to create the virtual environment for building the specified habushu module.

The `#HABUSHU_FINAL_STAGE` tag will be replaced with the habushu final stage logic to copy over the built virtual environment to the final stage.
```

README.md
```diff
@@ -184,6 +184,24 @@ If the execution is specified (does not run by default), the `containerize-depen
...
+```dockerfile
+#HABUSHU_BUILDER_STAGE
+
+FROM redhat/ubi9-minimal:latest
+RUN microdnf install -y python3.11
```
Cho-William, line 198: S: I thought that this was not needed, since the image declaration and setup will already be in the generated logic?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -264,4 +288,31 @@ protected void setAnchorSourceDirectory(File anchorSourceDirectory) {
...
+        }
+    }
+
+    protected String getBuilderStageTemplateFileName() {
+        return "dockerfile_builder_stage_template";
+    }
+
+    protected String getFinalStageTemplateFileName() {
+        return "dockerfile_final_stage_template";
+    }
```
Cho-William: S: these should just be final strings like `protected final String FINAL_STAGE_TEMPLATE = "dockerfile_final_stage_template"`

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/ContainerizeDepsSteps.java
```diff
@@ -87,6 +90,27 @@ public void no_source_files_are_staged_in_the_build_directory() {
...
+    }
+
+    @Then("Dockerfile is updated to build a virtual environment for the dependency")
+    public void dockerfile_is_updated_to_build_a_virtual_env_for_the_dependency() {
+        String updatedDockerfile = getUpdatedDockerfile();
+        StringBuilder contentBuilder = new StringBuilder();
+        Assertions.assertTrue(updatedDockerfile.contains(DockerfileUtil.HABUSHU_BUILDER_STAGE + DockerfileUtil.HABUSHU_COMMENT_START));
+        Assertions.assertTrue(updatedDockerfile.contains(DockerfileUtil.HABUSHU_BUILDER_STAGE + DockerfileUtil.HABUSHU_COMMENT_END));
+        Assertions.assertTrue(updatedDockerfile.contains(DockerfileUtil.HABUSHU_FINAL_STAGE + DockerfileUtil.HABUSHU_COMMENT_START));
+        Assertions.assertTrue(updatedDockerfile.contains(DockerfileUtil.HABUSHU_FINAL_STAGE + DockerfileUtil.HABUSHU_COMMENT_END));
```
Cho-William: S: these should have message strings to describe context if an error occurs

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
```diff
@@ -9,4 +9,11 @@ Feature: Test staging source files of specified dependencies and transitive path
...
-    Then no source files are staged in the build directory
\ No newline at end of file
+    Then no source files are staged in the build directory
+
+  Scenario: One Habushu-type dependency is specified with Dockerfile updates enabled
+    Given a single dependency with packaging type habushu
+     And a dockerfile to update
+    When the containerize-dependencies goal is executed
+    Then all the source files to build the dependency are staged in the build directory
+     And Dockerfile is updated to build a virtual environment for the dependency
```
Cho-William:
```suggestion
     And the Dockerfile is updated to leverage a virtual environment for the dependency
```

README.md
```diff
@@ -184,6 +184,24 @@ If the execution is specified (does not run by default), the `containerize-depen
...
+
+#HABUSHU_FINAL_STAGE
+
+ENTRYPOINT ["/venv/bin/python3.11", "-m", "pets.main", "--enable_docs_url", "True"]
+```
+
```
Cho-William, line 204: C: we need to add the new configuration of `dockerfile` path to this README, preferably at the bottom of the configurations section

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -264,4 +288,31 @@ protected void setAnchorSourceDirectory(File anchorSourceDirectory) {
...
         return anchorOutputDirectory;
     }
+
+    protected void setDockerfile(File dockerfile) {
+        this.dockerfile = dockerfile;
+    }
+
+    protected void performDockerfileUpdateForVirtualEnvironment() {
+        String updatedDockerfile =
+                DockerfileUtil.updateDockerfileWithContainerStageLogic(dockerfile,
```
aaron-gary: For clarity - is the dockerfile variable on this line the class property?  i.e. `this.dockerfile`?  If so can you tack on `this.` 
It's not a big deal, but would make it a little more readable.

aaron-gary (approved):

README.md
```diff
@@ -163,19 +163,19 @@ For example, developers may use this feature to bind a Habushu module's `compile
...
+If the execution is specified (does not run by default), the `containerize-dependencies` goal will collect a single `habushu` dependency specified the project's `pom.xml`, including all transitive habushu-packaged dependencies. After collecting the set of necessary dependencies, Habushu will copy the project files to the build directory under `containerize-support`, while preserving the original structure of the dependency modules to ensure that any path-based dependencies can be leveraged as-is. This directory can then be copied onto a Docker container and used to create a virtual environment capable of running the target Habushu project.
 
 ```xml
 	<plugin>
 		<configuration>
-		...
+          <dockerfile>${project.basedir}/src/main/docker/Dockerfile</dockerfile>
```
carter-cundiff: Think this needs to be spaced in more, looks like the indentation is off

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/DockerfileUtil.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param anchorDirectory the anchor directory
+     * @param moduleBaseDir the module base directory
+     * @return
+     */
+    public static String getContainerStageContentFrom(String template, String anchorDirectory, String moduleBaseDir) {
+        StringBuilder content = new StringBuilder();
+        InputStream inputStream = DockerfileUtil.class.getClassLoader().getResourceAsStream(template);
+
+        try (InputStreamReader inputStreamReader = new InputStreamReader(inputStream);
+             BufferedReader buffer = new BufferedReader(inputStreamReader)) {
```
carter-cundiff: Indentation off by one space here

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/DockerfileUtil.java
```diff
@@ -0,0 +1,104 @@
...
+    public static final String ANCHOR_DIRECTORY = "ANCHOR_DIRECTORY";
+
+    /**
+     * Create the container stage content with given template, anchor directory and the single module's base directory
+     * @param template container stage template name
+     * @param anchorDirectory the anchor directory
+     * @param moduleBaseDir the module base directory
+     * @return
+     */
+    public static String getContainerStageContentFrom(String template, String anchorDirectory, String moduleBaseDir) {
```
carter-cundiff: Would make more sense to call this method `createContainerStageContentFrom`.

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
carter-cundiff: What about a scenario where the Dockerfile already has generated code in it, like this:
```
# syntax=docker/dockerfile:1
FROM scratch
ADD hello /

#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
FROM registry.access.redhat.com/ubi9/python-311:latest AS habushu_builder
WORKDIR /work-dir
COPY ./target/containerize-support/example-monorepo ./containerize-support/example-monorepo

# Poetry and supporting plugin installations
RUN pip install poetry && \
    poetry self add poetry-monorepo-dependency-plugin && \
    poetry self add poetry-plugin-bundle

USER root
WORKDIR /work-dir/containerize-support/example-monorepo/extensions/extensions-python-dep-X
# ensure Poetry's cache directory is propertly set
ENV POETRY_CACHE_DIR="/.cache/pypoetry"
# instruct Docker to persistently store the container's Poetry cache while
# resolving dependencies during the lock process of the venv-specifier and
# building/exporting the virtual environment of the venv-specifier to /venv
RUN --mount=type=cache,target=/.cache/pypoetry/ \
    sed -i 's|develop[[:blank:]]*=[[:blank:]]*true|develop = false|g' pyproject.toml && \
    poetry lock && \
    poetry bundle venv /venv
#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (END)

#HABUSHU_FINAL_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
# copy the pre-built venv from the builder and
COPY --from=habushu_builder /venv /venv
# configure the container to use the venv
ENV PATH="/venv/bin:$PATH"
#HABUSHU_FINAL_STAGE - HABUSHU GENERATED CODE (END)

CMD ["/hello"]
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/DockerfileUtil.java
```diff
@@ -0,0 +1,104 @@
...
+
+                // start skipping the line if reads HABUSHU_COMMENT_START
+                if (!skipLine && line.contains(HABUSHU_COMMENT_START)) {
+                    skipLine = true;
+                }
+
+                // end skipping the line when reads HABUSHU_COMMENT_END
+                if (skipLine && line.contains(HABUSHU_COMMENT_END)) {
+                    skipLine = false;
+                }
```
carter-cundiff: Q: So based on this logic if we run into ` - HABUSHU GENERATED CODE (DO NOT MODIFY)` in the `Dockerfile` then we remove it by using `skipLine`? However, then the new content will never get added as `skipLine` is set to true

carter-cundiff: For example, it seems to me like an input `Dockerfile` like this:
```
# syntax=docker/dockerfile:1
FROM scratch
ADD hello /

#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
FROM registry.access.redhat.com/ubi9/python-311:latest AS habushu_builder
WORKDIR /work-dir
COPY ./target/containerize-support/example-monorepo ./containerize-support/example-monorepo

# Poetry and supporting plugin installations
RUN pip install poetry && \
    poetry self add poetry-monorepo-dependency-plugin && \
    poetry self add poetry-plugin-bundle

USER root
WORKDIR /work-dir/containerize-support/example-monorepo/extensions/extensions-python-dep-X
# ensure Poetry's cache directory is propertly set
ENV POETRY_CACHE_DIR="/.cache/pypoetry"
# instruct Docker to persistently store the container's Poetry cache while
# resolving dependencies during the lock process of the venv-specifier and
# building/exporting the virtual environment of the venv-specifier to /venv
RUN --mount=type=cache,target=/.cache/pypoetry/ \
    sed -i 's|develop[[:blank:]]*=[[:blank:]]*true|develop = false|g' pyproject.toml && \
    poetry lock && \
    poetry bundle venv /venv
#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (END)

#HABUSHU_FINAL_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
# copy the pre-built venv from the builder and
COPY --from=habushu_builder /venv /venv
# configure the container to use the venv
ENV PATH="/venv/bin:$PATH"
#HABUSHU_FINAL_STAGE - HABUSHU GENERATED CODE (END)

CMD ["/hello"]
```
Would result in the following output:
```
# syntax=docker/dockerfile:1
FROM scratch
ADD hello /

CMD ["/hello"]
```

README.md
epwilkins: It's because the surrounding lines use tabs, and this line uses spaces. Might just convert them all to spaces.

README.md
epwilkins: I'd prefer "logic to containerize the python application"

README.md
epwilkins: That would only be true for the builder stage.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -241,6 +259,12 @@ protected void collectHabushuDependenciesAsProjects(MavenProject currentProject,
             if (habushuDeps.contains(toGav(project))) {
                 logger.info("Found project {} as habushu-type dependency.", project);
                 habushuProjects.add(project);
+                if (this.singleHabushuDependency != null) {
+                    if (toGav(project).equals(toGav(this.singleHabushuDependency))) {
+                        String path = project.getBasedir().getPath();
+                        this.singleHabushuDependencyBaseDir = path.substring(path.indexOf(this.anchorSourceDirectory.getName()));
```
epwilkins: A: I feel like this might be easier to follow if we have `getHabushuProjects` return both the full set of projects and the primary project in one result object, then have `copySourceCode` return the relevant path of the primary project.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/DockerfileUtil.java
```diff
@@ -0,0 +1,104 @@
...
+import org.technologybrewery.habushu.HabushuException;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+
+public class DockerfileUtil {
```
epwilkins: A: `DockerfileUtil` sounds a bit too generic given how heavily this class is tied to the functionality of the Mojo.  Something like `ContainerizeDepsDockerfileHelper` might be more clear.

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
epwilkins: There are two more scenarios that will be handled in a follow-on, one of which is the case where the generated content already exists.  With the current implementation not supporting this, you'd have to modify the dockerfile in the build directory and run from there.  Once we implement the follow-on scenarios, you'll be able to more easily directly modify your checked-in Dockerfile if that's the pattern you want to follow.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -264,4 +288,31 @@ protected void setAnchorSourceDirectory(File anchorSourceDirectory) {
...
+
+    protected void setDockerfile(File dockerfile) {
+        this.dockerfile = dockerfile;
+    }
+
+    protected void performDockerfileUpdateForVirtualEnvironment() {
+        String updatedDockerfile =
+                DockerfileUtil.updateDockerfileWithContainerStageLogic(dockerfile,
+                DockerfileUtil.getContainerStageContentFrom(getBuilderStageTemplateFileName(), this.anchorSourceDirectory.getName(), this.singleHabushuDependencyBaseDir),
+                DockerfileUtil.getContainerStageContentFrom(getFinalStageTemplateFileName(), null, null));
```
epwilkins: A: If we rebrand the `DockerfileUtil` class, it probably makes sense to push these two method calls into the `udpateDockerfile...` method.  As it is, I had to jump back and forth between the classes to understand exactly what was happening.

README.md
csun-cpointe: yeah, for the final stage, we only insert the copy over logic.

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
carter-cundiff: Have we verified that the resulting docker image is configured to use the resulting `venv` correctly? Based on my testing, when I run the build and run a docker image with the generated habushu code, my `which python` isn't pointing to the `venv` so it doesn't get used.

carter-cundiff: Could just be an issue with my underlying image though.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/DockerfileUtil.java
csun-cpointe: yeah.. this is actually done by intellij, it was trying to align with the InputStreamReader since it's part of the initial group.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/DockerfileUtil.java
csun-cpointe: no, the expected BUILDER/FINAL stage content will still be added since the follow on logic will add the new content if it finds the #HABUSHU_BUILDER/FINAL_STAGE tag and this is part of the final test scenario too.

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
carter-cundiff: For me, I am using `FROM python:3.11-slim as final` as my final image, and I had to add the following lines to point python to the venv:
```
RUN unlink /venv/bin/python
RUN ln -s /usr/local/bin/python /venv/bin/python
```
Will probably vary by final image though. The original symlink was:
```
/venv/bin/python -> /usr/bin/python3.11
```
which didn't exist on my final image.

carter-cundiff: Could make this generic to link the venv to the underlying python executable by including the following in `#HABUSHU_FINAL_STAGE`:
```
RUN ln -sf $(which python) /venv/bin/python
```
This will delete the potentially incorrect symlink and create a correct one base on the location of the final image python executable.

README.md
csun-cpointe: the `dockerfile` path configuration is added to the readme but I can move it down to be more obvious.

README.md
csun-cpointe: yeah.. reformat the doc with spaces

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
csun-cpointe: I can add this scenario quickly since the code already covers it.

habushu-maven-plugin/src/main/resources/dockerfile_final_stage_template
```diff
@@ -0,0 +1,5 @@
+RUN ln -sf $(which python) /venv/bin/python
```
carter-cundiff: I would think we want to run this at the end of this template. Not sure it matters, but would rather copy `/venv/bin/python` into the final image, then update it's sym link

carter-cundiff: Also would be helpful to have a comment with it, something like:
```
# Update the venv python symlink to point to the underlying python executable
```

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies.feature
```diff
@@ -6,7 +6,16 @@ Feature: Test staging source files of specified dependencies and transitive path
     When the containerize-dependencies goal is executed
     Then the source files of the dependency and transitive Habushu-type dependencies are staged for containerization
 
-  Scenario: No Habushu-type dependencies are specified with the default configuration
-    Given no Habushu-type dependencies
```
csun-cpointe, line 10: I: we no longer needs this scenario since we want to make sure user always insert the dependency

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
```diff
@@ -0,0 +1,108 @@
...
+
+        boolean skipLine = false;
+        try (BufferedReader buffer = new BufferedReader(new FileReader(dockerFile))) {
+            String line = buffer.readLine();
+
+            while (line != null) {
+                line = line.strip();
+
+                // start skipping the line if reads HABUSHU_COMMENT_START
+                if (!skipLine && line.contains(HABUSHU_COMMENT_START)) {
```
carter-cundiff, line 76: So this means we're not going to modify the code if we encounter `- HABUSHU GENERATED CODE (DO NOT MODIFY)`. I would have thought we still want to overwrite this code in case some changes are made to the underlying monorepo and it's dependencies. 

In short, I thought we always wanted to overwrite the habushu generated code, is that not the case?

habushu-maven-plugin/src/test/resources/containerize-dependencies/default-single-monorepo-dep/test-monorepo/extensions/extensions-monorepo-dep-consuming-application/src/main/resources/docker/UpdatedDockerfile
```diff
@@ -0,0 +1,34 @@
...
+sed -i 's|develop[[:blank:]]*=[[:blank:]]*true|develop = false|g' pyproject.toml && \
+poetry lock && \
+poetry bundle venv /venv
+#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (END)
+#HABUSHU_FINAL_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
+# copy the pre-built venv from the builder and
+COPY --from=habushu_builder /venv /venv
+# configure the container to use the venv
+ENV PATH="/venv/bin:$PATH"
+#HABUSHU_FINAL_STAGE - HABUSHU GENERATED CODE (END)
```
carter-cundiff, line 34: Should include 
```
# Update the venv python symlink to point to the underlying python executable
RUN ln -sf $(which python) /venv/bin/python
```
in this section

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
csun-cpointe: yes, the previous updated code will be removed and updated again with latest code

carter-cundiff (approved):

Cho-William (approved):

README.md
Cho-William: We have a separate section that lists out all of the available configurations. We need to add a new section there for the dockerfile config

epwilkins (approved):

PR #164, opened 2024-06-27
Simplify configuration of `containerize-dependencies` goal

epwilkins:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -162,74 +101,65 @@ protected void doExecute() throws MojoExecutionException, MojoFailureException {
...
-        );
-        virtualEnvironmentPath = HabushuUtil.getCleanVirtualEnvironmentPath(virtualEnvironmentPath);
-
-        // if a valid virtual environment was found, relativize and add to the fileSet for exclusions
-        if (virtualEnvironmentPath != null && !virtualEnvironmentPath.isEmpty()
-                && new File(virtualEnvironmentPath).isDirectory()) {
-            Path venvDirRelativeToBasedir = getWorkingDirectoryPath(project)
-                    .relativize(Paths.get(virtualEnvironmentPath));
-            fileSet.addExclude(venvDirRelativeToBasedir + GLOB_RECURSIVE_ALL);
-        }
```
d-ryan-ashcraft, line 232: S: I'm good with the rest of this change, but not good w/ the hardcoded paths.  I just don't think we can let you customize these in the rest of habushu and then force defaults here.

epwilkins: Wouldn't you still be able to customize them?  You'd just provide the alternative like this:
```xml
                <configuration>
                    <defaultSourceSet>
                        <excludes>
                            <exclude>my-build-dir/**</exclude>
                        </excludes>
                    </defaultSourceSet>
                </configuration>
```

d-ryan-ashcraft: Let me rephrase.  If I've already customized them in habushu, I now have to manually do it again.  When we support a variable for path A in the plugin in one place, it seems incongruent to not respect it in another.

d-ryan-ashcraft: Synced w/ Emily offline - I think there is a middle ground approach here that I'll provide when I have a couple.

Issue #165, opened 2024-07-01
Build cache includes from `src` and `tests` not working


d-ryan-ashcraft:
Currently, if you change a file in a habushu module that is in the `src` or `tests` folders, the cache does not trigger a rebuild when it should.

## Step to Recreate
* Update a file in `src` or `tests` of `habushu-mixology`
* Run `mvn clean install`
* The cache is hit and no rebuild is triggered

PR #166, opened 2024-07-01
#165 :bug: fix broken build cache config so changes to src and tests …

d-ryan-ashcraft:
…are picked up

### DIFF
--- a/.mvn/maven-build-cache-config.xml
+++ b/.mvn/maven-build-cache-config.xml
@@ -22,8 +22,8 @@
                 {*.java,*.xml,*.proto,*.properties,*.py,*.txt,*.toml,.env*,*.feature,*.json}
             </glob>
             <includes>
-                <include>src/**</include>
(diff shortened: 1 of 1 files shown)


.mvn/maven-build-cache-config.xml
```diff
@@ -22,8 +22,8 @@
                 {*.java,*.xml,*.proto,*.properties,*.py,*.txt,*.toml,.env*,*.feature,*.json}
             </glob>
             <includes>
```
d-ryan-ashcraft, line 24: I (Information): Rationale explained at: https://codifyiq.substack.com/i/145767837/include-correct-cache-inputs

epwilkins (approved):

cpointe-ibllanos (approved):
Makes sense - without the use of fileset patterns and with the filetype specified as a glob, the name of the path should be all that's needed to search recursively. Approved.

Issue #167, opened 2024-07-05
`containerize-dependencies` configuration could be simplified


epwilkins:
We're currently requiring a complex set of parameters to configure the source files that are copied during the `containerize-dependencies` goal that don't obviously map to the end result.  This is largely because we're relying on `excludes` and not re-using existing configs used by other goals.  We could simplify usage by switching to an includes pattern. By default it makes sense to include:

- the `habushu.sourceDirectory` configured directory
- Poetry files: `poetry.lock`, `poetry.toml`, and `pyproject.toml`
- `README.md` (required by Poetry to build project -- Poetry defaults to the `md` extension but it can be customized)

In order to support edge cases, (such as a non-default readme), the full source fileset should be configurable.

csun-cpointe:
DoD completed!!

PR #168, opened 2024-07-08
#167 Simplify configuration of `containerize-dependencies` goal

epwilkins:
Contains the following changes:

- Pull python module source inclusively based on the existing `habushu.sourceDirectory` configuration and remove all of the configs used to configure module source files exclusively
- Move to the `prepare-package` phase, as the docker build itself should occupy the `package` phase
- Add `dockerContext`, `dockerUser` and `dockerBase` configurations for injecting logic into the `Dockerfile`

`dockerBase` (the base image from which to create the builder and final stages) is required to ensure the virtual environment that is created is compatible with the final image.

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
@@ -13,7 +13,6 @@ import org.apache.maven.shared.model.fileset.FileSet;
 import org.apache.maven.shared.model.fileset.util.FileSetManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.technologybrewery.habushu.util.HabushuUtil;
 import org.technologybrewery.habushu.util.ContainerizeDepsDockerfileHelper;
 
 import java.io.File;
@@ -22,8 +21,6 @@ import java.io.IOException;
 import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
 import java.util.*;
 import java.util.stream.Collectors;
 
@@ -32,7 +29,7 @@ import java.util.stream.Collectors;
  * to the target directory along with the source files of
  * any transitive path-based dependencies.
  */
-@Mojo(name = "containerize-dependencies", defaultPhase = LifecyclePhase.PACKAGE)
+@Mojo(name = "containerize-dependencies", defaultPhase = LifecyclePhase.PREPARE_PACKAGE)
 public class ContainerizeDepsMojo extends AbstractHabushuMojo {
 
     private static final Logger logger = LoggerFactory.getLogger(ContainerizeDepsMojo.class);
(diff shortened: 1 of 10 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
```diff
@@ -26,21 +29,31 @@ public class ContainerizeDepsDockerfileHelper {
...
+    public static String createContainerStageContentFrom(String template, String anchorDirectory, String moduleBaseDir, String owner, String baseImage) {
         StringBuilder content = new StringBuilder();
         InputStream inputStream = ContainerizeDepsDockerfileHelper.class.getClassLoader().getResourceAsStream(template);
 
         try (InputStreamReader inputStreamReader = new InputStreamReader(inputStream); BufferedReader buffer = new BufferedReader(inputStreamReader)) {
             String line = buffer.readLine();
 
             while (line != null) {
-                line = line.strip();
+                line = line.stripTrailing();
```
epwilkins, line 40: I: Preserve leading-space formatting

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -138,98 +99,88 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-            fileSet.addExclude(venvDirRelativeToBasedir + GLOB_RECURSIVE_ALL);
-        }
+        Path srcPath = sourceDirectory.toPath();
+        Path basePath = project.getBasedir().toPath();
+        Path relativeSrc = basePath.relativize(srcPath);
+        fileSet.addInclude(relativeSrc +"/**");
+        fileSet.addInclude("pyproject.toml");
+        fileSet.addInclude("poetry.toml");
+        fileSet.addInclude("poetry.lock");
+        fileSet.addInclude("README.md");
```
epwilkins, line 183: I: This is the primary change in this PR.  Instead of discovering all of the excludes, we drive the fileset based on including the configured source directory and the necessary poetry project files.  The readme is unfortunately required to execute `poetry bundle venv`, so we set it to the default README that poetry generates for new projects.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
```diff
@@ -58,19 +71,23 @@ public static String createContainerStageContentFrom(String template, String anc
...
+        int firstFromLine = -1;
 
         boolean skipLine = false;
         try (BufferedReader buffer = new BufferedReader(new FileReader(dockerFile))) {
             String line = buffer.readLine();
 
             while (line != null) {
-                line = line.strip();
+                line = line.stripTrailing();
+                if(firstFromLine < 0 && line.strip().startsWith("FROM")) {
```
epwilkins, line 88: I: `ARG` values that are used in _any_ `FROM` line must be declared before the _first_ `FROM` line.  By always inserting the builder stage logic just prior to the first `FROM`, we ensure that any `ARG` values used in `FROM` are functionally preserved.

habushu-maven-plugin/src/main/resources/dockerfile_final_stage_template
```diff
@@ -1,6 +1,4 @@
-# copy the pre-built venv from the builder and
-COPY --from=habushu_builder /venv /opt/venv
-# configure the container to use the venv
-ENV PATH="/opt/venv/bin:$PATH"
-# update the venv python symlink to point to the default python executable
-RUN ln -sf $(which python) /opt/venv/bin/python
\ No newline at end of file
+FROM BASE_IMAGE AS final
```
epwilkins, line 1: I: I explained this in the PR body, but this change is required to ensure that the virtual environment works. There are 2 primary things that are impacted if we allow the final export image to differ from the builder image:
 - symlinks within the venv are broken (namely the `python` link)
 - platform-specific packages may be installed incorrectly (I was seeing pydantic installed as `lib64` instead of `lib` if the UBI minimal image was used to build and the python:3.11-slim image was used to execute.)
 
 Theoretically, any of the distributable [compatibility tags](https://peps.python.org/pep-0425/#overview) could be mismatched, causing the wrong package to be installed.

Cho-William (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -41,93 +38,57 @@ public class ContainerizeDepsMojo extends AbstractHabushuMojo {
...
+     * The base image to use for the Dockerfile. The base image will be used both to bundle the virtual environment for
+     * the target project and to run the final container, as the venv must be built on the same platform as the final
+     * runtime. The base image must have the correct Python version resolvable via the PATH.
      */
-    @Parameter(property = "habushu.dockerfile")
-    protected File dockerfile;
-
-    private Path anchorOutputDirectory;
+    @Parameter(defaultValue = "python:3.11-slim", property = "habushu.dockerBase")
+    protected String dockerBase;
```
carter-cundiff, line 88: Should we add `dockerUser` and `dockerBase` to the README?

carter-cundiff (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
epwilkins: We've talked a bit about trying to move away from our current exhaustive pattern of documentation, as it's become so unwieldy that its usefulness is questionable.  But maybe it makes sense to at least touch on `dockerBase` in the main documentation for the goal/feature? I'll see if it's a natural fit there while still being succinct.

PR #170, opened 2024-07-08
Add bootstrap to release so it can be run on a clean environment

epwilkins:


release-habushu.sh
```diff
@@ -19,6 +19,13 @@ if [[ "$?" -ne 0 ]] ; then
   echo 'Process failed! Unable to check out release branch!'; exit 1
 fi
 
+echo "///////////  Bootstrap Habushu plugin ///////////"
+mvn clean install -Pbootstrap
```
epwilkins, line 23: I: Allows `scm:check-local-modification` and `versions:set` to run as the SNAPSHOT version isn't always released.

release-habushu.sh
```diff
@@ -41,8 +48,7 @@ if [[ "$?" -ne 0 ]] ; then
 fi
 
 echo "///////////  Commit all changes that reflect the target release version and create a tag ///////////"
-mvn scm:checkin -Dmessage="Prepare release habushu-$1"
-mvn scm:tag -Dtag=habushu-$1
+mvn scm:checkin -Dmessage="Prepare release habushu-$1" && mvn scm:tag -Dtag=habushu-$1
```
epwilkins, line 51: I: Technically because these were two separate commands, the exit code check below only checked if the tag was successful and not if the checkin was successful.

d-ryan-ashcraft (approved):

carter-cundiff (approved):

Issue #173, opened 2024-07-09
Split `dockerBase` between builder and final stages so that final images can be slimmed down.


epwilkins:
Right now, both the image used to bundle the virtual environment during `containerize-dependencies` is exactly the same as the image used as the final packaging/runtime image.  Because dev tools like `gcc` are often required during the build but not at runtime, final images can be much smaller if we allow building the virtual env with a more "full" base image and use a slimmed down version of that base image for the final image.

This does introduce some complexity for users as they will need to ensure that the value set for the builder stage is compatible with the value set for the final stage, but the old functionality can easily be used by simply providing the same value for both.

PR #174, opened 2024-07-09
#173 add support for a separate final image in containerize-dependencies

epwilkins:


README.md
```diff
@@ -172,8 +172,10 @@ onto a Docker container and used to create a virtual environment capable of runn
 The plugin will automatically inject logic for building and using this virtual environment into a pre-existing Dockerfile
 by default. The `dockerfile` configuration must be set to the target Dockerfile. To disable the Dockerfile update
 altogether, set the `updateDockerfile` configuration to `false`. Because the final image must match the image that builds
-the virtual environment, Habushu must know what the base image for the final output should be. By default, `python:3.11-slim`
-is used, but this can be customized with the `dockerBase` and `dockerUser` configurations.
+the virtual environment, Habushu defaults to using `python:3.11` to build the virtual environment and  `python:3.11-slim`
+as the final image that packages/runs the virtual environment. This can be customized with the `dockerBuilderBase`,
+`dockerFinalBase`, and `dockerUser` configurations, but care must be taken to ensure the builder image platform is
+compatible with the final image platform so that the virtual environment is compatible.
```
Cho-William: A: doesn't have to be exactly this, but the current working just felt like it was using the word compatible too much
```suggestion
`dockerFinalBase`, and `dockerUser` configurations, but care must be taken to ensure the virtual environment is compatible with both the builder image and final image platforms.
```

epwilkins: Ah I didn't even notice I used the word twice. I'll clean it up.

Cho-William (approved):

README.md
```diff
@@ -172,8 +172,10 @@ onto a Docker container and used to create a virtual environment capable of runn
 The plugin will automatically inject logic for building and using this virtual environment into a pre-existing Dockerfile
 by default. The `dockerfile` configuration must be set to the target Dockerfile. To disable the Dockerfile update
 altogether, set the `updateDockerfile` configuration to `false`. Because the final image must match the image that builds
-the virtual environment, Habushu must know what the base image for the final output should be. By default, `python:3.11-slim`
-is used, but this can be customized with the `dockerBase` and `dockerUser` configurations.
+the virtual environment, Habushu defaults to using `python:3.11` to build the virtual environment and  `python:3.11-slim`
```
carter-cundiff: `Because the final image must match the image that builds the virtual environment`. This is no longer the case right?

carter-cundiff (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
```diff
@@ -69,11 +69,14 @@ public static String createContainerStageContentFrom(String template, String anc
...
+     * @param finalBaseImage the image to use for running the virtual environment
      * @return updated Dockerfile content
      */
-    public static String updateDockerfileWithContainerStageLogic(File dockerFile, String anchorDirectory, String moduleBaseDir, String owner, String baseImage) {
-        String builderStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(BUILDER_STAGE_TEMPLATE, anchorDirectory, moduleBaseDir, owner, baseImage);
-        String finalStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(FINAL_STAGE_TEMPLATE, null, null, owner, baseImage);
+    public static String updateDockerfileWithContainerStageLogic(File dockerFile, String anchorDirectory, String moduleBaseDir, String owner, String builderBaseImage, String finalBaseImage) {
+        String builderStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(BUILDER_STAGE_TEMPLATE, anchorDirectory, moduleBaseDir, owner, builderBaseImage);
+        String finalStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(FINAL_STAGE_TEMPLATE, null, null, owner, finalBaseImage);
```
csun-cpointe, line 79: Q: are we override their final image? if there is a startup script in their final image, will that be an issue?

epwilkins: Yes we set the final image `FROM`, so they will have to use the one that is in the snippet. (This was a decision made in the [last PR](https://github.com/TechnologyBrewery/habushu/pull/168).)  Because of the nuance of creating a compatible venv, IMO it's best to force them to think about it.  If they don't like the default final image, they can still customize it but then should also customize the builder base or the venv likely won't work.

As for script or customization, they can easily add more logic _after_ the FINAL block injected by Habushu, including copying over any startup scripts.

README.md
epwilkins: Reworded this section more significantly to put the problem of compatible venv up front.  Made this part more clear as a result.

csun-cpointe (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
epwilkins: Discussed offline the various concerns and came to consensus that this is probably a good base for the functionality to start from.  As we get more feedback and see more edge cases, we can fine tune based on that info.

PR #182, opened 2024-07-31
#181 Update habushu container support to be more consistent across build platforms

### LINKED ISSUE #181: Update habushu container support to be more consistent across build platforms
Release `1.4.0` by the [poetry bundle plugin](https://github.com/python-poetry/poetry-plugin-bundle) appears to have a bug so that if a project has an existing virtual environment, the target location is ignored. This means the venv bundle at `/opt/venv` wasn't being created as expected.

You might not see this issue locally if you are using in-project virtual environments. For reasons outside the scope of this issue, our remote environment was using a `poetry.toml` file that did not use an in-project env thus encountering the bug. To provide better consistency between CI and local (and between Maven/Habushu and native Poetry) we recommend committing `poetry.toml`'s and lock `poetry bundle plugin` to a specific version.

carter-cundiff:
Fixes #181

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
@@ -171,23 +171,18 @@ public class InstallDependenciesMojo extends AbstractHabushuMojo {
     }
 
     private void setUpInProjectVirtualEnvironment(PoetryCommandHelper poetryHelper) throws MojoExecutionException {
-        String inProjectVirtualEnvironmentPath = HabushuUtil.getInProjectVirtualEnvironmentPath(getPoetryProjectBaseDir());
-        File venv = new File(inProjectVirtualEnvironmentPath);
-        if (this.useInProjectVirtualEnvironment) {
-            if (!venv.exists()) {
-                getLog().info("Configuring Poetry to use an in-project virtual environment...");
-                configureVirtualEnvironmentsInProject(true);
-            }
-        } else {
-            List<String> arguments = new ArrayList<>();
-            arguments.add("config");
-            arguments.add("virtualenvs.in-project");
-            arguments.add("--local");
-            String currentInProjectSetting = poetryHelper.execute(arguments);
-
-            if (Boolean.TRUE.equals(Boolean.valueOf(currentInProjectSetting))) {
-                configureVirtualEnvironmentsInProject(false);
-            }
+        List<String> arguments = new ArrayList<>();
+        arguments.add("config");
+        arguments.add("virtualenvs.in-project");
+        arguments.add("--local");
+        String currentInProjectSetting = poetryHelper.execute(arguments);
+        
+        // update the poetry config to match the useInProjectVirtualEnvironment boolean
+        if (this.useInProjectVirtualEnvironment && Boolean.FALSE.equals(Boolean.valueOf(currentInProjectSetting))) {
+            getLog().info("Configuring Poetry to use an in-project virtual environment...");
+            configureVirtualEnvironmentsInProject(true);
+        } else if (!this.useInProjectVirtualEnvironment && Boolean.TRUE.equals(Boolean.valueOf(currentInProjectSetting))) {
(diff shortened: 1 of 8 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -171,23 +171,18 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
...
-            String currentInProjectSetting = poetryHelper.execute(arguments);
-
-            if (Boolean.TRUE.equals(Boolean.valueOf(currentInProjectSetting))) {
-                configureVirtualEnvironmentsInProject(false);
-            }
+        List<String> arguments = new ArrayList<>();
+        arguments.add("config");
+        arguments.add("virtualenvs.in-project");
+        arguments.add("--local");
+        String currentInProjectSetting = poetryHelper.execute(arguments);
```
epwilkins, line 178: Q: This is just to ensure we've got the poetry.toml file sync'd in the case that the user has a pre-existing .venv folder, but the config hasn't been set in the toml file? (Perhaps in the case it was manually modified to remove the config?)

epwilkins (approved):

epwilkins:
Update draft release notes if you have access. If not let me know and I can update them.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -171,23 +171,18 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
...
+        
+        // update the poetry config to match the useInProjectVirtualEnvironment boolean
+        if (this.useInProjectVirtualEnvironment && Boolean.FALSE.equals(Boolean.valueOf(currentInProjectSetting))) {
+            getLog().info("Configuring Poetry to use an in-project virtual environment...");
+            configureVirtualEnvironmentsInProject(true);
```
carter-cundiff, line 183: I: Previously we only updated the poetry config for `virtualenvs.in-project` when a `.venv` didn't exist. This will set the `virtualenvs.in-project` regardless of if a `.venv` exists, ensuring the `poetry.toml` always matches the value of `useInProjectVirtualEnvironment`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
carter-cundiff: Correct, see [comment below](https://github.com/TechnologyBrewery/habushu/pull/182#discussion_r1698948930)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
d-ryan-ashcraft: We should test going both in-project versus default location and back for this. I recall that the original logic accounted for a challenge in that sequencing.

carter-cundiff: Tested going back and forth between in project and external virtual environments without any issue. Both the current version (`2.16.1`) and the updated version (`2.17.0-SNAPSHOT`) showed the following behavior:

Case 1: external -> in project
- External environment is deleted
- In project environment is created and used
- `poetry.toml` is updated with `in-project = true`

Case 2: in project -> external
- In project environment is deleted
- External environment is created and used
- `poetry.toml` is updated with `in-project = false`


The only difference in behavior is if you have `habushu.useInProjectVirtualEnvironment=true` and don't have `in-project = true` in your `poetry.toml` but have an in project environment:
- `2.17.0-SNAPSHOT` will add `in-project = true` to your `poetry.toml` and use the in project environment
- `2.16.1` will leave the `poetry.toml` unchanged and create an external environment, ignoring the in project environment

The current behavior of `2.16.1` seems like a bug to me that would be resolved with the `2.17.0-SNAPSHOT` behavior.

d-ryan-ashcraft (approved):

Issue #177, opened 2024-07-24
pypiUploadSuffix with defaultValue = "" interpreted as null instead of an empty string


chang-annie:
Currently, the `pypiUploadSuffix` has its `defaultValue` set to `""` inside the plugin mojo, which causes it to be interpreted as `null` instead of an empty string. Because of this, when the user does not explicitly set the `pypiUploadSuffix` to another value, the generated `repoUrl` looks something like this:

```
https://nexus.test.com/repository/test-pypi-internal/null/
```

Note: this GH issue only calls out `pypiUploadSuffix` because it's currently the only property that has its `defaultValue = ""` but this behavior is expected with any and all future properties with the same `defaultValue`.

[related Stack Overflow post](https://stackoverflow.com/questions/77546310/how-to-make-a-maven-plugin-mojo-parameters-default-value-an-empty-string)

PR #184, opened 2024-08-09
#177 Fixed pypiUploadSuffix null issue

jacksondelametter:
PR to fix an issue with the pypiUploadSuffix property where Maven was not setting the default value to an empty String. 

Release Notes:
* Fix issue where pypiUploadSuffix was not defaulting to a empty String value.

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
@@ -90,9 +90,10 @@ public class PublishToPyPiRepoMojo extends AbstractHabushuMojo {
      * "legacy/").  This variable allows customization in a manner that does not impact the installation API for the
      * same repository.  Defaults to empty as the most common scenario when overriding the repository URL is to leverage
      * one of the repositories mentioned above.
+     * Note: The property must be set equal to "" in order for Maven to not set the default value to null
      */
     @Parameter(property = "habushu.pypiUploadSuffix", defaultValue = "")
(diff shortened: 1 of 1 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
```diff
@@ -308,14 +308,13 @@ protected Retryer<Boolean> getRetryer() {
...
 
     String getRepositoryUrl(boolean publishToDev) {
-        String repoUrl;
+        String repoUrl = addTrailingSlash(pypiRepoUrl);
         if (publishToDev) {
-            repoUrl = addTrailingSlash(devRepositoryUrl) + devRepositoryUrlUploadSuffix;
-        } else {
-            repoUrl = addTrailingSlash(pypiRepoUrl) + pypiUploadSuffix;
+            repoUrl = addTrailingSlash(devRepositoryUrl) + addTrailingSlash(devRepositoryUrlUploadSuffix);
+        } else if(!pypiUploadSuffix.isEmpty()) {
```
cwoods-cpointe: a: Could use StringUtils to prevent npe in case of the upload suffix being set to null

cwoods-cpointe (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
jacksondelametter: Changed

epwilkins (approved):

Issue #180, opened 2024-07-31
Update to use parent 10 (and transitively Python 3.11.9 and Java 17)


d-ryan-ashcraft:
Upgrade to extend from `technology-brewery:parent` version 10.  

This will also move the default Python version to 3.11.9 (the oldest patch version with an installer available) and Java 17 for building. We will continue to emit Java 11-compatible jars until they are sunsetted in October 2024.

This also involves the following upgrades:
* `io.cucumber:*`: `7.16.1` --> `7.18.1`
* `org.codehaus.plexus:plexus-utils`: `3.4.1` --> `4.0.1`
* `com.electronwill.night-config:toml`: `3.6.7` --> `3.8.1`
* `org.technologybrewery.baton:baton-maven-plugin`: `1.0.0` --> `1.1.1`
* `org.apache.maven.plugin-tools:maven-plugin-annotations`: `3.12.0` --> `3.13.1`
* `org.apache.maven.plugin-tools:maven-plugin-plugin`: `3.12.0` --> `3.13.1`
* `net.masterthought:maven-cucumber-reporting`: `5.8.0` --> `5.8.1`

PR #185, opened 2024-08-14
#180 :arrow_up: upgrade to parent; transitively to Java 17

d-ryan-ashcraft:


pom.xml
```diff
@@ -68,13 +66,6 @@
                 <artifactId>commons-codec</artifactId>
                 <version>1.16.1</version>
             </dependency>
```
d-ryan-ashcraft, line 68: I (Informational): This round of upgrades allowed this workaround to be removed with the `5.8.1` upgrade of `net.masterthought:cucumber-reporting`.

cpointe-ibllanos (approved):

Issue #186, opened 2024-08-29
Repetitive python and poetry checks slow down build process


d-ryan-ashcraft:
When running a multi-module build, the checks for the Python version and poetry installation take time.  For instance:
* Initial check on a `mvn clean install`:
  * `clean`: 2.5s
  * `install`: 1.5s

This is even more noticeable when running the above with build cache hits, where there is a 2.5-second pause on the first cache hit and then a 1.5-second hit on every subsequent one. The repetitive nature of these checks underscores the need for a more efficient process.  

Standard use should be able to replay on validating Poetry exactly once.  Python, in practice, can be validated once as well.  But switching target Python versions by module is possible, so it would be nice if that option is still respected in a fix.

Definition of Done
- No-op validation of Poetry after the first successful validation
- No-op validation of Python version after the first successful validation - but only when the expected version matches the previously matched version
- Ensure support for multithreaded execution

This saves about 1 second on the first module, then 2 seconds on subsequent modules.  This can be seen in the original and subsequent runs of the Habushu mixology modules below:

Original Run:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 2.17.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.137 s]
[INFO] habushu::Mixing Drinks with Habushu ................ SUCCESS [ 22.709 s]
[INFO] habushu::Mixing Drinks with Habushu Consumer ....... SUCCESS [ 19.440 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  42.902 s
[INFO] ------------------------------------------------------------------------
```

Optimized Run:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 2.17.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.136 s]
[INFO] habushu::Mixing Drinks with Habushu ................ SUCCESS [ 21.419 s]
[INFO] habushu::Mixing Drinks with Habushu Consumer ....... SUCCESS [ 17.368 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  39.481 s
[INFO] ------------------------------------------------------------------------
```

PR #187, opened 2024-08-29
#186 :zap: improve runtime speed by not duplicating repetitive python…

d-ryan-ashcraft:
… and poetry validations

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
@@ -22,7 +22,7 @@ import java.util.List;
  * developer's machine to support the same functionality across multiple Mojo implementations. These include:
  * <ul>
  * <li>pyenv</li>
- * <li>Poetry (installed version must satisfy {@link #POETRY_VERSION_REQUIREMENT})</li>
+ * <li>Poetry (installed version must satisfy {@link PoetryUtil#POETRY_VERSION_REQUIREMENT})</li>
  * <li>Required Poetry plugins (currently only {@code poetry-monorepo-dependency-plugin})</li>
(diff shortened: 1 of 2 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -34,6 +34,10 @@ public class PyenvAndPoetrySetup {
      */
     static final String PYTHON_DEFAULT_VERSION_REQUIREMENT = "3.11.4";
 
+    public static boolean alreadyValidatedPoetryInstallation;
+    public static String alreadyValidatedPoetryVersion;
+    public static String priorActivePythonVersionActivated;
```
epwilkins: Q: Is Habushu generally not marked thread-safe and/or do we not have have to worry about any race conditions for these checks?

epwilkins (approved):
If the concern around parallel builds is unfounded, then LGTM

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
d-ryan-ashcraft: That's a good question. To be sure, I can add a synchronized block to the update for Poetry. The Python version can be added to a ThreadLocal to ensure consistency across the threads.  

We could also mark it explicitly as not thread-safe, as we rarely see anyone do a multi-threaded build due to the complexity of debugging it.

But it's easy enough to support this, so I'll make the updates.

d-ryan-ashcraft:
> If the concern around parallel builds is unfounded, then LGTM

Updated. If you run it with `-T2` now, you'll see it respects both threads with independent validations for their first runs.

epwilkins (approved):

PR #193, opened 2025-01-06
Support poetry 2.x.x

cpointe-ibllanos:


habushu-maven-plugin/src/main/resources/dockerfile_builder_stage_template
```diff
@@ -3,7 +3,7 @@ FROM BASE_IMAGE AS habushu_builder
 RUN python -m ensurepip --upgrade && \
     pip install poetry && \
     poetry self add poetry-monorepo-dependency-plugin && \
-    poetry self add poetry-plugin-bundle@1.3.0
+    poetry self add poetry-plugin-bundle@1.5.0
```
d-ryan-ashcraft, line 6: S: I wonder if we don't want to specify a version here, given Python's "always forward compatible" mantra.  Poetry should find the right version based on what version of Poetry is installed this was a well.

meliz19:
Given some of the changes in Poetry 2.0.0, have we verified that this fix will work?

chang-annie:
this was recently resolved with [Issue 194](https://github.com/TechnologyBrewery/habushu/issues/194)

Issue #194, opened 2025-01-06
Add support for Poetry 2.0


epwilkins:
## Description
Poetry 2.0.0 was [released yesterday](https://python-poetry.org/blog/announcing-poetry-2.0.0).  There are some breaking changes to the CLI that cause Habushu to be inoperable. We should detect when version 2.0 is being used and adapt the CLI commands to the new format.

For the containerization logic, we'll also need to determine what approach we want to take for the poetry version. We've explicitly set the `poetry-bundle-plugin` to version 1.3.0 but we did not set the Poetry version. This means that Poetry 2.0 is installed, but instead of using the latest plugin version (which is compatible with 2.0) we use 1.3 which is incompatible.  We should either explicitly set both versions or set neither.

## DOD
- [x] Add support for Poetry 2.0.0 by adding logic to check if Poetry version 2.0.0, then addressing any breaking changes in the CLI commands and in the containerization logic
- [x] Validate 2.0.0 version, one older version, and one newer version are all compatible
- [x] Update documentation and tests, if needed
- [ ] Reach goal: cut Habushu release - to do after migration is written

## Test Steps/Strategy 
1. Install Poetry v2.0.0 `pipx install poetry==2.0.0`
2. Pull the latest dev branch of the technology brewery habushu repo and build using `mvn clean install -Pbootstrap`
3. Download the attached project [example-monorepo.zip](https://github.com/user-attachments/files/18677197/example-monorepo.zip) and unzip it
4. Create a Dockerfile under `example-monorepo/extensions/extensions-monorepo-dep-consuming-application/src/main/docker/` directory with below content
```
#HABUSHU_BUILDER_STAGE

FROM redhat/ubi9-minimal:latest
RUN microdnf install -y python3.11

#HABUSHU_FINAL_STAGE

ENTRYPOINT ["/opt/venv/bin/python3.11", "-m", "pets.main", "--enable_docs_url", "True"]
```
5. Run `mvn clean install` at the root of the unzipped project
6. Verify `example-monorepo/extensions/extensions-monorepo-dep-consuming-application/src/main/docker/Dockerfile` has been updated to include the following section 
```
RUN python -m ensurepip --upgrade && \
    pip install poetry==2.0.1 && \
    poetry self add poetry-monorepo-dependency-plugin@1.2.0 && \
    poetry self add poetry-plugin-bundle@1.5.0
```
7. Run `cd extensions/extensions-monorepo-dep-consuming-application`
[...]

PR #202, opened 2025-02-06
 #194 Add support for Poetry 2.0.0

chang-annie:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -97,6 +97,24 @@ public class ContainerizeDepsMojo extends AbstractHabushuMojo {
     @Parameter(defaultValue = "docker.io/python:3.11-slim", property = "habushu.dockerFinalBase")
     protected String dockerFinalBase;
 
+    /**
+     * The version of Poetry to install in the container.
+     */
+    @Parameter(defaultValue = "2.0.1", property = "habushu.dockerPoetryVersion")
```
chang-annie, line 103: I: take the Poetry and plugin versions as parameters to the `containerize-dependencies` Mojo, then use the parameters to update the generated Dockerfile. Defaults to the highest versions tested.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -23,13 +23,21 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
...
+        );
+
         try {
             poetryHelper.execute(Arrays.asList("check"));
         } catch (HabushuException e) {
             getLog().debug("Failure encountered while running 'poetry check'!", e);
             getLog().warn("poetry check failed (debug contains more details) - this is likely due to a "
                     + "mismatch between your pyproject.toml and poetry.lock file - attempting to correct...");
-            poetryHelper.execute(Arrays.asList("lock", "--no-update"));
+            if (Boolean.TRUE.equals(isPoetryVersionAtLeast2)){
```
chang-annie: I: for Poetry v2.0.0+,  `--no-update` is the default for the `poetry lock` command

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -148,12 +151,24 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
 
         if (!this.skipPoetryLockUpdate) {
             getLog().info("Locking dependencies specified in pyproject.toml...");
-            poetryHelper.executePoetryCommandAndLogAfterTimeout(Arrays.asList("lock"), 2, TimeUnit.MINUTES);
+            if (Boolean.TRUE.equals(isPoetryVersionAtLeast2)){
+                poetryHelper.executePoetryCommandAndLogAfterTimeout(Arrays.asList("lock", "--regenerate"), 2, TimeUnit.MINUTES);
```
chang-annie: I: to regenerate the lock file from scratch (old default behavior), we need to add the `--regenerate` argument

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -148,12 +151,24 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
...
+            } else {
+                poetryHelper.executePoetryCommandAndLogAfterTimeout(Arrays.asList("lock"), 2, TimeUnit.MINUTES);
+            }
         }
 
         List<String> installCommand = new ArrayList<>();
 
-        installCommand.add("install");
+        if (Boolean.TRUE.equals(isPoetryVersionAtLeast2) && this.forceSync){
```
chang-annie: I: with Poetry v2.0.0.+, `poetry install --sync` deprecated and replaced with `poetry sync`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -170,7 +173,11 @@ public void execute() throws MojoExecutionException {
 
         if (usePyenv) {
             log.info("Configuring Poetry to use the pyenv-activated Python binary...");
-            poetryHelper.executeAndLogOutput(Arrays.asList("config", "--local", "virtualenvs.prefer-active-python", "true"));
+            if (Boolean.TRUE.equals(isPoetryVersionAtLeast2)) {
+                poetryHelper.executeAndLogOutput(Arrays.asList("config", "--local", "virtualenvs.use-poetry-python", "false"));
```
chang-annie: I: with Poetry v2.0.0+,  `virtualenvs.prefer-active-python` (default: `false`) has been replaced with inverse setting `virtualenvs.use-poetry-python` (default: `false`)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -239,4 +241,18 @@ protected ProcessExecutor createPoetryExecutor(List<String> arguments, Map<Strin
...
+     * Returns a {@link Boolean} indicating whether the specified Poetry version is at least 2.0.0.
+     *
+     * @param poetryVersion
+     * @return
+     */
+    public Boolean isPoetryVersionAtLeast2(String poetryVersion){
+        DefaultArtifactVersion currentVersion = new DefaultArtifactVersion(poetryVersion);
+        DefaultArtifactVersion minimumVersion = new DefaultArtifactVersion(BREAKING_POETRY_VERSION);
+
+        return currentVersion.compareTo(minimumVersion) >= 0;
```
chang-annie, line 255: I:  `compareTo` returns:

- `< 0`  if `currentVersion` is lower than `minimumVersion=2.0.0`
- `= 0 ` if they are equal
- `> 0`  if `currentVersion` is higher

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -263,7 +281,7 @@ protected void performDockerfileUpdateForVirtualEnvironment(Path targetProjectPa
         Path outputDir = dockerContext.toPath().relativize(getStagingPath());
         String updatedDockerfile =
             ContainerizeDepsDockerfileHelper.updateDockerfileWithContainerStageLogic(this.dockerfile,
```
epwilkins, line 283: A: This might be a bit cleaner if we just make this an actual object that can be configured, and then operate on Dockerfiles. Something like a `PythonContainerizer` that can be configured with the various versions and docker images. Then the actual call is something like `containerizer.updateDockerFile(this.dockerfile, outputDir.toString(), targetProjectPath.toString())`. (Names and specifics were just off the cuff so could be adjusted as you see fit.)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -239,4 +241,18 @@ protected ProcessExecutor createPoetryExecutor(List<String> arguments, Map<Strin
...
         return new ProcessExecutor(workingDirectory, fullCommandArgs, Platform.guess(), environmentVariables);
     }
+
+    /**
+     * Returns a {@link Boolean} indicating whether the specified Poetry version is at least 2.0.0.
+     *
+     * @param poetryVersion
+     * @return
+     */
+    public Boolean isPoetryVersionAtLeast2(String poetryVersion){
```
epwilkins: S: Since we aren't ever returning `null` I think we should just be using a primitive here. It would also make it much easier to use/read in `if` statements since you could use the value directly instead of comparing it to a constant.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -23,13 +23,21 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
 
         getLog().info("Validating Poetry-based project structure...");
         PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        Boolean isPoetryVersionAtLeast2 = poetryHelper.isPoetryVersionAtLeast2(
+                poetryHelper.getIsPoetryInstalledAndVersion().getRight()
```
epwilkins: Q: Is there a reason not to just have `poetryHelper.isPoetryVersionAtLeast2()` call `getIsPoetryInstalledAndVersion` directly itself instead of passing it in?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
d-ryan-ashcraft: S (Substantive): Do we want to specify these, or should they default to the latest (unspecified) unless specifically listed?  In Java, we'd lock them to a specific version like this.  But in Python, I think the "intended behavior" of versioning it to float up with the assumption that everything will be forward-compatible (even if we don't really buy that or experience it consistently in practice).

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
```diff
@@ -74,9 +86,9 @@ public static String createContainerStageContentFrom(String template, String anc
...
-        String builderStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(BUILDER_STAGE_TEMPLATE, anchorDirectory, moduleBaseDir, owner, builderBaseImage);
-        String finalStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(FINAL_STAGE_TEMPLATE, null, null, owner, finalBaseImage);
+    public static String updateDockerfileWithContainerStageLogic(File dockerFile, String anchorDirectory, String moduleBaseDir, String owner, String builderBaseImage, String finalBaseImage, String poetryVersion, String poetryMonorepoDependencyPluginVersion, String poetryPluginBundleVersion) {
+        String builderStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(BUILDER_STAGE_TEMPLATE, anchorDirectory, moduleBaseDir, owner, builderBaseImage, poetryVersion, poetryMonorepoDependencyPluginVersion, poetryPluginBundleVersion);
+        String finalStageContent = ContainerizeDepsDockerfileHelper.createContainerStageContentFrom(FINAL_STAGE_TEMPLATE, null, null, owner, finalBaseImage, null, null, null);
```
epwilkins, line 91: A: We're getting close to the breaking point where this approach doesn't make sense anymore. I'd think the Apache Velocity Engine would give us a good feature set that simplifies our own code a ton. I also think transitioning to using Velocity wouldn't be that much of a lift.  If it turns out to be, we can always make it a separate issue that we tackle on its own.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -23,13 +23,21 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
...
+
         try {
             poetryHelper.execute(Arrays.asList("check"));
         } catch (HabushuException e) {
             getLog().debug("Failure encountered while running 'poetry check'!", e);
             getLog().warn("poetry check failed (debug contains more details) - this is likely due to a "
                     + "mismatch between your pyproject.toml and poetry.lock file - attempting to correct...");
-            poetryHelper.execute(Arrays.asList("lock", "--no-update"));
+            if (Boolean.TRUE.equals(isPoetryVersionAtLeast2)){
+                poetryHelper.execute(Arrays.asList("lock"));
```
epwilkins: A: We might consider creating methods on `PoetryHelper` for the specific commands. So for example, we could have `PoetryHelper.lock(boolean update)` that checks the version and chooses the correct format/args. If `PoetryHelper` can't guarantee it can know what version is being targeted by the method call, we could add the version as a parameter.

An upshot of this approach is that I think we could then write some unit tests around PoetryHelper to make sure we're consistently and correctly accounting for version differences. We'd just need to extend it to intercept the `execute` call and match the passed array to the command/args we expect.

epwilkins: I'm noticing the complexity with `LogAfterTimeout` etc.  Another approach would be to have a `createLockCommand` method that just returns the proper List that could then be passed into `execute`/`executePoetryCommandAndLogAfterTimeout`/etc.

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
d-ryan-ashcraft: This is fair.  Since we have a breaking change, I'd favor tackling this in 3.0.0.  I also wonder if we could combine Velocity with some docker containers we build to simplify the resulting Dockerfiles?

d-ryan-ashcraft (approved):

meliz19 (approved):
great work!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
epwilkins: It's a bit tricky since these are Docker containers and I think in general the Docker community is moving towards more reproducible builds.  If we think it makes more sense to favor the Python approach (especially given that Docker builds just have inherent reproducibility issues) I could buy that. However, my gut reaction would be that since these aren't libraries of the packaged project and instead just build tooling that moving away from the Python standard would be fine.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
chang-annie: @ewilkins-csi updated `isPoetryVersionAtLeast2` to call `getIsPoetryInstalledAndVersion` directly and updated to primitive

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
chang-annie: @ewilkins-csi i updated poetry helper to include `createLockCommand`, `createInstallCommand`, `createUsePyenvCommand` helper methods that check the poetry versions and return the proper Lists of commands to be run. i'll look into adding/updating unit tests now!

chang-annie: @ewilkins-csi created unit tests for the new version-dependent command builder functions in `PoetryCommandHelper`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
chang-annie: created [Issue #207](https://github.com/TechnologyBrewery/habushu/issues/207l) to handle the changes to the `containerize-dependencies` mojo mentioned above

Issue #200, opened 2025-02-04
Habushu: add ability to switch between poetry and uv configuration


jaebchoi:
## Description
This work is to support UV in the habushu.
Before we conduct series of MOJOs for UV, it would be great idea to support ability to switch between uv and poetry configuration.

Once Utils for switch between uv and poetry configuration, add it to validate-python-package-and-dependency-manager only for this iteration.

For switching between uv vs poetry configuration, Pyproject toml have specific template for UV and poetry see migration guide [here](https://www.loopwerk.io/articles/2024/migrate-poetry-to-uv/) 

Since UV doesn't have build-backend to differentiate this toml file is UV based while poetry does. 
So, we will assume if toml file is not poetry, it would be uv.

## DoD

- [x] Figure out way to differentiate uv and poetry (using build backend for pyproject.toml)
- [x] habushu plugin for poetry can still function as before.
- [x] implement PackageManager Check and apply it to validate-python-package-and-dependency-manager MOJO

## Testing Step
- Pull 3.0.0-add-uv-support branch
- On the root of habushu project, run `mvn clean install -Pbootstrap` to build habushu maven plugin
- run `mvn clean install -Pdefault` to build all modules that uses poetry
- run `mvn clean install --Puv` to build module and fail with UvSetup not yet implemented Error

meliz19:
Completed DoD & OTS with @meliz19

jaebchoi:
Re-DOD/OTS Completed @d-ryan-ashcraft

d-ryan-ashcraft:
Test passed (fails, as intended):
```
[INFO] 
[INFO] --- habushu:2.18.0-SNAPSHOT:clean-habushu (default-clean-habushu) @ habushu-mixology-uv ---
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 2.18.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.126 s]
[INFO] habushu::Mixing Drinks with Habushu ................ FAILURE [  0.176 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.316 s (Wall Clock)
[INFO] Finished at: 2025-02-11T15:22:48-05:00
[INFO] ------------------------------------------------------------------------
[...]

PR #201, opened 2025-02-05
Initial refactoring of tool setup to prepare for uv integration

meliz19:
Refactored `PyenvAndPoetrySetup` such that `PythonPackageAndDependencyManagerSetup` can select between poetry or uv; uv functionality tbd

PR #204, opened 2025-02-06
#200 Habushu: add ability to switch between poetry and uv configuration 

jaebchoi:
https://github.com/TechnologyBrewery/habushu/issues/200


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -84,68 +85,71 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
-            getLog().info("-------------------------------------------------------");
-            getLog().info("T E S T S");
-            getLog().info("-------------------------------------------------------");
-            poetryHelper.executeAndLogOutput(executeBehaveTestArgs, behaveTestEnvironmentVariables);
-        } else {
-            getLog().warn(String.format("No tests found in %s", getCanonicalPathForFile(behaveDirectory)));
-        }
-
+      if(HabushuUtil.currentPackageManager() == HabushuUtil.PackageManager.POETRY)
+      {
```
jaebchoi: I: Most of MOJO change in this PR deals with adding current package manager check.

pom.xml
```diff
@@ -120,6 +120,24 @@
...
+                <module>habushu-mixology-uv</module>
+                <module>habushu-mixology-consumer-uv</module>
+            </modules>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-install-plugin</artifactId>
+                        <configuration>
+                            <skip>true</skip>
```
jaebchoi, line 135: I: This is intentional as UV implementation for most of MOJO is noop so nothing to install.
To get success build, we need to skip install for now.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -204,4 +211,39 @@ public static String getCleanVirtualEnvironmentPath(String virtualEnvFullPath) {
...
+	 */
+	public static PackageManager checkPythonPackageManager(File pyProjectTomlFile) {
+
+		try (BufferedReader reader = new BufferedReader(new FileReader(pyProjectTomlFile))) {
+			String line = reader.readLine();
+
+			while (line != null) {
+				line = line.strip();
+				if (line.contains("[tool.poetry")) {
+					currentPackageManager = PackageManager.POETRY;
```
d-ryan-ashcraft: A (Administrative): This is minor, but we can probably `break` after making the call.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -25,6 +26,12 @@
  */
 public final class HabushuUtil {
 
+	public enum PackageManager {
+		POETRY,
+		UV
+	}
+
+	private static PackageManager currentPackageManager = PackageManager.UV;
```
d-ryan-ashcraft: S (Substantive): We want to keep the code thread-safe for `mvnd` use.  As a result, I'd move this to inside the `checkPythonPackageManager(...)` call.

habushu-mixology-uv/tests/features/environment-variables.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Environment variables are configured during behave execution through python-dotenv
```
d-ryan-ashcraft: S: These tests seem heavy for Habushu.  What are we trying to accomplish here?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -84,68 +85,71 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
-
-        boolean hasTests;
-        try {
-            hasTests = behaveDirectory.exists() && Files.list(behaveDirectory.toPath()).findAny().isPresent();
-        } catch (IOException e) {
-            throw new MojoExecutionException("Could not load behave features directory", e);
-        }
-
-        if (hasTests) {
-            PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
```
meliz19, line 103: S (suggestion): Both poetry and uv run the behave command via `<run behave <behave commands>`, so I think it might flow better if we abstract the code block in lines 103 - 108 to call the poetry or uv command based on this logic `if(HabushuUtil.currentPackageManager() == HabushuUtil.PackageManager.POETRY)`.  The code after line 108 will be the same whether poetry or uv is used.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -84,68 +85,71 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
-
-        if (skipTests) {
-            getLog().warn("Tests are skipped (-DskipTests=true)");
-            return;
-        }
-
-        File behaveDirectory = new File(testDirectory, "features");
```
meliz19, line 93: I (informational): Note for awareness/future implementation: Unfortunately, `uv` doesn't automatically add a test/features directory for users, so we'll have to include that feature in habushu's uv functionality.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -110,11 +110,17 @@ public class CleanHabushuMojo extends CleanMojo {
 
     @Override
     public void execute() throws MojoExecutionException {
-        if ("habushu".equals(packaging)) {
```
meliz19, line 121: This code block is agnostic to whether poetry or uv is used. I don't think this change is necessary.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
```diff
@@ -22,32 +23,37 @@ public class FormatPythonMojo extends AbstractHabushuMojo {
...
+		if (this.testDirectory.exists()) {
+			directoriesToFormat.add(getCanonicalPathForFile(testDirectory));
+		}
+
+		if (directoriesToFormat.isEmpty()) {
+			getLog().warn(String.format("Neither configured source (%s) nor test (%s) directories exist - skipping...",
+					sourceDirectory, testDirectory));
+		}
+
+		PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
```
meliz19, line 40: Lines 41 - 47 will need to have the `if(HabushuUtil.currentPackageManager() == HabushuUtil.PackageManager.POETRY)` logic, but the rest of this code should remain the same since we're running black via the `run` command, which is the same for both `poetry` and `uv`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
```diff
@@ -22,32 +23,37 @@ public class FormatPythonMojo extends AbstractHabushuMojo {
...
+					String.format("%s dependency not specified in pyproject.toml - installing now...", BLACK_PACKAGE));
+			poetryHelper.installDevelopmentDependency(BLACK_PACKAGE);
+		}
+
+		List<String> executeBlackFormatterArgs = new ArrayList<>();
+		executeBlackFormatterArgs.addAll(Arrays.asList("run", BLACK_PACKAGE));
+		executeBlackFormatterArgs.addAll(directoriesToFormat);
+
+		getLog().info("Formatting configured source and test directories using black...");
+		poetryHelper.executeAndLogOutput(executeBlackFormatterArgs);
```
meliz19, line 53: Line 54 will need to be converted to an agnostic function with the `if(HabushuUtil.currentPackageManager() == HabushuUtil.PackageManager.POETRY)` logic under the hood.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoMojo.java
```diff
@@ -126,36 +127,39 @@ public class PublishToPyPiRepoMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
-        if (this.skipDeploy) {
```
meliz19, line 129: S: Lines 129 - 134 can remain as is. No need to place it under the if/else logic

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
```diff
@@ -32,12 +33,18 @@ public class RunCommandInVirtualEnvMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
-	PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
```
meliz19, line 35: The run command is the same for both poetry and uv. I'd recommend we abstract the parameters out such that poetry isn't in the name. E.g., `runCommandArgs`. You can add the if/else logic to determine if the log statement says poetry or uv.

habushu-mixology-uv/tests/features/environment-variables.feature
jaebchoi: This is duplicate of habushu-mixology only change in uv version is deleting poetry.toml and pyproject.toml conforming uv. 
Given that uv is noop at the moment, I think we can strip out the test and revisit

jaebchoi: also we might see pyenv support in uv roadmap as discussed in [here](https://github.com/TechnologyBrewery/habushu/pull/201#discussion_r1945104678) so might needed to make sure pyenv can set env variable?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
jaebchoi: Make sense, updated accordingly!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
jaebchoi: I just return at that point to make thing more readable!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
jaebchoi: yup understood, I think we should flow that in later ticket when we implement BehaveBDDTestMoJo for UV because all UV impl is noop at the moment, so test will fail if we run common code

jaebchoi: but definitely good information which I can add in the comments.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
jaebchoi: same as above, can update comments and keep noop

Issue #198, opened 2025-02-03
Add support for uv - Refactor setup mojo to account for uv or poetry


meliz19:
## Description
There is a demand signal to be able to use [uv](https://docs.astral.sh/uv/) as a Python package and dependency management tool within the context of the `habushu-maven-plugin`. The purpose of this ticket is to develop a script to check the `uv` version, similar to the current `PyenvAndPoetrySetup.java` script.

### Update - 2/6/25
As I began working on this ticket, I discovered there is a good bit of refactoring necessary. I am going to split this ticket into two parts. This ticket will focus on refactoring the set up mojo of habushu such that the shared set up functionality is in the `PythonPackageAndDependencyManagerSetup` class and the poetry specific code remains in the `PyenvAndPoetrySetup` class. The next ticket will focus on adding the uv specific code in the `UvSetup` and `PythonPackageAndDependencyManagerSetup` classes.

## DOD
- ~[ ] Develop a script to check a user's installed version of `uv`~
- Refactor the `PyenvAndPoetrySetup` class such that the shared set up functionality is moved to the `PythonPackageAndDependencyManagerSetup` class and the poetry specific code remains in the `PyenvAndPoetrySetup` class

## Test Strategy/Script
- pull down the `3.0.0-add-uv-support` feature branch
- Follow the building habushu section of the README (i.e., run `mvn clean install -Pbootstrap` and then `mvn clean install -Pdefault` in the root project directory)

meliz19:
Completed DoD with @jaebchoi & @cpointe-ibllanos

meliz19:
Completed re-DoD & OTS with @d-ryan-ashcraft

cwoods-cpointe:
Final test: Build completes successfully

PR #205, opened 2025-02-06
#198 Initial refactoring of tool setup to prepare for uv integration

meliz19:
Refactored `PyenvAndPoetrySetup` such that `PythonPackageAndDependencyManagerSetup` can select between poetry or uv; uv functionality tbd

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
@@ -1,0 +1,123 @@
+package org.technologybrewery.habushu;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.logging.Log;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Abstract class that ensures pre-requisite tools that Habushu leverages are installed and available on the
+ * developer's machine to support the same functionality across multiple Mojo implementations.
+ */
+public abstract class AbstractPythonPackageAndDependencyManagerSetup {
+
+    private static final ThreadLocal<ValidationTrackingStatus> validationStatusContainer = ThreadLocal.withInitial(ValidationTrackingStatus::new);
+    public static final String VALIDATED_IN_PRIOR_BUILD_PHASE = " (validated in prior build phase)";
+
+    /**
+     * The desired version of Python to use.
+     */
+    protected String pythonVersion;
+
+    /**
+     * Base directory from which to write Python package and dependency management files.
+     */
+    protected File baseDir;
+
+    /**
+     * Indicates whether Habushu should leverage the
+     * {@code poetry-monorepo-dependency-plugin} or the
+     * {@code uv-monorepo-dependency-plugin} (<- todo) to rewrite any local path
+     * dependencies (to other Poetry/uv projects) as versioned packaged dependencies in
+     * generated wheel/sdist archives. If {@code true}, Habushu will replace
+     * invocations of Poetry/uv's {@code build} and {@code publish} commands in the
+     * {@link BuildDeploymentArtifactsMojo} and {@link PublishToPyPiRepoMojo} with
+     * the extensions of those commands exposed by the
+     * {@code poetry-monorepo-dependency-plugin}/{@code uv-monorepo-dependency-plugin}, which are
+     * {@code build-rewrite-path-deps} and {@code publish-rewrite-path-deps}
+     * respectively.
+     * <p>
+     * Typically, this flag will only be {@code true} when deploying/releasing
+     * Habushu modules within a CI environment that are part of a monorepo project
+     * structure in which multiple Poetry/uv projects depend on one another.
+     */
+    protected boolean rewriteLocalPathDepsInArchives;
+
+    /**
+     * Logger from calling class to leverage.
+     */
+    protected Log log;
+
+    /**
+     * New instance - these values are typically passed in from Maven-enabled parameters in the calling Mojo.
+     *
+     * @param pythonVersion                  version of python to leverage
+     * @param baseDir                        base directory from which to operate for this module
+     * @param rewriteLocalPathDepsInArchives see member variable for details
+     * @param log                            the logger to use for output
+     */
+    public AbstractPythonPackageAndDependencyManagerSetup(String pythonVersion, File baseDir, boolean rewriteLocalPathDepsInArchives, Log log) {
+        this.pythonVersion = pythonVersion;
+        this.baseDir = baseDir;
+        this.rewriteLocalPathDepsInArchives = rewriteLocalPathDepsInArchives;
+        this.log = log;
+    }
+
+    public void execute() throws MojoExecutionException {
+        List<String> missingRequiredToolMsgs = new ArrayList<>();
+        String currentPythonVersion;
+
+        ValidationTrackingStatus validationTracker = validationStatusContainer.get();
+        String ActivePythonVersion = validationTracker.getActivePythonVersion();
+
+        if (pythonVersion.equals(ActivePythonVersion)) {
+            log.info("Using Python version: " + ActivePythonVersion + VALIDATED_IN_PRIOR_BUILD_PHASE);
+
+        } else {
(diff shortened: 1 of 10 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PythonPackageAndDependencyManagerSetup.java
```diff
@@ -0,0 +1,191 @@
...
+        String currentPythonVersion;
+
+        ValidationTrackingStatus validationTracker = validationStatusContainer.get();
+        String priorActivePythonVersionActivated = validationTracker.getPriorActivePythonVersionActivated();
+
+        if (pythonVersion.equals(priorActivePythonVersionActivated)) {
+            log.info("Using Python version: " + priorActivePythonVersionActivated + VALIDATED_IN_PRIOR_BUILD_PHASE);
+
+        } else {
+            if (pythonPackageAndDependencyManager.equals("poetry")) {
```
epwilkins: S: Instead of `if` statements, I'd favor using inheritance. This would make it much easier to extend in the future. For instance. `PyenvAndPoetrySetup` could be a subclass to `PythonPackageAndDependencyManagerSetup` where `configurePython` is an abstract method in the superclass that calls (or inlines) `configurePyenvOrStraightPython` in the subclass. We could have similar abstract methods everywhere else that poetry/uv differ.

That way, to implement a new Package/Dep mgr, you simply extend this class and implement the abstract methods.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonPackageAndDependencyManagerMojo.java
```diff
@@ -8,28 +8,41 @@
...
-    @Parameter(defaultValue = PyenvAndPoetrySetup.PYTHON_DEFAULT_VERSION_REQUIREMENT, property = "habushu.pythonVersion")
+    @Parameter(defaultValue = PythonPackageAndDependencyManagerSetup.PYTHON_DEFAULT_VERSION_REQUIREMENT, property = "habushu.pythonVersion")
     protected String pythonVersion;
 
+    /**
+     * The desired Python package and dependency manger to use.
+     * todo: update later when poetry vs uv detection functionality is available.
+     */
+    @Parameter(defaultValue = "poetry", property = "habushu.pythonPackageAndDependencyManager")
+    protected String pythonPackageAndDependencyManager;
```
epwilkins, line 44: A: Building on the former comment about using inheritance for this functionality, it might be interesting to allow pluggable implementations here by specifying the fully-qualified class name. That would allow someone to extend Habushu with their own implementation of the backing build tooling.

The downside is that it becomes harder to choose between the out-of-the-box options (poetry v. uv). I'm honestly not too familiar with a good pattern for this in the Maven plugin space, and it may be enough of a barrier that we just keep this as food for thought.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidationTrackingStatus.java
```diff
@@ -25,6 +27,22 @@ public void setAlreadyValidatedPoetryVersion(String alreadyValidatedPoetryVersio
         this.alreadyValidatedPoetryVersion = alreadyValidatedPoetryVersion;
     }
 
+    public boolean isAlreadyValidatedUvInstallation() {
```
epwilkins, line 30: A: We lost the comments on this, but I think it's OK for us to be opinionated about mixing build platforms/tooling and only support _either_ uv **or** poetry and not mixing.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -175,21 +172,19 @@ public static String getInProjectVirtualEnvironmentPath(File workingDirectory) {
...
-		configureTools.execute();
 
-		try {
-			PoetryCommandHelper poetryHelper = new PoetryCommandHelper(workingDirectory);
-			virtualEnvFullPath = poetryHelper.execute(Arrays.asList("env", "list", "--full-path"));
-		} catch (RuntimeException e) {
-			log.debug("Could not retrieve Poetry-managed virtual environment path - it likely does not exist",
-					e);
-		}
+		if (pythonPackageAndDependencyManager.equals("poetry")) {
```
epwilkins: A/S: tabs -> space, but also this stands out as maybe another instance polymorphism/inheritance would simplify our code.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/PoetryUtil.java
```diff
@@ -15,4 +23,24 @@ public final class PoetryUtil {
      * must be installed and available for Habushu to use.
      */
     public static final String POETRY_CORE_VERSION_REQUIREMENT = "^1.6.0";
+
+
+    public static String findCurrentVirtualEnvironmentFullPathForPoetry(String pythonVersion, String pythonPackageAndDependencyManager,
```
epwilkins: A: tabs -> spaces for the method, and potentially
```suggestion
    public static String findVirtualEnvironmentAbsolutePath(String pythonVersion, String pythonPackageAndDependencyManager,
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/UvUtil.java
```diff
@@ -0,0 +1,36 @@
...
+ * Common utility methods for handling TOML uv.
+ */
+public final class UvUtil {
+    /**
+     * Specifies the semver compliant requirement for the version of uv that
+     * must be installed and available for Habushu to use.
+     */
+    public static final String UV_VERSION_REQUIREMENT = ">=0.5.0";
+
+    public static String findCurrentVirtualEnvironmentFullPathForUv(String pythonVersion, String pythonPackageAndDependencyManager,
```
epwilkins: A: If you like the rename of the poetry method, would make this match:
```suggestion
    public static String findVirtualEnvironmentAbsolutePath(String pythonVersion, String pythonPackageAndDependencyManager,
```

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidationTrackingStatus.java
d-ryan-ashcraft: For a given module/subproject, you can be `uv` or `poetry`.  In an overall project, you can mix and match them as you see fit.  But each habushu invocation should be one and only one for that module.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
```diff
@@ -0,0 +1,123 @@
+package org.technologybrewery.habushu;
```
meliz19, line 1: I: Added abstract class based off @ewilkins-csi's feedback

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -18,64 +18,20 @@
 import java.util.List;
```
meliz19, line 18: I: Refactored to be subclass of `AbstractPythonPackageAndDependencyManagerSetup`

d-ryan-ashcraft (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PythonPackageAndDependencyManagerFactory.java
```diff
@@ -0,0 +1,23 @@
+package org.technologybrewery.habushu;
+
+import java.io.File;
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.logging.Log;
+import org.technologybrewery.habushu.util.HabushuUtil;
+
+public class PythonPackageAndDependencyManagerFactory {
+    public static AbstractPythonPackageAndDependencyManagerSetup createPythonPackageAndDependencyManagerSetup(
```
epwilkins, line 9: A: Since this is a factory, you don't really know the type of Setup class you'll get back or what it will need in `additionalParameters`. It will likely be that we'll just have to pass in the kitchen sink, and the factory will sort what it needs/doesn't based on the selected Manager. For that reason, I think we might as well make them explicit named parameters.

PR #206, opened 2025-02-06
 #200 Habushu: add ability to switch between poetry and uv configuration

jaebchoi:
https://github.com/TechnologyBrewery/habushu/issues/200


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -84,68 +85,70 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
-            getLog().info(String.format("Executing behave tests in %s...", getCanonicalPathForFile(behaveDirectory)));
-            getLog().info("-------------------------------------------------------");
-            getLog().info("T E S T S");
-            getLog().info("-------------------------------------------------------");
-            poetryHelper.executeAndLogOutput(executeBehaveTestArgs, behaveTestEnvironmentVariables);
-        } else {
-            getLog().warn(String.format("No tests found in %s", getCanonicalPathForFile(behaveDirectory)));
-        }
-
+      if(HabushuUtil.checkPythonPackageManager(getPoetryPyProjectTomlFile()) == HabushuUtil.PackageManager.POETRY) {
```
jaebchoi: I: Most of MOJO change in this PR deals with adding current package manager check.

pom.xml
```diff
@@ -120,6 +120,24 @@
...
+                <module>habushu-mixology-consumer-uv</module>
+            </modules>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-install-plugin</artifactId>
+                        <configuration>
+                            <skip>true</skip>
+                        </configuration>
```
jaebchoi: I: This is intentional as UV implementation for most of MOJO is noop so nothing to install.
To get success build, we need to skip install for now.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -204,4 +211,31 @@ public static String getCleanVirtualEnvironmentPath(String virtualEnvFullPath) {
 		return StringUtils.replace(virtualEnvFullPath, " (Activated)", StringUtils.EMPTY);
 
 	}
+
+	/**
+	 * Finds and returns Python Package Manager.
+	 *
+	 * @return PackageManager returns which package manager does Habushu uses.
```
carter-cundiff:
```suggestion
	 * @return PackageManager returns which package manager Habushu uses.
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
carter-cundiff: S: If this method `getPoetryPyProjectTomlFile()` is strictly for returning the `pyproject.toml`, probably worth refactoring it to a more generic name: `getPyProjectTomlFile()`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -21,36 +22,39 @@ public class InitializeHabushuMojo extends AbstractHabushuMojo {
...
+                    poetryHelper.executeAndLogOutput(Arrays.asList("version", expectedPythonPackageVersion));
+                } else {
+                    getLog().debug(String.format(
+                            "Poetry package version set to %s in pyproject.toml does not align with expected POM-derived version of %s",
+                            currentPythonPackageVersion, expectedPythonPackageVersion));
+                }
 
+            }
+        } else{
+            getLog().info("Validating uv-based project structure...");
```
carter-cundiff: A: Would leave a `//TODO: UV Impl` so we don't accidently miss it going back through.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
```diff
@@ -32,12 +33,17 @@ public class RunCommandInVirtualEnvMojo extends AbstractHabushuMojo {
...
+        poetryRunCommandArgs.add(0, "run");
 
-	getLog().info("Executing command in virtual environment via 'poetry run'...");
-	poetryHelper.executeAndLogOutput(poetryRunCommandArgs);
+        getLog().info("Executing command in virtual environment via 'poetry run'...");
+        poetryHelper.executeAndLogOutput(poetryRunCommandArgs);
+    }
+    else{
+        //TODO: UV impl
+    }
```
carter-cundiff: S: This all needs to be tabbed over one indent

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePyenvAndPoetryMojo.java
```diff
@@ -8,28 +8,42 @@
 import org.apache.maven.plugins.annotations.Mojo;
```
carter-cundiff: Coordinate these file changes with #205 since both PRs are touching the same file.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -204,4 +211,31 @@ public static String getCleanVirtualEnvironmentPath(String virtualEnvFullPath) {
...
+	 */
+	public static PackageManager checkPythonPackageManager(File pyProjectTomlFile) {
+		PackageManager currentPackageManager = PackageManager.UV;
```
carter-cundiff: A: This can be removed as you don't end up returning the variable

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
jaebchoi: good call, updated!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePyenvAndPoetryMojo.java
jaebchoi: got it, had mixed some commit with #205, fixed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
```diff
@@ -342,14 +342,14 @@ protected PyenvCommandHelper createPyenvCommandHelper() {
      * @return
      */
     protected PoetryCommandHelper createPoetryCommandHelper() {
-        return new PoetryCommandHelper(getPoetryProjectBaseDir());
+        return new PoetryCommandHelper(getProjectBaseDir());
     }
 
     /**
      * Base directory in which Poetry projects will be located - should always be
```
carter-cundiff:
```suggestion
     * Base directory in which Python projects will be located - should always be
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
```diff
@@ -359,8 +359,8 @@ protected File getPoetryProjectBaseDir() {
      *
```
carter-cundiff, line 380: Update comment to remove Poetry

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
```diff
@@ -22,32 +23,36 @@ public class FormatPythonMojo extends AbstractHabushuMojo {
     @Override
     public void doExecute() throws MojoExecutionException {
 
-	List<String> directoriesToFormat = new ArrayList<>();
-	if (this.sourceDirectory.exists()) {
-	    directoriesToFormat.add(getCanonicalPathForFile(sourceDirectory));
+    if(HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.POETRY) {
```
carter-cundiff: S: Need to tab `if` statement over one

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
carter-cundiff: Still needs to be tabbed over

jaebchoi: Sorry If I am mistaken
so should it be something like this?
```
        getLog().info("Executing command in virtual environment via 'poetry run'...");
        poetryHelper.executeAndLogOutput(poetryRunCommandArgs);
    } else{
        //TODO: UV impl
    }
```
or 

```
        getLog().info("Executing command in virtual environment via 'poetry run'...");
        poetryHelper.executeAndLogOutput(poetryRunCommandArgs);
    } else{
      //TODO: UV impl
    }
```
align with else statement? or above code?

carter-cundiff: These should not be inline:
```
public void doExecute() {
if(HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.POETRY) {

}
else {

}
}
```
Should be:
```
public void doExecute() {
    if(HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.POETRY) {

    }
    else {

    }
}
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -84,68 +85,70 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
-            getLog().info(String.format("Executing behave tests in %s...", getCanonicalPathForFile(behaveDirectory)));
-            getLog().info("-------------------------------------------------------");
-            getLog().info("T E S T S");
-            getLog().info("-------------------------------------------------------");
-            poetryHelper.executeAndLogOutput(executeBehaveTestArgs, behaveTestEnvironmentVariables);
-        } else {
-            getLog().warn(String.format("No tests found in %s", getCanonicalPathForFile(behaveDirectory)));
-        }
-
+      if(HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.POETRY) {
```
meliz19: See my comment from the old PR.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -204,4 +211,30 @@ public static String getCleanVirtualEnvironmentPath(String virtualEnvFullPath) {
...
+	 */
+	public static PackageManager checkPythonPackageManager(File pyProjectTomlFile) {
+		try (BufferedReader reader = new BufferedReader(new FileReader(pyProjectTomlFile))) {
+			String line = reader.readLine();
+
+			while (line != null) {
+				line = line.strip();
+				//if pyproject.toml contains poetry specific section, the package manager is poetry.
+				//if not, it would direct to uv.
+				if (line.contains("[tool.poetry")) {
```
d-ryan-ashcraft: S (Substantive): I'm wondering if this is the best way to determine things for Poetry.  I might think it would be looking for the `build-backend` starts with 'poetry` instead?  Might be easier to use NightOwl or a similar tool for that.  https://python-poetry.org/docs/pyproject/#poetry-and-pep-517

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -68,46 +69,50 @@ public class BuildDeploymentArtifactsMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
-        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
```
meliz19, line 71: @ewilkins-csi suggested we take advantage of inheritance via abstract classes instead of using if/else statements. I'd wait to refactor until I merge in #205 though.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
jaebchoi: Ahh, make sense now, ill update them!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -110,11 +110,16 @@ public class CleanHabushuMojo extends CleanMojo {
 
     @Override
     public void execute() throws MojoExecutionException {
-        if ("habushu".equals(packaging)) {
```
meliz19: this functionality shouldn't change between poetry and uv

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
jaebchoi: yeap understood, I put kind of placeholder until we decide which mojo we use abstract class vs just plain check. I can def wait for 205 to refactor!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
```diff
@@ -22,32 +23,36 @@ public class FormatPythonMojo extends AbstractHabushuMojo {
...
+		if (this.testDirectory.exists()) {
+			directoriesToFormat.add(getCanonicalPathForFile(testDirectory));
+		}
+
+		if (directoriesToFormat.isEmpty()) {
+			getLog().warn(String.format("Neither configured source (%s) nor test (%s) directories exist - skipping...",
+					sourceDirectory, testDirectory));
+		}
+
+		PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
```
meliz19: - echoing earlier comment re: refactor to use abstract class instead of if/else.
- Both poetry and uv use the `run` command to execute tools (e.g., behave or black), so I wonder if we could abstract lines 40 - 45 into an abstract method.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
chang-annie: ^ agreed! the `[tool.poetry]` line is not reliable because with Poetry v2.0.0, they've moved to using a standard `project` configuration https://python-poetry.org/blog/announcing-poetry-2.0.0#supporting-the-project-section-in-pyprojecttoml-pep-621

jaebchoi: sounds good, updated to use toml4j and uses build-backend to classify package manager.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
jaebchoi: updated to abstract out command helper and uses that!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
```diff
@@ -284,15 +286,15 @@ protected String findPlaintextPasswordForServer(String repoId) {
...
      * @return the poetry cache directory path as a FILE object.  
      */
     public File getCachedWheelDirectory(String artifactId) {
         try {
-            PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
-            String poetryCacheDirectoryPath = poetryHelper.getPoetryCacheDirectoryPath();
-            return new File(String.format("%s/cache/repositories/wheels/%s", poetryCacheDirectoryPath, artifactId));
+            PackageManagerCommandHelper packageManagerCommandHelper = createPackageManagerCommandHelper();
+            String packageManagerCacheDirectoryPath = packageManagerCommandHelper.getPackageManagerCacheDirectoryPath();
+            return new File(String.format("%s/cache/repositories/wheels/%s", packageManagerCacheDirectoryPath, artifactId));
```
meliz19: This path is specific to Poetry. Since this is an abstract class anyways, you could make this an abstract method and create a poetry-specific subclass and implement this logic there.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -7,7 +7,11 @@
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.habushu.exec.PackageManagerCommandHelper;
 import org.technologybrewery.habushu.exec.PoetryCommandHelper;
```
meliz19: You can remove this package as it is no longer used.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -86,67 +91,71 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
-            getLog().warn(String.format("No tests found in %s", getCanonicalPathForFile(behaveDirectory)));
-        }
-
+       if (skipTests) {
+           getLog().warn("Tests are skipped (-DskipTests=true)");
+           return;
+       }
+
+      //TODO: Skipping since UV impl is no op. Remove once UV Impl is done.
+      if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.UV) {
```
meliz19: It looks like you're using an abstract class with a factory to generate the `createPackageManagerCommandHelper`, so is this `if` statement necessary?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -5,7 +5,9 @@
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
+import org.technologybrewery.habushu.exec.PackageManagerCommandHelper;
 import org.technologybrewery.habushu.exec.PoetryCommandHelper;
```
meliz19: remove

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -68,7 +70,12 @@ public class BuildDeploymentArtifactsMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
-        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        //TODO: Delete Once UV Impl is finished..
+        if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.UV) {
```
meliz19: Is this `if` statement necessary since you're using an abstract method with a factory for `createPackageManagerCommandHelper`?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -9,6 +9,8 @@
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.clean.CleanMojo;
 import org.apache.maven.plugins.clean.Fileset;
+import org.technologybrewery.habushu.exec.PackageManagerCommandHelper;
+import org.technologybrewery.habushu.exec.PackageManagerCommandHelperFactory;
 import org.technologybrewery.habushu.exec.PoetryCommandHelper;
```
meliz19: remove

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -119,6 +121,11 @@ public class CleanHabushuMojo extends CleanMojo {
 
     @Override
     public void execute() throws MojoExecutionException {
+        //TODO: Skipping since UV impl is no op. Remove once UV Impl is done.
```
meliz19: remove `if` statement since you're using an abstract method with a factory for `createPackageManagerCommandHelper`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
```diff
@@ -8,7 +8,9 @@
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.technologybrewery.habushu.exec.PackageManagerCommandHelper;
 import org.technologybrewery.habushu.exec.PoetryCommandHelper;
```
meliz19: remove

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
```diff
@@ -22,33 +24,37 @@ public class FormatPythonMojo extends AbstractHabushuMojo {
...
     public void doExecute() throws MojoExecutionException {
-
-	List<String> directoriesToFormat = new ArrayList<>();
-	if (this.sourceDirectory.exists()) {
-	    directoriesToFormat.add(getCanonicalPathForFile(sourceDirectory));
-	}
-	if (this.testDirectory.exists()) {
-	    directoriesToFormat.add(getCanonicalPathForFile(testDirectory));
+		//TODO: Delete Once UV Impl is finished..
+		if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.UV) {
```
meliz19: Is this `if` statement necessary since you're using an abstract method with a factory for `createPackageManagerCommandHelper`?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -7,7 +7,9 @@
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
+import org.technologybrewery.habushu.exec.PackageManagerCommandHelper;
 import org.technologybrewery.habushu.exec.PoetryCommandHelper;
```
meliz19: remove

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -20,37 +22,40 @@ public class InitializeHabushuMojo extends AbstractHabushuMojo {
...
-            } else {
-                getLog().debug(String.format(
-                        "Poetry package version set to %s in pyproject.toml does not align with expected POM-derived version of %s",
-                        currentPythonPackageVersion, expectedPythonPackageVersion));
-            }
-
-        }
-
+      //TODO: Delete Once UV Impl is finished..
+      if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.UV) {
```
meliz19: same as above.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -86,67 +91,71 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
+          return;
```
cpointe-ibllanos: a: Might be a good place to add a info/debug/warning log instead of a comment for now.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
```diff
@@ -68,7 +70,12 @@ public class BuildDeploymentArtifactsMojo extends AbstractHabushuMojo {
...
+            return;
```
cpointe-ibllanos: Same as above

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -20,37 +22,40 @@ public class InitializeHabushuMojo extends AbstractHabushuMojo {
...
+         return;
+      }
+
+      getLog().info("Validating Poetry-based project structure...");
+      PackageManagerCommandHelper packageManagerCommandHelper = createPackageManagerCommandHelper();
+      try {
```
meliz19, line 26: Lines 32 - 39 should be moved into a Poetry specific method. I'd create an abstract method for this and call that here instead.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
jaebchoi: good call!

habushu-mixology-consumer-uv/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+                    <managedDependencies>
+                        <packageDefinition>
+                            <packageName>black</packageName>
+                            <operatorAndVersion>>=24.3.0</operatorAndVersion>
+                            <!-- optional - active is true by default:-->
+                            <active>true</active>
+                        </packageDefinition>
+                        <packageDefinition>
+                            <packageName>uvicorn</packageName>
+                            <operatorAndVersion>^0.18.0</operatorAndVersion>
```
cpointe-ibllanos: a: Might be good to introduce these versions as properties in the parent pom since they are referenced more than once.

habushu-mixology-consumer-uv/pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[[tool.pdm.source]]
+name = "dev-pypi"
+url = "https://test.pypi.org/simple/"
+verify_ssl = true
+
+[project]
+authors = [
+    {name = "Eric Konieczny", email = "ekoniec1@gmail.com"},
```
cpointe-ibllanos: I think it's only fair to add yourself here :)

habushu-mixology-uv/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+                    <extensions>true</extensions>
+                    <configuration>
+                        <pythonVersion>${version.python.default}</pythonVersion>
+                        <!-- Skips any scenarios tagged with @integration-test - these will
+                            be executed via the "integration-test" Maven profile -->
+                        <behaveOptions>--tags ~integration_test</behaveOptions>
+                        <managedDependencies>
+                            <packageDefinition>
+                                <packageName>black</packageName>
+                                <operatorAndVersion>^23.3.0</operatorAndVersion>
```
cpointe-ibllanos: a: same as above -- set these as a property in the root pom

habushu-mixology-uv/tests/features/environment-variables.feature
cpointe-ibllanos, line 1: S: Would note this in a follow-up ticket/issue responsible for implementing it just to be prudent

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -20,37 +22,41 @@ public class InitializeHabushuMojo extends AbstractHabushuMojo {
...
+
+      if (!StringUtils.equals(currentPythonPackageVersion, expectedPythonPackageVersion)) {
+          if (overridePackageVersion) {
+              getLog().info(String.format("Setting Package Manager package version to %s", expectedPythonPackageVersion));
+              getLog().info(
+                      "If you do *not* want the Poetry package version to be automatically synced with the POM version, set <overridePackageVersion>false</overridePackageVersion> in the plugin's <configuration>");
+              packageManagerCommandHelper.executeAndLogOutput(Arrays.asList("version", expectedPythonPackageVersion));
+          } else {
+              getLog().debug(String.format(
+                      "Package Manager package version set to %s in pyproject.toml does not align with expected POM-derived version of %s",
```
meliz19, line 48: The `pyproject.toml` doesn't list the `poetry` or `uv` version. Update to say "<Poetry or uv> version is set to %s and does not align with expected POM-derived version of %s." Might need to create an abstract method to specify poetry or uv.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
cpointe-ibllanos: s: Should PoetryCommandHelper be removed here?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -14,7 +14,7 @@
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
-import org.checkerframework.checker.units.qual.A;
+import org.technologybrewery.habushu.exec.PackageManagerCommandHelper;
 import org.technologybrewery.habushu.exec.PoetryCommandHelper;
```
meliz19, line 18: same as above.

cpointe-ibllanos (approved):
Looks good once feedback is addressed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
jaebchoi: yes, otherwise it would throw nullpointer since UVcommandHelper is No-OP

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsMojo.java
jaebchoi: same answer above, executing with UVcommandHelper will throw nullpointer, need to return to skip early

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
jaebchoi: need if statement check early as we havn't implemented UV (This will throw MojoExecutionException from PythonPackageAndDependencyManagerFactory) 

Screenshot 2025-02-10 at 2 44 28 PM

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
jaebchoi: same above

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
jaebchoi: I think making abstraction solely on this method seems to add more complication, I would think it would make sense just call `HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile());` check.

jaebchoi: plus, class that calls this method are all deprecated. (i.e. RetrieveWheelsMojo, CacheWheelsMojo)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
jaebchoi: I was intending to do this work on follow up for ( aissemble lite build - Habushu/UV set up initialize Habushu) 
My initial scope of work was to make check for uv/poetry call and simple no-op test uv to make sure it can stand up. I fear this might get bigger ticket then expected.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UVCommandHelper.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public UVCommandHelper(File workingDirectory) {
+        super();
+        this.workingDirectory = workingDirectory;
+    }
+
+    @Override
+    public Pair<Boolean, String> getIsPackageManagerInstalledAndVersion() {
+        //TODO: NO-OP NEED TO IMPLEMENT
+        return null;
```
cwoods-cpointe: C: These no-op functions seem like not what we want to do. We should implement them as we need them. This pattern could make it hard to determine whats needed in the future.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PackageManagerCommandHelperFactory.java
```diff
@@ -0,0 +1,15 @@
...
+import java.io.File;
+import org.technologybrewery.habushu.util.HabushuUtil;
+
+public class PackageManagerCommandHelperFactory {
+    public static PackageManagerCommandHelper createPackageManagerCommandHelperSetup(File workingDirectory) {
+        if (HabushuUtil.checkPythonPackageManager(new File(workingDirectory, "pyproject.toml")) == HabushuUtil.PackageManager.POETRY){
+            return new PoetryCommandHelper(workingDirectory);
+        } else {
+            // TODO: return new UvSetup();
+            return new UVCommandHelper(workingDirectory);
```
cwoods-cpointe: q/c: If UV is not implemented at this point should we fail and explicitly state UV is not supported?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UVCommandHelper.java
jaebchoi: I just wanted to make it skeletal helper so that in next series of ticket we can implemented. If this feels risky, I can definitely delete this helper and either fail or not supported.

cwoods-cpointe: Yeah I think it could lead to tech debt down the road where these functions were originally created with the thought of implementing in the future but then we find out they aren't needed

jaebchoi: fair point, some of methods might not even being used. Ill kill them!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PackageManagerCommandHelperFactory.java
jaebchoi: Yeah I can state that UV is not supported "yet"

jaebchoi: updated!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -192,11 +200,11 @@ public static String getCleanVirtualEnvironmentPath(String virtualEnvFullPath) {
 
 	}
 
-	public static AbstractPythonPackageAndDependencyManagerSetup getPythonPackageAndDependencyManager(String pythonPackageAndDependencyManager,
+	public static AbstractPythonPackageAndDependencyManagerSetup getPythonPackageAndDependencyManager(PackageManager pythonPackageAndDependencyManager,
 	String pythonVersion, File baseDir, boolean rewriteLocalPathDepsInArchives, Log log, Boolean usePyenv, File patchInstallScript) throws MojoExecutionException {
     
     // Set the poetry-based parameters to null
-    if (!pythonPackageAndDependencyManager.equals(DEFAULT_PYTHON_PACKAGE_AND_DEPENDENCY_MANAGER)){
```
meliz19, line 199: S: Remove line 43 where `DEFAULT_PYTHON_PACKAGE_AND_DEPENDENCY_MANAGER` is defined and replace this parameter with the `PackageManager.POETRY`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -24,6 +26,12 @@
  */
 public final class HabushuUtil {
 
+	public enum PackageManager {
```
meliz19, line 29: Great idea. I like this!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -135,9 +135,17 @@ public class InstallDependenciesMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
```
meliz19, line 136: Much of the code in this class is specific to Poetry. I'm concerned with the way it's currently implemented, that it could lead to confusion down the road. I'm almost inclined to revert the changes and create an abstract class in a follow on ticket that we can build off of for poetry and uv, based on what's needed for each package manager.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -148,15 +156,15 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
...
+            packageManagerCommandHelper.executePackageManagerCommandAndLogAfterTimeout(
+                    packageManagerCommandHelper.createLockCommand(this.skipPoetryLockUpdate),
                     2,
                     TimeUnit.MINUTES
             );
 
         }
 
-        List<String> installCommand = poetryHelper.createInstallCommand(this.forceSync);
+        List<String> installCommand = packageManagerCommandHelper.createInstallCommand(this.forceSync);
```
meliz19: I: uv doesn't have an install command

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -168,15 +176,15 @@ public void doExecute() throws MojoExecutionException, MojoFailureException {
         }
 
         getLog().info("Installing dependencies...");
-        poetryHelper.executePoetryCommandAndLogAfterTimeout(installCommand, 2, TimeUnit.MINUTES);
+        packageManagerCommandHelper.executePackageManagerCommandAndLogAfterTimeout(installCommand, 2, TimeUnit.MINUTES);
     }
 
-    private void setUpInProjectVirtualEnvironment(PoetryCommandHelper poetryHelper) throws MojoExecutionException {
+    private void setUpInProjectVirtualEnvironment(PackageManagerCommandHelper packageManagerCommandHelper) throws MojoExecutionException {
```
meliz19, line 175: I: uv doesn't have use these arguments

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -7,9 +7,7 @@
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.logging.Log;
-import org.technologybrewery.habushu.exec.PoetryCommandHelper;
-import org.technologybrewery.habushu.exec.PyenvCommandHelper;
-import org.technologybrewery.habushu.exec.PythonVersionHelper;
+import org.technologybrewery.habushu.exec.*;
```
meliz19: I'm not sure what the best practice is in java, but Python best practice is to use explicit imports rather than the wildcard. Might be clearer to keep the explicit imports.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -119,7 +117,7 @@ protected String pythonSourceMessage() throws MojoExecutionException {
 
     @Override
     public void registerRepositoryToSupportAuthenticatedDependencyResolution(String repoId, String username, String password) throws MojoExecutionException {
```
meliz19, line 121: I/S: This entire function is specific to Poetry, so it might be okay in this case to keep the `PoetryCommandHelper`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -201,27 +199,27 @@ private void checkForPoetryToml() throws MojoExecutionException {
     }
 
     private void configurePoetryToUsePyenv() throws MojoExecutionException {
-        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        PackageManagerCommandHelper packageManagerCommandHelper = createPackageManagerCommandHelper();
         if (usePyenv) {
             log.info("Configuring Poetry to use the pyenv-activated Python binary...");
-            poetryHelper.executeAndLogOutput(poetryHelper.createUsePyenvCommand());
+            packageManagerCommandHelper.executeAndLogOutput(packageManagerCommandHelper.createUsePyenvCommand());
```
meliz19: I: `createUsePyenvCommand` is specific to Poetry

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -201,27 +199,27 @@ private void checkForPoetryToml() throws MojoExecutionException {
...
         }
     }
   
     void installPoetryMonorepoDependencyPlugin() throws MojoExecutionException {
-        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
+        PackageManagerCommandHelper packageManagerCommandHelper = createPackageManagerCommandHelper();
         log.info("Checking for updates to poetry-monorepo-dependency-plugin...");
-        poetryHelper.installPoetryPlugin("poetry-monorepo-dependency-plugin@latest");
+        packageManagerCommandHelper.installPackageManagerPlugin("poetry-monorepo-dependency-plugin@latest");
```
meliz19: I: This install is specific to Poetry

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PackageManagerCommandHelper.java
```diff
@@ -0,0 +1,151 @@
...
+    public abstract String getPackageManagerCacheDirectoryPath() throws MojoExecutionException;
+
+    /**
+     * Returns whether the specified dependency package is installed within this
+     * package manager project's virtual environment (and pyproject.toml).
+     *
+     * @param packageName
+     * @return
+     */
+    public boolean isDependencyInstalled(String packageName) {
```
meliz19: I: the `show` command is specific to Poetry

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PackageManagerCommandHelper.java
```diff
@@ -0,0 +1,151 @@
...
+     */
+    public abstract int installPackageManagerPlugin(String name) throws MojoExecutionException;
+
+    public List<String> createLockCommand() {
+        return createLockCommand(true);
+    }
+
+    public abstract List<String> createLockCommand(boolean skipLockUpdate);
+
+    public abstract List<String> createInstallCommand(boolean forceSync);
```
meliz19: I: this is specific to Poetry

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PackageManagerCommandHelper.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public abstract List<String> createUsePyenvCommand();
```
meliz19: I: this is specific to Poetry

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PackageManagerCommandHelperFactory.java
```diff
@@ -0,0 +1,17 @@
+package org.technologybrewery.habushu.exec;
+
+import java.io.File;
+
+import org.apache.commons.lang3.NotImplementedException;
+import org.technologybrewery.habushu.util.HabushuUtil;
+
+public class PackageManagerCommandHelperFactory {
+    public static PackageManagerCommandHelper createPackageManagerCommandHelperSetup(File workingDirectory) {
+        if (HabushuUtil.checkPythonPackageManager(new File(workingDirectory, "pyproject.toml")) == HabushuUtil.PackageManager.POETRY){
```
meliz19: S: It might be good to have "pyproject.toml" stored as a variable. Up to you though.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -221,7 +222,7 @@ public Integer executePoetryCommandAndLogAfterTimeout(List<String> arguments, in
      * @return
      * @throws MojoExecutionException
      */
-    public int installPoetryPlugin(String name) throws MojoExecutionException {
+    public int installPackageManagerPlugin(String name) throws MojoExecutionException {
```
meliz19, line 225: I: `self add` is specific to Poetry.

meliz19, line 225: I: `self add` is specific to Poetry

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -260,16 +261,16 @@ public List<String> createLockCommand() {
         return createLockCommand(true);
     }
 
-    public List<String> createLockCommand(boolean skipPoetryLockUpdate) {
+    public List<String> createLockCommand(boolean skipLockUpdate) {
```
meliz19: I: these `lock` arguments are specific to Poetry

habushu-mixology-consumer-uv/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+verify_ssl = true
+
+[project]
+authors = [
+    {name = "James Choi", email = "choi_james@bah.com"},
+]
+license = {text = "MIT License"}
+requires-python = "<4.0,>=3.11"
+dependencies = [
+    "habushu-mixology @ file:///${PROJECT_ROOT}/../habushu-mixology",
```
meliz19: S: It might make sense to make this dependent on `habushu-mixology-uv`.

habushu-mixology-consumer-uv/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+]
+name = "habushu_mixology_consumer"
```
meliz19: S: It might make sense to update this to `"habushu_mixology_consumer_uv"

habushu-mixology-consumer-uv/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "2.18.0.dev"
+description = "Example of how to use the habushu-maven-plugin to consume other Habushu modules"
```
meliz19:
```suggestion
description = "Example of how to use the habushu-maven-plugin to consume other Habushu modules in a `uv`-based setting"
```

habushu-mixology-consumer-uv/src/habushu_mixology_consumer/nested/simple_dependency.py
```diff
@@ -0,0 +1,6 @@
+from habushu_mixology.reusable_module.worker import SubWorker
```
meliz19:
```suggestion
from habushu_mixology_uv.reusable_module.worker import SubWorker
```

habushu-mixology-consumer-uv/src/habushu_mixology_consumer/verify_install.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/python
+from habushu_mixology.reusable_module.worker import SubWorker
```
meliz19:
```suggestion
from habushu_mixology_uv.reusable_module.worker import SubWorker
```

habushu-mixology-consumer-uv/tests/features/steps/generate-report-with-manual-tests.py
```diff
@@ -0,0 +1,21 @@
+from behave import given, when, then  # pylint: disable=no-name-in-module
+
+
+@given("a Python module with at least one automated test and at least one manual test")
+def step_impl(context):
```
meliz19, line 5: S: I might replace `return` with `NotImplementedError` for these steps.

habushu-mixology-consumer-uv/tests/features/steps/unpack_deps.py
```diff
@@ -0,0 +1,27 @@
+from behave import given, when, then  # pylint: disable=no-name-in-module
+from krausening.properties import PropertyManager
+from habushu_mixology_consumer.nested.simple_dependency import (
```
meliz19:
```suggestion
from habushu_mixology_consumer_uv.nested.simple_dependency import (
```

habushu-mixology-consumer-uv/tests/features/unpack-deps.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Verify that dependencies get unpacked correctly for tests
+    I want to be able to test python code that have dependencies unpacked by habushu
+
+  Scenario: Files can reference dependency files and build successfully
+    When I reference a dependency in my python code
+    Then the build can successfully execute the tests
+
+Scenario: Files can reference transitive dependencies and build successfully
+    When I reference krausening transitively from habushu-mixology
```
meliz19:
```suggestion
    When I reference krausening transitively from habushu-mixology-uv
```

habushu-mixology-consumer-uv/tests/features/unpack-deps.feature
```diff
@@ -0,0 +1,10 @@
...
+    Then I can access it in habushu-mixology-consumer without further intervention
```
meliz19:
```suggestion
    Then I can access it in habushu-mixology-consumer-uv without further intervention
```

habushu-mixology-uv/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.technologybrewery.habushu</groupId>
+        <artifactId>habushu</artifactId>
+        <version>2.18.0-SNAPSHOT</version>
+    </parent>
+
+    <name>habushu::Mixing Drinks with Habushu</name>
+    <description>Example of how to use the habushu-maven-plugin</description>
```
meliz19:
```suggestion
    <description>Example of how to use the habushu-maven-plugin in a uv-based project</description>
```

habushu-mixology-uv/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+                        <useDevRepository>true</useDevRepository>
+                    </configuration>
+                    <!-- Utilize the run-command-in-virtual-env goal to demonstrate how
+                        commands/scripts may be executed via "poetry run" and bound to the build
+                        lifecycle phases -->
+                    <executions>
+                        <execution>
+                            <configuration>
+                                <runCommandArgs>python -m grpc_tools.protoc -I=src
+                                    --python_out=src/habushu_mixology/generated src/person.proto
```
meliz19: ``
```suggestion
                                    --python_out=src/habushu_mixology_uv/generated src/person.proto
```

habushu-mixology-uv/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+                                <packageName>uvicorn</packageName>
+                                <operatorAndVersion>${version.uvicorn}</operatorAndVersion>
+                            </packageDefinition>
+                        </managedDependencies>
+                        <updateManagedDependenciesWhenFound>false</updateManagedDependenciesWhenFound>
+                        <failOnManagedDependenciesMismatches>false</failOnManagedDependenciesMismatches>
+                        <useDevRepository>true</useDevRepository>
+                    </configuration>
+                    <!-- Utilize the run-command-in-virtual-env goal to demonstrate how
+                        commands/scripts may be executed via "poetry run" and bound to the build
```
meliz19:
```suggestion
                        commands/scripts may be executed via "poetry run"/"uv run" and bound to the build
```

habushu-mixology-uv/src/habushu_mixology/helloworld.py
```diff
@@ -0,0 +1,17 @@
+import string
```
meliz19: I just noticed you didn't change the underlying directory to `habushu_mixology_uv`. Feel free to make the change or ignore my comments related to this above.

pom.xml
meliz19: Q: Would it be clearer to uses/leave less room for potential issues later on if we install the plugin and just let the build/tests fail?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
jaebchoi: Sounds good, Ill make note on that for follow up ticket!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
jaebchoi: will take note on follow up ticket

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
jaebchoi: same above

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
jaebchoi: same here in java, I think it got automatically changed it to wildcard upon lint, updated!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
jaebchoi: I think class itself is pyenvAndPoetry Setup, so I think this is okay

jaebchoi: Make sense, updated to PoetryCommandHelper

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PackageManagerCommandHelper.java
jaebchoi: updated to abstract method

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
jaebchoi: Sounds good, this is poetry specific class so fine to keep up good info on how we figure out uv impl!

jaebchoi: same above

habushu-mixology-consumer-uv/pyproject.toml
jaebchoi: make sense, missed that when I copied from habushu mixology

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
d-ryan-ashcraft: S: We want to list them all rather than use the wildcard.  You can get into a situation where there are collisions with wildcard. https://next.sonarqube.com/sonarqube/coding_rules?open=java%3AS2208&rule_key=java%3AS2208

habushu-mixology-uv/src/habushu_mixology/helloworld.py
jaebchoi: updating with _uv on directory def make sense!

pom.xml
jaebchoi: maybe you are right, as long as default one works, it can give something to debug on instead of blindly letting uv succeeded, removed!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
jaebchoi: hmm I see I can keep this until we figure uv install plug in for later!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -86,67 +90,74 @@ public class BehaveBddTestMojo extends AbstractHabushuMojo {
...
-
+       if (skipTests) {
+           getLog().warn("Tests are skipped (-DskipTests=true)");
+           return;
+       }
+
+      //TODO: Skipping since UV impl is no op. Remove once UV Impl is done.
+      // Note this doesn't imply we will implement UV solely on this check, implementation will differ based on investigation on further ticket.
+      // (whether we abstract out or just use simple check)
+      if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.UV) {
```
cwoods-cpointe: q/c: if uv is not implemented then should we just fail? Ideally we could do this in one location instead of in each of the mojos. 
The user should be able to perform individual commands that are implemented like clean. If they are doing a full build then it could be confusing if the build passes when it actually skips a bunch of steps

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -164,7 +163,7 @@ public List<String> validatePoetryInstallationAndVersion(ValidationTrackingStatu
         PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
         if (!validationTracker.isAlreadyValidatedPoetryInstallation()) {
             log.debug("Checking if Poetry is installed...");
-            Pair<Boolean, String> poetryInstallStatusAndVersion = poetryHelper.getIsPoetryInstalledAndVersion();
+            Pair<Boolean, String> poetryInstallStatusAndVersion = poetryHelper.getIsPackageManagerInstalledAndVersion();
```
meliz19: should this also be reverted back?
```suggestion
            Pair<Boolean, String> poetryInstallStatusAndVersion = poetryHelper.getIsPoetryInstalledAndVersion();
```
```suggestion
            Pair<Boolean, String> poetryInstallStatusAndVersion = poetryHelper.getIsPackageManagerInstalledAndVersion();
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -211,7 +210,7 @@ private void configurePoetryToUsePyenv() throws MojoExecutionException {
     void installPoetryMonorepoDependencyPlugin() throws MojoExecutionException {
         PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
         log.info("Checking for updates to poetry-monorepo-dependency-plugin...");
-        poetryHelper.installPoetryPlugin("poetry-monorepo-dependency-plugin@latest");
+        poetryHelper.installPackageManagerPlugin("poetry-monorepo-dependency-plugin@latest");
```
meliz19, line 213: same here

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
jaebchoi: Yeah I was more worried about failing build than making it confusion to consumers that it passes.
seems like majority of crew feels latter so I can update to Not implemented exception to fail.

> Ideally we could do this in one location instead of in each of the mojos.

yeah I can just do this check at first mojo per [components](https://github.com/TechnologyBrewery/habushu/blob/dev/habushu-maven-plugin/src/main/resources/META-INF/plexus/components.xml)  to fail instead of each mojos!

d-ryan-ashcraft:
C (Critical): We're going to stop this here.  This has gotten a little out of hand, and we need to reevaluate what this change was about and stay on target.  Once we hit more than about 15 comments, we have likely jumped the shark.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
jaebchoi: updated!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
jaebchoi: make sense updated!

jaebchoi:
> C (Critical): We're going to stop this here. This has gotten a little out of hand, and we need to reevaluate what this change was about and stay on target. Once we hit more than about 15 comments, we have likely jumped the shark.

I can do after at standup on the scope of this PR and we can evaluate from there.

d-ryan-ashcraft:
Given the level of back and forth and growing complexity, we decided to close this PR and flow a series of PRs instead:
- Apply the pattern to exactly one Mojo (Maven Plugin), work that PR through the process to gain consensus and harden the pattern
- Flow the next group (~3-5 Mojos)
- Flow remaining group

Issue #208, opened 2025-02-10
Add support for uv - Ensure uv requirements are installed on a user's machine if a uv-based pyproject.toml is present in a module


meliz19:
There is a demand signal to be able to use [uv](https://docs.astral.sh/uv/) as a Python package and dependency management tool within the context of the habushu-maven-plugin. The purpose of this ticket is to continue the work from #198 and develop a subclass to ensure that habushu's `uv` requirements are installed on a user's machine, similar to the current `PyenvAndPoetrySetup` class.

## DoD
- [x] Develop the necessary functionality to ensure that habushu's `uv` requirements are installed on a user's machine.
- [x] Ensure that a `uv`-based example project exists in the repo and verify the functionality works as expected.

## DoD
- Pull down the `3.0.0-add-uv-support` feature branch
- `mvn clean install -Pbootstrap`

### Test 1: `.python-version` and `.venv` are missing
1. Using `pipx` install uv by running `pipx install uv`.
2. Run `mvn clean install`.
3. The `habushu-uv-package` module will fail, but search the logs for the following line: `(default-validate-python-package-and-dependency-manager) @ habushu-uv-package`. You should see something like the following below this line.
```
[INFO] Found uv 0.5.31
[INFO] The Python version is not currently set for the project. Setting the version...
[INFO] Executing uv command: uv python pin 3.11.9
[INFO] Using Python 3.11.9 (managed by uv)
```
4. Verify that `.python-version` exists in `examples/habushu-uv-package`.

### Test 2: `.python-version` is missing but `.venv` is present
1. Run `rm examples/habushu-uv-package/.python-version`.
2. Run `cd examples/habushu-uv-package && uv venv --python 3.12 && cd - > /dev/null`.
3. Run `mvn clean install -pl :habushu-uv-package`.
2. The `habushu-uv-package` module will fail, but search the logs for the following line: `(default-validate-python-package-and-dependency-manager) @ habushu-uv-package`. You should see something like the following below this line.
```
[INFO] Found uv 0.5.31
[INFO] Executing uv command: uv run python --version
[INFO] A .python-version does not currently exist for this package. Creating it now with the Python version being used in the current virutal environment (in .venv).
[INFO] Executing uv command: uv python pin 3.12.8
```
4. Verify that `.python-version` exists in `examples/habushu-uv-package`.

### Test 3: `.python-version` is present
1. Run `mvn clean install -pl :habushu-uv-package`.
[...]

PR #221, opened 2025-02-12
[#208] default-validate-python-package-and-dependency-manager goal en…

meliz19:
…abled for uv


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -36,6 +36,7 @@ public enum PackageManager {
      * Specifies the semver compliant requirement for the default version of Python that
      * must be installed and available for Habushu to use.
      */
+	// MARG TODO: Is there a reason this is set to 3.11.4? Why not the max(min poetry requirement, min uv requirement)?
     public static final String PYTHON_DEFAULT_VERSION_REQUIREMENT = "3.11.4";
```
meliz19: Q: Is there a reason this is set to 3.11.4? Why not the max(min poetry requirement, min uv requirement)?

examples/habushu-uv-package/pyproject.toml
```diff
@@ -26,4 +22,20 @@ dev = [
...
-]
\ No newline at end of file
+]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+
+[[tool.pdm.source]]
```
meliz19, line 32: I: Lines 32 and below will be changed once the full `uv` functionality is enabled.

examples/habushu-uv-package/pyproject.toml
```diff
@@ -17,6 +12,7 @@ dependencies = [
 name = "habushu_uv_package"
 version = "2.18.0.dev"
 description = "Example of how to use the habushu-maven-plugin in a `uv`-based setting"
+readme = "README.md"
```
meliz19, line 15: I: Required for `uv` to build a package.

examples/pom.xml
```diff
@@ -22,6 +22,7 @@
             <modules>
                 <module>habushu-poetry-package</module>
                 <module>habushu-poetry-package-consumer</module>
+                <module>habushu-uv-package</module>
```
meliz19, line 25: I: Added so that when we run `mvn clean install`, the `uv` package is included. Currently, only the `default-validate-python-package-and-dependency-manager` goal is supported for `uv`, so the rest of this module breaks.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -121,6 +121,8 @@ public void execute() throws MojoExecutionException {
     private void clean() throws MojoExecutionException {
         List<Fileset> filesetsToDelete = new ArrayList<>();
         boolean removeVenvManually = false;
+
+        // TODO: This code is specific to Poetry. uv does not provide a command to remove a virtual environment like Poetry.
```
meliz19, line 125: I: In the interest of keeping small PRs, I wanted to note this for whoever's ticket touches this mojo.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -0,0 +1,115 @@
...
+   
+    public UvSetup(String pythonVersion, File baseDir, boolean rewriteLocalPathDepsInArchives, Log log) throws MojoExecutionException {
+        super(pythonVersion, baseDir, rewriteLocalPathDepsInArchives, log); 
+    }
+
+    /**
+     * Checks if the .python-version file exists. If it exists, it checks the listed python 
+     * version and updates the listed version to the pythonVersion, if necessary. If the .python-version
+     * file does not exisit, it throws a MojoExecutionException.
+     */
```
meliz19, line 35: I: Unlike Poetry, `uv` users don't have to install the desired python version before setting up the virtual environment. For this reason, I decided that the best way to check of the correct python version will be used in `uv` generated virtual environments is to verify that the `.python-version` file is present in the package directory. If it is not present, habushu will raise an error message saying it must be present.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/SharedCommandHelper.java
```diff
@@ -0,0 +1,166 @@
+package org.technologybrewery.habushu.exec;
```
meliz19: I: If this class looks good, I'll flow a quick follow-on ticket to replace the current Poetry-specific methods with their corresponding methods herein.

examples/habushu-poetry-package/src/habushu_poetry_package/generated/person_pb2.py
d-ryan-ashcraft, line 1: A (Administrative): I think we missed this before, but should probably delete this and ignore the `generated/` folders

examples/pom.xml
d-ryan-ashcraft: Breaks as in fails the build?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
d-ryan-ashcraft: I'm wondering if we want to default to a version of Python instead.  Failing loudly is good for forcing user action, but at the same time, this type of behavior may be a turnoff for Python developers.  Thoughts?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -0,0 +1,115 @@
...
+                uvHelper.updatePythonVersion(pythonVersion);
+                currentPythonVersion = uvHelper.getCurrentPythonVersion();
+            }
+        return currentPythonVersion;
+    }
+
+    @Override
+    protected List<String> validatePackageAndDependencyManagerInstallationAndVersion(ValidationTrackingStatus validationTracker, List<String> missingRequiredToolMsgs) 
+        throws MojoExecutionException {
+            String alreadyValidatedUvVersion = validationTracker.getAlreadyValidatedUvVersion();
```
d-ryan-ashcraft: A: Unrelated to this specific PR, we should probably have `ValidationTrackingStatus` just have generic `alreadyValidatedInstallation` and `alreadyValidatedVersion` so it's more generic and fungible.  It doesn't seem like the tools names add value, and this makes it more extensible and OO-friendly.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -0,0 +1,115 @@
...
+                alreadyValidatedUvVersion += VALIDATED_IN_PRIOR_BUILD_PHASE;
+            }
+    
+            log.info("Found uv " + alreadyValidatedUvVersion);
+    
+            return missingRequiredToolMsgs;
+    }
+
+    @Override
+    protected void finalizePythonPackageAndDependencyManagerConfiguration() throws MojoExecutionException {
```
d-ryan-ashcraft, line 83: A: we could make this an empty implementation in the super class so this, and another future tool, won't have to override it unless they need to.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/SharedCommandHelper.java
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Facilitates the execution of package manager commands.
+ */
+public class SharedCommandHelper {
```
d-ryan-ashcraft: A: could probably shorten to just `CommandHelper`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.habushu.HabushuException;
+/**
+ * Facilitates the execution of uv commands.
+ */
+public class UvCommandHelper {
```
d-ryan-ashcraft: S: Should this extend `Shared/CommandHelper` instead of compositing it?  I would initially think it should.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.logging.Log;
+import org.technologybrewery.habushu.exec.UvCommandHelper;
+import org.technologybrewery.habushu.util.UvUtil;
+
+import java.io.File;
+import java.util.List;
+
+/**
+ * Subclass of PythonPackageAndDependencyManagerSetup. Ensures uv-related pre-requisite tools that Habushu leverages are installed and available on the
```
aaron-gary: Very minor: I would leave off mentioning that this is a subclass of x.  Line 24 is easy enough to read to get that information and the rest of your comment provides the important information.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -0,0 +1,115 @@
...
+    public UvSetup(String pythonVersion, File baseDir, boolean rewriteLocalPathDepsInArchives, Log log) throws MojoExecutionException {
+        super(pythonVersion, baseDir, rewriteLocalPathDepsInArchives, log); 
+    }
+
+    /**
+     * Checks if the .python-version file exists. If it exists, it checks the listed python 
+     * version and updates the listed version to the pythonVersion, if necessary. If the .python-version
+     * file does not exisit, it throws a MojoExecutionException.
+     */
+
```
carter-cundiff: A: Think there's an extra space between the comment and method

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
carter-cundiff: Q: Would it be possible to create the `.python-version` file with the correct version if it doesn't exist?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -0,0 +1,115 @@
...
+                currentPythonVersion = uvHelper.getCurrentPythonVersion();
+            }
+        return currentPythonVersion;
+    }
+
+    @Override
+    protected List<String> validatePackageAndDependencyManagerInstallationAndVersion(ValidationTrackingStatus validationTracker, List<String> missingRequiredToolMsgs) 
+        throws MojoExecutionException {
+            String alreadyValidatedUvVersion = validationTracker.getAlreadyValidatedUvVersion();
+            UvCommandHelper uvHelper = createUvCommandHelper();
```
carter-cundiff: A: Could move this to within the below if block

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -0,0 +1,180 @@
...
+     *
+     * @return
+     */
+    public Pair<Boolean, String> getIsUvInstalledAndVersion() {
+        try {
+            ProcessExecutor executor = sharedCommandHelper.createPackageManagerExecutor(Arrays.asList("--version"));
+            String versionResult = executor.executeAndGetResult(logger);
+
+            // Extracts version number from output, given the following format "uv 0.5.20 (1c17662b3 2025-01-15)"
+            Pattern versionPattern = Pattern.compile(extractUvVersionRegex);
```
d-ryan-ashcraft: A: I believe it is best to make `Pattern.compile(...)` a static member variable as the Pattern is thread safe and it is a relatively more expensive call.  Probably fine here, but a good habit to build.

The Matcher would stay where it is, however.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
meliz19: I thought about that. If we were to initialize a `uv`- based project via `habushu`, it would make sense for `habushu` to create the `.python-version` file. However, if a user already has a `uv`- based project setup and they're adding `habushu` to the project, I'm afraid that if `habushu` were to automatically create the `.python-version` file for the user, it could potentially mess up their current virtual environment.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/SharedCommandHelper.java
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @param arguments
+     * @param environmentVariables
+     */
+    public void executeAndLogOutput(List<String> arguments, Map<String, String> environmentVariables) {
+        if (logger.isInfoEnabled()) {
+            logger.info("Executing " + packageManagerCommand + " command: {} {}", packageManagerCommand, StringUtils.join(arguments, " "));
+        }
+        ProcessExecutor executor = createPackageManagerExecutor(arguments, environmentVariables);
+        executor.executeAndRedirectOutput(logger);
```
carter-cundiff: Q: Why not return `int` like above in `public int executeAndLogOutput(List<String> arguments)`?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -0,0 +1,180 @@
...
+     * directory via the .python-version file.
+     *
+     * @return
+     */
+    public String getCurrentPythonVersion() throws MojoExecutionException {
+        String currentPythonVersion = pythonVersionFileContents();
+
+        if (!currentPythonVersion.isEmpty()) {
+            return currentPythonVersion;
+        } else throw new MojoExecutionException("A .python-version file is required to have a reproducible build. To create this file, from the package directory, run `uv python pin <python-version>`.");
```
d-ryan-ashcraft: S: always use braces.  This avoids inevitable logic errors and improves readability.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Returns whether the specified dependency package is installed within this
+     * uv project's virtual environment (and pyproject.toml).
+     *
+     * @param packageName
+     * @return
+     */
+    public boolean isDependencyInstalled(String packageName) {
+        try {
+            sharedCommandHelper.execute(Arrays.asList("pip", "show", packageName));
```
d-ryan-ashcraft: S: is `pip` really how `uv` determines is a package is installed?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -0,0 +1,180 @@
...
+        } catch (Throwable e) {
+            return false;
```
d-ryan-ashcraft, line 119: S: If this results in an exception, probably something exceptional happened that should cause us to fail hard?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
d-ryan-ashcraft: We typically update this periodically to the oldest version with a MacOS/Windows installer that has supported security patches.  It's been a while, so we may want to consider bumping it up.

d-ryan-ashcraft (approved):
Looks good overall - some feedback inlined to check out and action.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/SharedCommandHelper.java
```diff
@@ -0,0 +1,166 @@
...
+     * @return
+     */
+    public Integer executePackageManagerCommandAndLogAfterTimeout(List<String> arguments, int timeout, TimeUnit timeUnit, String cleanCacheMessage) {
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        Future<Integer> future = executor.submit(() -> this.executeAndLogOutput(arguments));
+        try {
+            return future.get(timeout, timeUnit);
+        } catch (TimeoutException e) {
+            logger.warn(packageManagerCommand + " " + String.join(" ", arguments)
+                    + " has been running for quite some time, you may want to quit the mvn process (Ctrl+c) and run " + cleanCacheMessage + " and restart your build.");
```
carter-cundiff: S: May be worth generalizing this method so that it's not only for cache related issues to provide more flexibility down the road. So instead of `cleanCacheMessage` we could just have `messageToDisplay`.

examples/pom.xml
meliz19: yes, it fails the build.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
carter-cundiff: > if a user already has a uv- based project setup and they're adding habushu to the project, I'm afraid that if habushu were to automatically create the .python-version file for the user

If this were true, wouldn't the user already have a  .python-version file as part of their uv project?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -0,0 +1,115 @@
...
+    };
+
+    @Override
+    public void registerRepositoryToSupportAuthenticatedDependencyResolution(String repoId, String username, String password) throws MojoExecutionException, NotImplementedException {
+        // TODO: move this method out of the abstract methods as this is specific to Poetry.
+        // uv does not support a config command to store username and password credentials. Users must enter this information via the publish command.
+    }
+
+    /**
+     * Creates a {@link UvCommandHelper} that may be used to invoke Poetry
```
jaebchoi: A:  
```suggestion
     * Creates a {@link UvCommandHelper} that may be used to invoke UV
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
jaebchoi: might also be good to provide error message as well.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/UvUtil.java
```diff
@@ -0,0 +1,13 @@
+package org.technologybrewery.habushu.util;
+
+/**
+ * Common utility methods for handling TOML Poetry.
```
jaebchoi: A: 
```suggestion
 * Common utility methods for handling TOML UV.
```

jaebchoi (approved):
Looks good!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
aaron-gary: Hmmm... good question

aaron-gary (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
meliz19: Ah, good to know. 

- [x] It looks like Python `3.12.9` is the oldest version with a MacOS/Windows installer that has supported security patches. I'll update accordingly.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
meliz19: No. To create a virtual environment with `uv`, users have a few options.

1. They can do nothing and the latest available version of python will be used within the virtual environment. 
2. They can run `uv --python <desired-version>` to create a virtual environment with the desired version of python.
3. They can have a `.python-version` file in the module and `uv` will use that file to create the virtual environment with the specified Python version. 

I'm forcing users to use the last option since it's the only option that allows for reproducible builds.

carter-cundiff: Okay, if the goal is reproducible builds and we need that file to make it happen, I think it's fair to have them create the file otherwise fail

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
meliz19: Good to know. I should have googled how the validation tracker works. Updated.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
meliz19: Thoughts? ^

carter-cundiff: Could we interrogate the python version from the venv and create the file with that?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
meliz19: oh duh. I got too excited about using an abstract method. :woman_facepalming:

meliz19: I started to update this, but I'm going to have to change the PyenvAndPoetrySetup class, so I'll add that to my follow on.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
meliz19: yes, I think I can do that. I'll incorporate that in. 

- [x] update python version checker to create .python-version with existing venv's python version if it exists

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/SharedCommandHelper.java
meliz19: Good question. This was an existing method that I moved to this class. If I were to return an `int`, then I'm assuming whatever method(s) calls this method will verify that a passing exit code was returned? Right now, the method(s) that call this method don't expect a return value.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
meliz19: lol yes. see https://github.com/astral-sh/uv/issues/3164#issuecomment-2067801609

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
meliz19: good catch. ty!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
meliz19: This method is used to check if `behave`, `black` etc. are installed. If false is returned, the user is notified in the next step of the pipeline that it is being installed.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
meliz19: Yes, I think so. My original understanding of abstract classes was a little off. ~I'll fix in my follow on.~

- [x] fixed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuMojo.java
```diff
@@ -7,7 +7,7 @@
 import org.technologybrewery.habushu.util.HabushuUtil;
 
 /**
- * Ensures that the current project is a valid Poetry or UV project and initializes
+ * Ensures that the current project is a valid Poetry or uv project and initializes
```
meliz19, line 10: I: uv's docs have "uv" lowercased everywhere, so just standardizing this across the project.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InitializeHabushuUV.java
```diff
@@ -3,44 +3,60 @@
 import org.apache.commons.lang3.StringUtils;
```
meliz19, line 3: I/Q: I'm not sure why this filename is showing UV as capitalized. I corrected it locally, but it's not showing up as a change in git. 🤔

PR #230, opened 2025-02-18
#227 pyenv install missing python

### LINKED ISSUE #227: BUG: pyenv fails to install python version when no current version is set
If pyenv is installed but no global or local version is set, Habushu fails to install the desired python version. The same failure occurs when a version is set with the .python-version file but is not installed. This behavior is only observed when building the poetry modules.

## Steps to reproduce:
1. Ensure pyenv is installed
2. Uninstall the default python version 3.11.9 - `pyenv uninstall 3.11.9`
4. Build the habushu-poetry-package - `mvn clean install -pl :habushu-poetry-package`
6. Observe the exception and build failure.

## Expected Behavior: 
Habushu sees pyenv is installed but no python version is installed. It then uses pyenv to install the desired python version.

## Actual Behavior
Habushu attempts to get the current python version installed, finds none, and fails the build without attempting to install the desired version.

## Testing
1. Checkout the 3.0.0-add-uv-support
2. Build the Habushu plugin - `mvn clean install -Pbootstrap` 
3. Uninstall python 3.11.9 - `pyenv uninstall 3.11.9`
4. Build the full project - `mvn clean install -pl :habushu-poetry-package`
5. Verify the python version 3.11.9 is installed and the build passes

woods-cpointe:
issue - #227

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
@@ -69,7 +69,11 @@ public class PyenvAndPoetrySetup extends AbstractPythonPackageAndDependencyManag
             missingRequiredToolMsgs.add(
                     "'pyenv' is not currently installed! Please install pyenv and try again. Visit https://github.com/pyenv/pyenv for more information.");
         } else {
-            currentPythonVersion = pyenvHelper.getCurrentPythonVersion();
+            try {
+                currentPythonVersion = pyenvHelper.getCurrentPythonVersion();
+            } catch (Exception e) {
+                log.info("Failed to find current python version. Attempting to install it now.");
+            }
             if (!pythonVersion.equals(currentPythonVersion)) {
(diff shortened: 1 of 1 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -69,7 +69,11 @@ private String validateAndConfigurePyenv(List<String> missingRequiredToolMsgs, S
             missingRequiredToolMsgs.add(
                     "'pyenv' is not currently installed! Please install pyenv and try again. Visit https://github.com/pyenv/pyenv for more information.");
         } else {
-            currentPythonVersion = pyenvHelper.getCurrentPythonVersion();
+            try {
+                currentPythonVersion = pyenvHelper.getCurrentPythonVersion();
+            }catch (Exception e) {
```
meliz19:
```suggestion
            } catch (Exception e) {
```

meliz19 (approved):

d-ryan-ashcraft (approved):

PR #232, opened 2025-02-18
[#266] Fix error message when uv-based project is present but uv is n…

meliz19:
…ot installed

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -37,6 +37,20 @@ jobs:
               if: steps.cached-poetry.outputs.cache-hit != 'true'
               uses: snok/install-poetry@v1
 
+            - name: Load cached uv installation
+              id: cached-uv
+              uses: actions/cache@v4
+              with:
+                path: ~/.local
+                key: uv-0  # increment to reset cache
+
+            - name: Install uv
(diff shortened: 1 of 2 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
```diff
@@ -79,6 +79,10 @@ public void execute() throws MojoExecutionException {
 
         missingRequiredToolMsgs = validatePackageAndDependencyManagerInstallationAndVersion(validationTracker, missingRequiredToolMsgs);
 
+        if (!missingRequiredToolMsgs.isEmpty()) {
```
meliz19, line 82: I: Because of previous refactoring, we needed move this line before line 86.

aaron-gary (approved):

.github/workflows/maven.yaml
```diff
@@ -37,6 +37,20 @@ jobs:
               if: steps.cached-poetry.outputs.cache-hit != 'true'
               uses: snok/install-poetry@v1
 
+            - name: Load cached uv installation
```
meliz19, line 40: I: Required for GitHub Actions

Issue #235, opened 2025-02-18
Refactor Pyenv installation verification


meliz19:
## Description
Currently, if a project is setup to use Poetry, habushu checks to see if Poetry is installed. At this point the code checks to see if all the necessary tools are installed and breaks if they are not installed. After this, habushu checks to see if Pyenv is to be used and if so, it checks to see if it's installed. The purpose of this ticket is to move the Pyenv installation check to be in the same location as the Poetry installation check. This will make the logic of the code more straightforward. 

## DOD
- [x] Perform all tool installation checks at one location within the codebase

## Test Strategy/Script
-  Run `mvn clean install -Pbootstrap`
- Uninstall `pyenv`. Make sure that it's not on your `PATH` and comment out the `pyenv`-related lines from your `~/.zshrc` or `~/.bashrc`. 
- Run `mvn clean install -Dmaven.build.cache.skipCache=true -pl :habushu-poetry-package`
- You should get the following error message:
```
[INFO] --- habushu:2.18.0-SNAPSHOT:validate-python-package-and-dependency-manager (default-validate-python-package-and-dependency-manager) @ habushu-poetry-package ---
[INFO] Found Poetry 2.0.0
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.244 s
[INFO] Finished at: 2025-02-19T08:55:41-05:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.technologybrewery.habushu:habushu-maven-plugin:2.18.0-SNAPSHOT:validate-python-package-and-dependency-manager (default-validate-python-package-and-dependency-manager) on project habushu-poetry-package: 
[ERROR] ['pyenv' is not currently installed! Please install pyenv and try again. Visit https://github.com/pyenv/pyenv for more information.]
[ERROR] 
```
- Reinstall `pyenv`
- Run `pyenv versions` and make sure your global Python version is **not** set to 3.11.9. If it is, change it to any other Python version.
- Run `source ~/.zshrc` or `source ~/.bashrc`, whichever is appropriate for you.
- Re-run `mvn clean install -Dmaven.build.cache.skipCache=true -pl :habushu-poetry-package`.
- You should get an error message similar to the following. 
```
[INFO] Found Poetry 2.0.0
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[...]

PR #236, opened 2025-02-19
[#235] Refactor pyenv install verification

meliz19:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -52,59 +52,63 @@ public PyenvAndPoetrySetup(String pythonVersion, File baseDir, boolean rewriteLo
         this.patchInstallScript = patchInstallScript;
     }
 
-    public String configurePyenvOrStraightPython(boolean usePyenv, List<String> missingRequiredToolMsgs, File patchInstallScript) throws MojoExecutionException {
```
meliz19, line 55: I: Cleaning code while I'm here - moved this logic to `configurePythonUsingPackageAndDependencyManager` (lines 96 - 101) to remove unnecessary methods

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -52,59 +52,63 @@ public PyenvAndPoetrySetup(String pythonVersion, File baseDir, boolean rewriteLo
...
-
-    private String validateAndConfigurePyenv(List<String> missingRequiredToolMsgs, String currentPythonVersion, File patchInstallScript) throws MojoExecutionException {
+    private List<String> validatePyenvInstallation(List<String> missingRequiredToolMsgs) throws MojoExecutionException {
         PyenvCommandHelper pyenvHelper = createPyenvCommandHelper();
         log.debug("Checking if pyenv is installed...");
         if (!pyenvHelper.isPyenvInstalled()) {
             missingRequiredToolMsgs.add(
                     "'pyenv' is not currently installed! Please install pyenv and try again. Visit https://github.com/pyenv/pyenv for more information.");
         } else {
-            try {
```
meliz19, line 72: I: Breaking up the logic of `validateAndConfigurePyenv` so that we check if `pyenv` is installed in one method  (`validatePyenvInstallation`) and then use it in another method to configure the Python version (`validateAndConfigurePythonViaPyenv`). This will allow us to check if both Poetry and pyenv are installed within the `validatePackageAndDependencyManagerInstallationAndVersion` method rather than checking if these tools are installed at separate stages of the pipeline.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -52,59 +52,63 @@ public PyenvAndPoetrySetup(String pythonVersion, File baseDir, boolean rewriteLo
...
+        PythonVersionHelper pythonVersionHelper = new PythonVersionHelper(baseDir, pythonVersion);
+        String postPyenvActivatedPythonVersion = pythonVersionHelper.getCurrentPythonVersion();
+        List<String> pyenvMisconfiguration  = new ArrayList<>();
+        if (!pythonVersion.equals(postPyenvActivatedPythonVersion)) {
+            pyenvMisconfiguration.add(String.format("Expected 'pyenv' to set Python to %s but instead found %s!",
+                    pythonVersion, postPyenvActivatedPythonVersion));
+                    pyenvMisconfiguration.add("'pyenv' is installed, but not configured correctly.  " +
+                    "Ensure your PATH includes 'pyenv init -' expected content " +
+                    "OR do not configure habushu to use 'pyenv' to manage the Python version!");
+            throw new MojoExecutionException(StringUtils.join(System.lineSeparator(), pyenvMisconfiguration, System.lineSeparator()));
```
meliz19: I: lnstead of passing these error messages back to the abstract method via `missingRequiredToolMsgs`, we're now throwing an exception here to fail fast because `pyenv` isn't configured correctly.

d-ryan-ashcraft: S: We initially did it via `missingRequiredToolMsgs` because people get aggravated when they fix one tool and find out another is missing.  So, if we can do both at once, it makes for a better onboarding experience.  I'm probably be inclined to keep it that way.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -52,59 +52,63 @@ public PyenvAndPoetrySetup(String pythonVersion, File baseDir, boolean rewriteLo
...
         }
         return currentPythonVersion;
     }
 
     @Override
     protected String configurePythonUsingPackageAndDependencyManager(List<String> missingRequiredToolMsgs) throws MojoExecutionException {
-        String currentPythonVersion = configurePyenvOrStraightPython(usePyenv, missingRequiredToolMsgs, patchInstallScript);
+        String currentPythonVersion = "";
```
d-ryan-ashcraft: A: no reason to set this to empty string since it get assigned no matter what via the if block.  It can just be:
```
String currentPythonVersion;  //implicitly sets it to null
```

meliz19: Because `currentPythonVersion` is an input to `validateAndConfigurePythonViaPyenv` in line 98, I get an error  that `variable currentPythonVersion might not have been initialized` if I don't set it to `""`.

d-ryan-ashcraft: Ah - I didn't see that.  That should probably be handled in `validateAndConfigurePythonViaPyenv` to make it more clean.  Likely by adding a `StringUtils.isNotBlank(currentPythonVersion)` check to line 74 (or similar change).

TLDR, we don't want to have a variable that needs to be initialized specifically outside a method for that method to work realiably.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -52,62 +57,101 @@ public PyenvAndPoetrySetup(String pythonVersion, File baseDir, boolean rewriteLo
...
-                        "OR do not configure habushu to use 'pyenv' to manage the Python version!");
-            }
-
             log.debug("pyenv already installed");
         }
+        return missingRequiredToolMsgs;
+    }
+
+    private String validateAndConfigurePythonViaPyenv(File patchInstallScript) throws MojoExecutionException {
+        String currentPythonVersion = StringUtils.EMPTY;
```
meliz19, line 69: I: Refactored to define `currentPythonVersion` here instead of taking it as an empty argument from `configurePythonUsingPackageAndDependencyManager`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -52,62 +57,101 @@ public PyenvAndPoetrySetup(String pythonVersion, File baseDir, boolean rewriteLo
...
         return currentPythonVersion;
     }
 
     @Override
     protected List<String> validatePackageAndDependencyManagerInstallationAndVersion(ValidationTrackingStatus validationTracker, List<String> missingRequiredToolMsgs) 
         throws MojoExecutionException {
             missingRequiredToolMsgs = validatePoetryInstallationAndVersion(validationTracker, missingRequiredToolMsgs);
+            if (usePyenv) {
+                missingRequiredToolMsgs = validatePyenvInstallation(missingRequiredToolMsgs);
+                missingRequiredToolMsgs = validatePyenvConfiguration(missingRequiredToolMsgs);
```
meliz19, line 100: I: Moved the Pyenv configuration validation here instead of when we go to use pyenv to install/set the local Python version.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -211,18 +211,38 @@ public static AbstractPythonPackageAndDependencyManagerSetup getPythonPackageAnd
...
+        Toml toml = new Toml().read(pyProjectTomlFile);
+        String buildBackend = toml.getString("build-system.build-backend");
+        if (buildBackend != null && buildBackend.contains("poetry")) {
+            return PackageManager.POETRY;
+        } else {
+            return PackageManager.UV;
+        }
+    }
+    
+    public static String getEnvironmentVariable(String environmentVariable){
```
meliz19, line 229: I: Added this method and the two methods below it for use in `PyenvAndPoetrySetup`. These aren't specific to Pyenv or Poetry, so I thought this was a better logical place to store these methods.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -25,10 +25,10 @@
  */
 public final class HabushuUtil {
 
-	public enum PackageManager {
```
meliz19, line 28: I: This file had a mix of tabs and spaces. When I added the code below, the alignment for the methods looked off on GitHub, so I swapped out the tabs for spaces so that the entire file is uniform. This resolved the alignment issue I was seeing.

d-ryan-ashcraft (approved):

Issue #240, opened 2025-02-20
Update AbstractPythonPackageAndDependencyManagerSetup to not use abstract methods where not necessary


meliz19:
## Description
When working on #208, I added `AbstractPythonPackageAndDependencyManagerSetup` and set all methods therein to be abstract methods; however, this is not necessary. The purpose of this ticket is to remove the abstract classification from methods where it's not needed. 

## DoD
- [x] Remove the abstract classification from methods where it's not needed in `AbstractPythonPackageAndDependencyManagerSetup`.
- [x] Remove unused abstract methods from `UVSetup`

## Test Steps
Make sure that python 3.11.9 is installed on your machine. If it's not installed, you'll hit an error. Bug ticket #245 has been created to resolved this issue. 
- `mvn clean install -Pbootstrap`
- `mvn clean install`
- `mvn clean initialize -Puv`

d-ryan-ashcraft:
Test passed:
* `bootstrap` profile:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 2.18.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.018 s]
[INFO] habushu::Maven Plugin for Python Projects .......... SUCCESS [ 12.236 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  12.263 s (Wall Clock)
[INFO] Finished at: 2025-02-20T14:30:07-05:00
[INFO] ------------------------------------------------------------------------
```
* 'default' profile:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 2.18.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.007 s]
[INFO] habushu::Examples .................................. SUCCESS [  0.005 s]
[INFO] habushu::Examples::Build a Poetry Package .......... SUCCESS [ 27.552 s]
[INFO] habushu::Examples::Consume a Local Poetry Package .. SUCCESS [ 20.910 s]
[INFO] habushu::Examples::Managed Dependencies ............ SUCCESS [ 18.887 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[...]

PR #242, opened 2025-02-20
[#240] Remove unnecessary abstract methods from AbstractPythonPackageAndDependencyManagerSetup

meliz19:
Removed unused abstract methods from `UvSetup` by removing the abstract modifier from these methods in the abstract class, `AbstractPythonPackageAndDependencyManagerSetup`.

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
@@ -3,8 +3,6 @@ package org.technologybrewery.habushu;
 import com.vdurmont.semver4j.Semver;
 import com.vdurmont.semver4j.Semver.SemverType;
 
-import org.apache.commons.lang3.NotImplementedException;
-import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


d-ryan-ashcraft (approved):

chang-annie (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -78,24 +76,13 @@ protected List<String> validatePackageAndDependencyManagerInstallationAndVersion
...
 
     @Override
     protected String pythonSourceMessage() throws MojoExecutionException {
         String sourceMessage = "(managed by uv)";
         return sourceMessage;
     };
 
-    @Override
-    public void registerRepositoryToSupportAuthenticatedDependencyResolution(String repoId, String username, String password) throws MojoExecutionException, NotImplementedException {
-        // TODO: move this method out of the abstract methods as this is specific to Poetry.
```
epwilkins, line 99: A: It'd be nice to document this as the root reasoning behind the change in the ticket and/or commit. I was confused at first until I saw this.

epwilkins (approved):

Issue #241, opened 2025-02-20
Poetry Plugin Installation is not thread safe


d-ryan-ashcraft:
When executing multiple Habushu projects simultaneously via `mvnd`, a race condition can develop during calls to `PoetryCommandHelper.installPoetryPlugin`, leading to failure.  This appears to be due to the simultaneous access of Poetry's underlying `pyproject.toml` file by two instances of Habushu.

## Test Steps ##
* Run `poetry self remove poetry-monorepo-dependency-plugin`
* Run `mvn clean install -Dmaven.build.cache.skipCache=true`
* Ensure that in the `validate-pyenv-and-poetry (default-validate-pyenv-and-poetry)`:
  * The first subproject (`habushu-mixology`) installs the `poetry-monorepo-dependency-plugin`
```
[INFO] --- habushu:2.18.0-SNAPSHOT:validate-pyenv-and-poetry (default-validate-pyenv-and-poetry) @ habushu-mixology ---
[INFO] Using Python version: 3.11.9 (validated in prior build phase)
...
[INFO] Checking for updates to poetry-monorepo-dependency-plugin...
[INFO] Executing Poetry command: poetry self show plugins
[INFO] Executing Poetry command: poetry self show plugins
[INFO] Executing Poetry command: poetry self add poetry-monorepo-dependency-plugin@latest
[INFO] Using version ^1.2.0 for poetry-monorepo-dependency-plugin
```
  * NOTE: the above output shows the double-checked lock of `Executing Poetry command: poetry self show plugins` being executed twice before install
  * The second subproject (`habushu-mixology-consumer`) does NOT install the `poetry-monorepo-dependency-plugin`
```
[INFO] --- habushu:2.18.0-SNAPSHOT:validate-pyenv-and-poetry (default-validate-pyenv-and-poetry) @ habushu-mixology-consumer ---
[INFO] Using Python version: 3.11.9 (validated in prior build phase)
...
[INFO] Checking for updates to poetry-monorepo-dependency-plugin...
[INFO] Executing Poetry command: poetry self show plugins
```
* NOTE: the above output shows no installation of the plugin because it was handled previously
* Run the build again and note that neither submodule installs the plugin

woods-cpointe:
Final test: Completes successfully. Confirmed the first build checks twice and installs the plugin. Then the subsequent checks and builds do not

PR #244, opened 2025-02-20
#241 :thread: add DCL to poetry plugin add call to improve thread safety

d-ryan-ashcraft:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -217,16 +219,50 @@ public Integer executePoetryCommandAndLogAfterTimeout(List<String> arguments, in
...
+     * `poetry self add <plugin>` call and avoidance of the add altogether if the plugin already exists.
+     *
      * @param name
      * @return
      * @throws MojoExecutionException
      */
     public int installPoetryPlugin(String name) throws MojoExecutionException {
-        List<String> args = new ArrayList<String>();
+        int result = EXIT_SUCCESS;
+        if (pluginNeedsInstalling(name)) {
```
woods-cpointe, line 232: s: We are doing this if check twice. Once here and once in `performInstallPoetryPlugin`. Could we remove it from here and have have the function be `return performInstallPoetryPlugin(name);`?

woods-cpointe (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
d-ryan-ashcraft: Right - this is the double-checked locking pattern.  I split the call in half so we could minimize synchronized block to just the functionality in question.  So it is:
1. Check without synchronization; if already installed, no need to hit the synchronized block
2. Synchronize on the method that performs the installation to prevent double execution
3. Check inside the synchronization so that if a queue is built up, it short-circuits on subsequent executions.

Issue #231, opened 2025-02-18
Habushu: Support UV/Poetry for maven plugin - InstallDependenciesMojo


jaebchoi:
## Description

Support UV/Poetry implementation for InstallDependenciesMojo in Habushu maven plugin

## DOD

- [x] implement UV/Poetry switch logic for mentioned InstallDependenciesMojo
- [x] implement UV specific logic for InstallDependenciesMojo
- [x] habushu plugin for poetry can still function as before.
- [x] make sure UV InstallDependenciesMojo passes
- [x] Create Unit test for UV InstallDependenciesMojo
- [x] Add README for UV_PROJECT_ENVIRONMENT

## Testing Steps

**Prerequisite Test**
- Pull` dev` branch 
- On the root of habushu project, `run mvn clean install -Pbootstrap -Dmaven.build.cache.enabled=false` to build habushu maven plugin
- Make sure it passes
- run `mvn clean install -Pdefault -Dmaven.build.cache.enabled=false` and makes sure it passes

**Test 1 : uv index generation**
- Go to `examples/habushu-uv-package/pyproject.toml`
- check if `[[tool.uv.index]]` section exist, If it is, delete it. (there will be two) 
- run `mvn clean install -Puv -Dmaven.build.cache.enabled=false` and makes sure default-install-dependencies MOJO Passes 
NOTE: build will fail with as we have not yet implemented UV on other MOJOs
- Go to examples/habushu-uv-package/pyproject.toml and check two [[tool.uv.index]] are regenerated.

**Test 2 : Managed Dependencies update if version mismatch**
- Go to `examples/habushu-uv-package/pyproject.toml`
- Update krausening version from `dependencies =` "krausening>=20", to "krausening>=17",
- run `mvn clean install -Puv -Dmaven.build.cache.enabled=false` and makes sure default-install-dependencies MOJO Passes 
NOTE: build will fail with as we have not yet implemented UV on other MOJOs
- Go to examples/habushu-uv-package/pyproject.toml and check version is changed back to "krausening>=20" 

**Test 3 : Managed Dependencies no change if version match**
- Go to `examples/habushu-uv-package/pyproject.toml`
- Make sure version of krausening in pyproject.toml align with managed dependencies versions in examples/pom.xml
- run `mvn clean install -Puv -Dmaven.build.cache.enabled=false` and makes sure default-install-dependencies MOJO Passes 
NOTE: build will fail with as we have not yet implemented UV on other MOJOs
[...]

PR #252, opened 2025-02-24
#231 Habushu: Support UV/Poetry for maven plugin - InstallDependenciesMojo 

jaebchoi:
Issue: https://github.com/TechnologyBrewery/habushu/issues/231


examples/pom.xml
```diff
@@ -23,6 +23,20 @@
...
+                                <!-- optional - active is true by default:-->
+                                <active>true</active>
+                            </packageDefinition>
+                            <packageDefinition>
+                                <packageName>uvicorn</packageName>
+                                <operatorAndVersion>>=0.19.0</operatorAndVersion>
+                                <!-- optional - active is true by default:-->
+                                <active>true</active>
+                            </packageDefinition>
+                        </managedDependencies>
```
jaebchoi: I: Added black and uvicorn as managed dependencies to test install dependencies from dev dependencies and project dependencies

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
```diff
@@ -423,33 +419,4 @@ protected static String replaceSnapshotWithDev(String pomVersion) {
     protected static boolean isPomVersionSnapshot(String pomVersion) {
         return pomVersion.endsWith(SNAPSHOT);
     }
-
```
jaebchoi, line 426: I: This is moved to InstallDependenciesPoetry.

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/PoetryDependencyManagementSteps.java
```diff
@@ -0,0 +1,148 @@
+package org.technologybrewery.habushu;
```
jaebchoi, line 1: I: I have separated Poetry and UV specific Steps, maybe we can improve this as most of steps logics are similar between uv/poetry Ill follow up with next ticket with this https://jira.boozallencsn.com/browse/AIOPS-4016

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractInstallDependencies.java
```diff
@@ -0,0 +1,242 @@
...
+     * @param failOnManagedDependenciesMismatches  Configures if the build should be failed when managed dependency mismatches are found/
+     * @param useInProjectVirtualEnvironment         Whether to configure Poetry's virtualenvs.in-project value for this project.
+     * @param pypiRepoId   Configures repo id for pypi
+     * @param pypiRepoUrl   Configures repo url for pypi
+     * @param useDevRepository        Configures whether to use dev repository
+     * @param devRepositoryId         Configures repo id for dev repository
+     * @param devRepositoryUrl   Configures repo url for dev repository
+     * @param overridePackageVersion   Configures whether the version of the encapsulated Poetry package should be automatically managed and overridden where necessary by Habushu.
+     */
+    public AbstractInstallDependencies(File baseDir, Log log, boolean addPypiRepoAsPackageSources,  String pypiSimpleSuffix, boolean skipLockUpdate,
```
cpointe-ibllanos: A: Should the constructors for the abstract class be protected?

cpointe-ibllanos: Also may want to consolidate all of these params into a configuration class.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/VersionExtraTuple.java
cpointe-ibllanos, line 1: Any reason we didn't use an existing Pair/Tuple class?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
```diff
@@ -0,0 +1,385 @@
...
+                            String original = replacement.getValue().getOriginalOperatorAndVersion();
+                            String updated = replacement.getValue().getUpdatedOperatorAndVersion();
+                            String packageNameWithOriginal = packageNameWithExtra + original;
+                            if (line.contains(packageNameWithOriginal)) {
+                                line = line.replace(original, updated);
+                                log.info(String.format("Updated %s: %s --> %s", replacement.getValue().getPackageName(),
+                                        original, updated));
+                            }
+                        }
+                        fileContent += line + "\n";
```
cpointe-ibllanos: a: Could probably use a string builder here

cpointe-ibllanos (approved):
My feedback is non-breaking and non-substantive, so I'm approving. Good work!

README.md
```diff
@@ -849,6 +849,12 @@ Enables the automatic injection of containerization logic to the dockerfile for
 
 Default: `true`
 
+#### projectEnvironmentPath ####
```
meliz19: Q/S: Should we put this parameter in the `examples/habushu-uv-package/README.md` instead?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractInstallDependencies.java
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * UV Specific configuration that specifies the path to the directory to use for a project virtual environment.
+     */
+    protected String projectEnvironmentPath;
+
+
+
+    /**
+     * New instance for Poetry - these values are typically passed in from Maven-enabled parameters in the calling Mojo.
```
meliz19, line 130:
```suggestion
     * New instance for Python project - these values are typically passed in from Maven-enabled parameters in the calling Mojo.
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractInstallDependencies.java
meliz19: Q: It could just be my unfamiliarity with Java, but would it be cleaner/clearer to define the poetry/uv-specific methods in their own separate classes rather than within this abstract class?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesPoetry.java
```diff
@@ -0,0 +1,411 @@
+package org.technologybrewery.habushu;
```
meliz19, line 1: A: I would suggest renaming this file to `PoetryInstallDependencies` to match the naming convention we're using throughout the project. The same comment applies to the `InstallDependenciesUv` class.

README.md
d-ryan-ashcraft: Yes, and add a link to the Examples section on the main page please.

examples/habushu-uv-package/pyproject.toml
```diff
@@ -20,22 +20,21 @@ dev = [
...
-    "grpcio-tools<2.0.0,>=1.48.0",
+    "grpcio-tools>=1.48.0",
     "kappa-maki>=1.0.0",
 ]
 
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"
 
+# Added by habushu-maven-plugin at 2025-02-21T16:48:12.741544 to use https://test.pypi.org/simple/ as source PyPi repository for installing dependencies
```
d-ryan-ashcraft: A (Administrative): this comment does not match this repo.  Also, should a duplicate `[[tool.uv.index]]` block exist?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
```diff
@@ -0,0 +1,385 @@
...
+            syncCommand.add("--no-group");
+            syncCommand.add(groupName);
+        }
+
+        log.info("Installing dependencies...");
+        uvCommandHelper.executePackageManagerCommandAndLogAfterTimeout(syncCommand, 2, TimeUnit.MINUTES, UV_CLEAN_CACHE_COMMAND);
+    }
+
+    private void setUpProjectEnvironmentPath() throws MojoExecutionException {
+        // check for existing .env file, warn that this file should be tracked in version control if one does not exist
```
meliz19: Q: I'm not familiar with the use of the `.env` file. What will this do?

examples/pom.xml
d-ryan-ashcraft: I'm not sure we want black and pylint as dependencies with `uv` as I believe it will use Ruff instead.  I'd instead separate these projects, so there are specific examples for poetry and uv versus having them glommed together.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -133,329 +96,27 @@ public class InstallDependenciesMojo extends AbstractHabushuMojo {
...
-
-        } else if (rawData instanceof CommentedConfig) {
-            operatorAndVersion = TomlUtils.convertCommentedConfigToToml((CommentedConfig) rawData);
+    @Parameter(property = "habushu.projectEnvironmentPath")
+    protected String projectEnvironmentPath;
 
+    @Override
+    public void doExecute() throws MojoExecutionException, MojoFailureException {
+        if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile()) == HabushuUtil.PackageManager.POETRY){
+            var installDependenciesPoetry = new InstallDependenciesPoetry(getPythonProjectBaseDir(), getLog(), addPypiRepoAsPackageSources, pypiSimpleSuffix,
```
d-ryan-ashcraft: S (Substantive) I'm probably just old-fashioned, but I'm not a big fan of `var` usage.  If we are worried about the readability/maintainability of this code block (and we should be IMO), I would think something like passing this instance into the constructor to pull all the variables might be a better way of improving the terseness and readability of the code.  See the [Methods should not have too many parameters](https://next.sonarqube.com/sonarqube/coding_rules?open=java%3AS107&rule_key=java%3AS107) SoanrQube rule as a reference.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
jaebchoi: This will set environmental values,
Upon discussion we will remove this setup and will defer this configuration at MOJO where we call uv venv command. since venv setup configuration happpens at that command level unlike poetry have specific config command to change path.

README.md
jaebchoi: will be deleting projectEnvironmentPath per discussion and when we create uv specific, we can consider putting habushu-uv-package/README.md so forth.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesPoetry.java
```diff
@@ -0,0 +1,411 @@
...
+ * specifically by running "poetry lock" followed by "poetry install". If a
+ * private PyPi repository is defined via
+ * {@link AbstractHabushuMojo#pypiRepoUrl} (and
+ * {@link AbstractHabushuMojo#pypiRepoId}), it will be automatically added to
+ * the module's pyproject.toml configuration as a supplemental source of
+ * dependencies, if it is not already configured in the pyproject.toml
+ */
+public class InstallDependenciesPoetry extends AbstractInstallDependencies {
+
+    private static final String EQUALS = "=";
```
d-ryan-ashcraft, line 39: A: very minor, I think this got left in the mojo and no longer needs to be there

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
d-ryan-ashcraft: For python developers, it might be nice to blend in a `.env`.  But at the same time, I'd prefer we defer that to its own ticket if we want to consider it.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
```diff
@@ -0,0 +1,385 @@
...
+        String envFilePath = baseDir.getAbsolutePath() + "/.env";
+        File envFile = new File(envFilePath);
+        try {
+            if (!envFile.exists()) {
```
d-ryan-ashcraft: A: OBE, but I think the favored way to do this these days is [`try-with-resources`](https://www.baeldung.com/java-try-with-resources)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+
+    private void prepareRepositoryForInstallation(String repoId, String repoUrl) throws MojoExecutionException {
+        if (StringUtils.isNotEmpty(repoUrl) && this.addPypiRepoAsPackageSources) {
+            String pypiRepoSimpleIndexUrl;
+            try {
+                pypiRepoSimpleIndexUrl = getPyPiRepoSimpleIndexUrl(repoUrl);
+            } catch (URISyntaxException e) {
+                throw new MojoExecutionException(
```
d-ryan-ashcraft: S: if we throw `HabushuException`, we benefit from using a `RuntimeException` and don't need to add all the `throws MojoExecutionException` statements to method signatures.

meliz19 (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
```diff
@@ -0,0 +1,385 @@
...
+            }
+        }
+    }
+
+    private void executeDetailedManagedDependencyMismatchActions(Map<String, TomlReplacementTuple> replacements,
+                                                                 FileConfig pyProjectConfig, String tomlSection) {
+        Optional<Config> uVDependencies = pyProjectConfig.getOptional(tomlSection);
+        if (uVDependencies.isPresent()) {
+            List<String> foundDependencies = new ArrayList<>();
+            if (tomlSection.equals(TomlUtils.TOOL_UV_PROJECT)) {
```
d-ryan-ashcraft: A: minor, but generally best to do `constant.equals(variable)` for better null safety

d-ryan-ashcraft (commented):
Lots of good work here - thanks James!  I think we need to reconsider the duplicated structure and make just pass the Mojo into the Poery/UV-specific classes and pull the values off it instead.  Likely will lead to much more compact and maintainable code.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
meliz19: that's good to know. thanks for the link!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/VersionExtraTuple.java
jaebchoi: main reason I didn't use original TomlReplacementTuple class was that it doesn't contain [extra] information, UV has different format for optional flag,

uv is something like "uvicorn[standard]<1.0.0,>=0.18.0",

while poetry is   uvicorn = {version = ">=0.18.0", extras = ["standard"]}

poetry can be parsed with CommentedConfig Object using NightConfig but UV can't be parsed so I had to save tuple with "extra" String

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesUv.java
jaebchoi: Sounds great, will update to HabushuException

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
jaebchoi: Definitely makes sense Ill update to use explicit Object call and clean up many params!

examples/pom.xml
jaebchoi: Sounds good, updated to add managed dependencies for poetry specific ( uvicorn, black) and uv specific(krausening)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractInstallDependencies.java
jaebchoi: Good call, updated to protected and consolidated to config class!

examples/habushu-uv-package/pyproject.toml
jaebchoi: yeah, updated comments and here are some explanation of  difference in poetry[[tool.poetry.source]] vs UV [[tool.uv.index]]

Poetry:
- They have priority flag so that if you only have one supplemental tool.poetry.source, it would have default source pypi even if we haven't speficied directly.
- If you have primary priority, poetry will consider primary as default source and find supplemental if they don't find one. 


UV:
- They don't have concept of priority, if you want similar to supplemental, you need to specify all uv.index and they will scan from top to bottom.
- If they don't have any [[tool.uv.index]] at all, then it would try to fetch from public pypi by default.

examples/habushu-uv-package/pom.xml
```diff
@@ -32,7 +32,16 @@
                     <groupId>org.technologybrewery.habushu</groupId>
                     <artifactId>habushu-maven-plugin</artifactId>
                     <version>${project.version}</version>
-
+                    <configuration>
+                        <managedDependencies>
+                            <packageDefinition>
+                                <packageName>krausening</packageName>
+                                <operatorAndVersion>==20</operatorAndVersion>
```
d-ryan-ashcraft: A: `==` or `>=`?  I think we generally just want to use the python pattern of `>=`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesConfigurations.java
d-ryan-ashcraft, line 1: S: why not just pass the `mojo` instance into the `AbstractInstallDependencies` instance itself (basically - we implement the decorator pattern for these mojos)?  It eliminates the need to duplicate all this configuration and keep it in sync w/ the mojo over time.  Seems like it might just then become something like:
```
    protected AbstractInstallDependencies(InstallDependenciesMojoInterface mojo) {
        this.mojo = mojo;
    }

   void someMethod() {
        // used to be: foo(pypiSimpleSuffix); - but then becomes:
        foo(mojo. pypiSimpleSuffix);
   }
```

I'd be fine committing this as is IFF we come back in the next ticket and update it.  This will need to be a common pattern used in pretty much every mojo, so I think we need to nail it.

d-ryan-ashcraft (approved):
One substantive/critical comment, but we can circle back in the next ticket for that.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesConfigurations.java
jaebchoi: Sounds good! https://jira.boozallencsn.com/browse/AIOPS-4016 is ticket that Ill be taking next will cover refactoring to decorator pattern and refactor BDD test for installDependenciesMOJO

Issue #225, opened 2025-02-14
Update the desired python version defaulting pattern for pre-existing projects


meliz19:
## Description
Currently the desired python version is always determined by the pom config `pythonVersion` and its default. If a pre-existing python app starts using Habushu and does not set the config, it will override what the project is set up to use. To alleviate this pain point, we need to refactor what we are expecting the desired python version to be. 

## DoD
The new logic is as follows:
If the pom config `pythonVersion` is set then it should always take priority as being the desired python version. If it is not set then the current projects python version should take priority. This is determined by the `.python-version`, the venv, or the set pyenv version. If neither of these are set then the default python version should take priority.

* If the pom config version is set then it will take priority
* If the pom version is not set then the current projects python version will take priority
* If neither the pom or the current project are set then the default will take priority
* Update the pom config to have strong wording telling the users is will override the version the project is currently using
* Ensure strong logging stating where and when we are looking for python versions
* Update the readme to indicate the new default python version - 3.12.9
* Create a new config value `defaultPythonStrategy`:
  * PYTHONVERSION - the current project version takes priority (default config)
  * POM - the pom default takes priority over the existing projects version (existing poetry pattern)
* Create a new example project testing the `defaultPythonStrategy` config

## Testing
* Build habushu
  * `mvn clean install -Pbootstrap`
  * `mvn clean install`
* Verify the build succeeds

Testing the UV behavior
* Create a UV app - `uv init 225-test`
* Pin the python version to 3.12.7 - `uv python pin 3.12.7`
  * Can install the version with `uv python install 3.12.7`
* Add the Habushu plugin by copying the attached pom.xml - [pom.xml.zip](https://github.com/user-attachments/files/18972753/pom.xml.zip)
* Build and verify the python version used in the build is 3.12.7
* Update the Habushu config `pythonVersion` to set the python version to 3.12.8
* Build and verify the python version used in the build is 3.12.8
* Delete the `pythonVersion` config and the `.python-version` file
* Build and verify the python version used in the build is 3.12.9
* Pin the python version to 3.12.7
* Update the Habushu config `defaultPythonStrategy` to `POM`
* Build and verify the python version is 3.12.9

woods-cpointe:
Refactored from the bug to and enhancement on the expected behavior of the desired python version and its installation.

woods-cpointe:
DoD completed on 2/24 with @meliz19

meliz19:
[...]

PR #255, opened 2025-02-26
[#225] Update python version defaulting

woods-cpointe:
Updating python version overrides for an existing python project. If the pythonVersion config is set then it will still be the desired version but if it is not then we will delegate the python version to the existing projects version. This can be updated by the new config defaultPythonStrategy.
Created example project for defaultPythonStrategy
#225

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,8 @@ mvn clean install -Dhabushu.pythonVersion=3.10.4
 
 #### pythonVersion ####
 
-The desired version of Python to use.
+The desired version of Python to use. If `pythonVersion` is not set then Habushu will use the `defaultPythonStrategy`
+configuration to determine where to get the default Python version.
 
 Default: `3.12.9`
 
@@ -323,6 +324,14 @@ Default: `3.12.9`
 This configuration applies only to Poetry projects. If true, Habushu will delegate to `pyenv` for managing and (if needed) installing the specified version of Python. If false, Habushu will look for the desired version of Python on the `PATH`. If Python is not found or if the version does not match the configured `pythonVersion`, the build will fail.
 
 Default: `true`
+
+#### defaultPythonStrategy ####
+
+When `pythonVersion` is not explicitly specified, this strategy helps decide where to get the default version. If set to 
+`PYTHONVERSION`, then Habushu will use the existing project's Python version. If set to `POM`, then Habushu will use 
+the default value for `pythonVerion`
+
+Default: `PYTHONVERSION`
 
 #### behaveOptions ####
 
@@ -883,6 +892,8 @@ configuration options.
 - [Adding Habushu to an Existing Poetry Project](./examples/add-habushu-to-new-or-existing-poetry-project/README.md) - Adding Habushu to an existing Poetry project
 - [Managed Dependencies](./examples/habushu-managed-dependencies/README.md) - supports common definition of dependency 
   versions across Maven modules
+- [Default Python Strategy](./examples/habushu-default-python-strategy/README.md) - Handles the strategy for setting 
+  the default python version if no version is explicitly set in `pythonVersion`
 
 ### Maven Reactor Integration ###
 Optionally, Habushu supports partial builds via the Maven Reactor. This allows functionality such as `-rf` (resume from)
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
@@ -5,6 +5,7 @@ import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.logging.Log;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.technologybrewery.habushu.util.DefaultPythonStrategy;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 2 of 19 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonPackageAndDependencyManagerMojo.java
```diff
@@ -34,9 +35,15 @@ public class ValidatePythonPackageAndDependencyManagerMojo extends AbstractHabus
     /**
      * The desired version of Python to use.
      */
-    @Parameter(defaultValue = HabushuUtil.PYTHON_DEFAULT_VERSION_REQUIREMENT, property = "habushu.pythonVersion")
```
woods-cpointe, line 37: i: In order to determine if the param was set we need to remove the default so it can be null. Setting the default now in the exec.

README.md
```diff
@@ -336,16 +336,26 @@ mvn clean install -Dhabushu.pythonVersion=3.10.4
...
 
 #### usePyenv ####
 
 If true, Habushu will delegate to `pyenv` for managing and (if needed) installing the specified version of Python. If false, Habushu will look for the desired version of Python on the `PATH`. If Python is not found or if the version does not match the configured `pythonVersion`, the build will fail.
 
 Default: `true`
 
+#### defaultPythonStrategy ####
+
+Where to get the default Python version if no version is explicitly set in `pythonVersion`. If set to
```
aaron-gary: A: Recommend rewording the first sentence
"When pythonVersion is not explicitly specified, this strategy helps decide where to get the default version."
or something like that

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
```diff
@@ -26,6 +27,18 @@ public abstract class AbstractPythonPackageAndDependencyManagerSetup {
      */
     protected String pythonVersion;
 
+    /**
```
aaron-gary, line 30: Q: Are there any tests that need to be updated to account for this new behavior?

aaron-gary (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
woods-cpointe: The new example project does test this pattern but there are no unit test  for this specific mojo

README.md
```diff
@@ -324,6 +325,14 @@ This configuration applies only to Poetry projects. If true, Habushu will delega
 
 Default: `true`
 
+#### defaultPythonStrategy ####
+
+When `pythonVersion` is not explicitly specified, this strategy helps decide where to get the default version. If set to 
+`PYTHONVERSION`, then Habushu will use the existing projects Python version. If set to `POM`, then Habushu will use
```
meliz19:
```suggestion
`PYTHONVERSION`, then Habushu will use the existing project's Python version. If set to `POM`, then Habushu will use 
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPythonPackageAndDependencyManagerSetup.java
```diff
@@ -107,7 +123,9 @@ public void execute() throws MojoExecutionException {
     protected abstract List<String> validatePackageAndDependencyManagerInstallationAndVersion(ValidationTrackingStatus validationTracker, List<String> missingRequiredToolMsgs) 
         throws MojoExecutionException;
 
-    protected void finalizePythonPackageAndDependencyManagerConfiguration() throws MojoExecutionException {};
```
meliz19, line 110: Thanks for cleaning this up. I'm learning I don't need to include a `throws MojoExecutionException` in every method signature. 😂 🙈

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -123,9 +130,15 @@ private void clean() throws MojoExecutionException {
         boolean removeVenvManually = false;
 
         // TODO: This code is specific to Poetry. uv does not provide a command to remove a virtual environment like Poetry.
-        HabushuUtil.PackageManager packageManager = HabushuUtil.checkPythonPackageManager(new File(workingDirectory, "pyproject.toml"));
+        PackageManager packageManager = HabushuUtil.checkPythonPackageManager(new File(workingDirectory, "pyproject.toml"));
+        boolean isPythonVersionSet = true;
```
meliz19: Q: I'm a little confused by this logic. We set `isPythonVersionSet` to true initially. If the `PythonVersion` isn't set, we then set `isPythonVersionSet` to false, but we then set `PythonVersion`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonPackageAndDependencyManagerMojo.java
```diff
@@ -53,18 +60,27 @@ public class ValidatePythonPackageAndDependencyManagerMojo extends AbstractHabus
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
+        PackageManager packageManager = HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile());
+        boolean isPythonVersionSet = true;
```
meliz19:
```suggestion
        boolean isPythonVersionConfigurationSet = true;
```
Is this a little clearer? I'm not sure. Feel free to redefine, but I think the current param name is a little confusing. See earlier comment.

meliz19 (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
woods-cpointe: I can updating the naming to `isPythonVersionConfigurationSet`. 
In order to know if we should use defaulting logic we first need to know if the python version was explicitly set by the configs (i.e isPythonVersionConfigurationSet)

Issue #258, opened 2025-02-27
Refactor the ValidatePythonSource Mojo


meliz19:
## Description
As an extension of #191, this ticket will replace `pylint` with `ruff` for linting. Additionally, we will refactor this Mojo to implement both Poetry and uv. 

## DoD
- [x] Replace `pylint` with `ruff`
- [x] Refactor the ValidatePythonSource Mojo to more easily switch between Poetry and uv implementations
- [x] Update example projects to use `ruff` instead of `pylint`
- [x] Update relevant documentation

## Test Steps
- `mvn clean install -Pbootstrap`
- `mvn clean install`
-  `mvn org.technologybrewery.habushu:habushu-maven-plugin:3.0.0-SNAPSHOT:validate-python -Puv`  ## Poetry v2 migration is broken (#267) and is breaking the build, so testing explicitly

meliz19:
Completed DoD with @jaebchoi & @woods-cpointe

cpointe-ibllanos:
OTS completed

meliz19:
Completed OTS with @cpointe-ibllanos

woods-cpointe:
Final test: All builds pass. Output shows ruff is being installed and used

PR #262, opened 2025-02-28
[#258] - Refactored habushu to use ruff instead of pylint; add abilit…

meliz19:
…y to implement linting with either poetry or uv


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/CleanHabushuMojo.java
```diff
@@ -136,11 +136,11 @@ private void clean() throws MojoExecutionException {
             isPythonVersionConfigurationSet = false;
             pythonVersion = HabushuUtil.PYTHON_DEFAULT_VERSION_REQUIREMENT;
         }
-        AbstractPythonPackageAndDependencyManagerSetup configureTools = HabushuUtil.getPythonPackageAndDependencyManager(packageManager,
+        AbstractPythonPackageAndDependencyManagerSetup pythonPackageAndDependencyManagerSetup = HabushuUtil.getPythonPackageAndDependencyManagerSetup(packageManager,
```
meliz19, line 139: I: Renamed to clarify object purpose/meaning

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonTestMojo.java
```diff
@@ -41,8 +45,7 @@ public class ValidatePythonTestMojo extends AbstractValidateMojo {
     @Override
     public void doExecute() throws MojoExecutionException {
         if (lintTest) {
-            List<String> extraArgs = List.of("--recursive=true");
```
meliz19, line 44: I: Ruff recursively lints directories by default

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
```diff
@@ -54,6 +55,13 @@ public abstract class AbstractHabushuMojo extends AbstractMojo {
     @Parameter(defaultValue = "${project.packaging}", readonly = true, required = true)
     protected String packaging;
 
+    /**
+     * Base directory in which Python projects will be located - should always be
+     * the basedir of the encapsulating Maven project.
+     */
+    @Parameter(defaultValue = "${project.basedir}", readonly = true, required = true)
+    protected File workingDirectory;
```
meliz19, line 63: @meliz19 why'd I add this?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ValidatePythonSourceMojo.java
```diff
@@ -23,8 +23,12 @@ public class ValidatePythonSourceMojo extends AbstractValidateMojo {
 
     /**
      * Specifies disabled checkers for lint on source module.
+     * By default, Ruff enables Pyflakes (`F`) and pycodestyle (`E4`, `E7` and `E9`) codes
+     * Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
+     * McCabe complexity (`C901`) by default.
+     * This default configuration turns off Ruff's defaults with the exception of E9
      */
-    @Parameter(property = "habushu.sourceLintDisabledChecker", required = false, defaultValue = "C,R,W")
+    @Parameter(property = "habushu.sourceLintDisabledChecker", required = false, defaultValue = "E4,E7,F")
```
meliz19: I: I'm only implementing the `E9` as it will catch an io error:

> This is not a regular diagnostic; instead, it's raised when a file cannot be read from disk.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/FormatPythonMojo.java
```diff
@@ -82,16 +82,11 @@ public void doExecute() throws HabushuException {
...
-        try {
-            if (!helper.isDependencyInstalled(FORMATTER_PACKAGE)) {
-                getLog().info(
-                        String.format("%s dependency not specified in pyproject.toml - installing now...", FORMATTER_PACKAGE));
-                helper.installDevelopmentDependency(FORMATTER_PACKAGE);
-            }
-        } catch (MojoExecutionException e) {
-            getLog().error(String.format("Error installing formatter package: %s", FORMATTER_PACKAGE));
-            throw new HabushuException(e);
+    protected void downloadFormatterIfNotPresent(AbstractCommandHelper helper) {
```
meliz19, line 85: I: Removed the try/catch since the `helper.installDevelopmentDependency` throws a `HabushuException` error if one occurs.

README.md
```diff
@@ -782,15 +782,15 @@ Default: `true`
 
 #### lintSource #### 
 
-Controls if linting is enabled for the source module.
+Controls if linting is enabled for the source module. Ruff is used for linting.
 
 Default: `true`
 
 #### sourceLintDisabledChecker ####
```
d-ryan-ashcraft: C (Critical): Can we remove these so they are simply controlled by `ruff`'s configuration?  I think we likely made a mistake by putting these into `habushu` in the first place when it was `pylint`.

the enable/disable option should be the only thing in `habushu` IMO.

examples/pom.xml
```diff
@@ -73,7 +73,6 @@
         <profile>
             <id>uv</id>
             <modules>
-                <module>habushu-uv-package</module>
```
woods-cpointe, line 76: q: Why did this get removed?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -191,7 +191,7 @@ public static String getCleanVirtualEnvironmentPath(String virtualEnvFullPath) {
 
     }
 
-    public static AbstractPythonPackageAndDependencyManagerSetup getPythonPackageAndDependencyManager(
+    public static AbstractPythonPackageAndDependencyManagerSetup getPythonPackageAndDependencyManagerSetup(
```
woods-cpointe, line 208: s: Function is missing javadocs

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractValidateMojo.java
```diff
@@ -57,12 +64,24 @@ protected void runLinter(File lintDirectory, String disabledCheckers, String ena
...
-            poetryHelper.executeAndLogOutput(executeLintArgs);
+            getLog().info("Validating code using Ruff...");
+            helper.executeAndLogOutput(executeLintArgs);
         } else {
             getLog().warn(String.format("Configured linting directory (%s) does not exist - skipping...",
                     lintDirectory));
         }
     }
+
+    protected AbstractCommandHelper getCommandHelper() {
```
woods-cpointe: s: this is a good candidate to put in the base mojo. The format mojo also does this same code i think

README.md
meliz19: I think that's a great idea!

README.md
```diff
@@ -780,47 +779,9 @@ Controls whether skipped tests should be completely omitted from test reports ra
 
 Default: `true`
 
-#### lintSource ####
```
meliz19, line 784: I: Per @d-ryan-ashcraft's suggestion, removing these configs so that users use Ruff's configuration method instead.

README.md
```diff
@@ -780,47 +779,9 @@ Controls whether skipped tests should be completely omitted from test reports ra
...
-
-#### testLintEnabledChecker ####
-
-Controls the enabled checkers for linting in the test module.
-
-#### testFailOnLintErrors ####
+#### lint #### 
 
-Controls whether the build will continue if lint identifies code that violate checkers in the test module.
+Controls if linting is enabled for the module. Leverages the [ruff linter](https://docs.astral.sh/ruff/linter/) package via the package manager's `run` command.
```
meliz19, line 785: I: By default, we were running linting on both the `src` and `test` directories. Now, we use ruff's defaults and users can exclude directories or files from linting via Ruff's configuration methods, if desired.

examples/add-habushu-to-new-or-existing-poetry-project/pyproject.toml
```diff
@@ -35,3 +33,6 @@ priority = "supplemental"
 indent-style="space"
 line-ending="lf"
 quote-style="preserve"
+
+[tool.ruff.lint]
+ignore = ["E4","E7","F"]
```
meliz19, line 38: I: Added this config to each example/**/pyproject.toml to use minimal linting.

examples/pom.xml
meliz19: I removed it because it was broken, but I saw you just removed the section that was breaking the build, so I'll do the same instead.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractHabushuMojo.java
```diff
@@ -337,6 +347,24 @@ protected UvCommandHelper createUvCommandHelper() {
         return new UvCommandHelper(getPythonProjectBaseDir());
     }
 
+    /**
+     Creates a {@link PoetryCommandHelper} or {@link UvCommandHelper} that may be used to invoke Poetry or uv
+     * commands from the project's working directory.
+     *
+     * @return helper
+     */
+    protected AbstractCommandHelper getCommandHelper() {
```
meliz19, line 392: I: I left the `PoetryCommandHelper` and `UvCommandHelper` here for now bc I was going to have to change a bunch of other files. We can flow another ticket to replace these methods with the new method below.

woods-cpointe (approved):

Issue #260, opened 2025-02-28
Update Behave Mojo to Support UV projects


woods-cpointe:
Update the BehaveBddTest mojo to support both Poetry and Behave in a way that makes it easy to add more package managers in the future.

**DoD:**

- [x] Habushu build runs behave test for UV projects
- [x] Behave is installed if not currently in the project
- [x] The test logs and output are displayed in the build
- [x] Habushu build still runs behave test for Poetry projects
- [x] Update the behave docs if necessary
- [x] Correct the behave tests in the UV example project

**Testing**
* Build the project - `mvn clean install -Pbootstrap`
* Run the tests for the UV project - `mvn org.technologybrewery.habushu:habushu-maven-plugin:3.0.0-SNAPSHOT:behave-bdd-test -Puv -pl :habushu-uv-package`
  * The full build fails before getting to the tests so running them directly for now
* Verify the behave tests are ran and the outcome is printed
  * The uv command is printed - `Executing uv command: uv run behave .../habushu/examples/habushu-uv-package/tests/features --format=kappa_maki.kappa_maki_formatter:PrettyCucumberJSONFormatter --outfile=target/cucumber-reports/cucumber.json --format=progress2 --no-skipped --no-capture --no-capture-stderr --no-logcapture --tags ~integration_test`
  * The test steps results are printed - `6 steps passed, 0 failed, 0 skipped, 0 undefined`
* Build the Poetry project - `mvn clean install`
* Verify the behave tests are ran and the outcome is printed
  * The poetry command is printed - `Executing poetry command: poetry run behave .../habushu/examples/add-habushu-to-new-or-existing-poetry-project/tests/features --format=kappa_maki.kappa_maki_formatter:PrettyCucumberJSONFormatter --outfile=target/cucumber-reports/cucumber.json --format=progress2 --no-skipped --no-capture --no-capture-stderr --no-logcapture --tags ~integration_test`
* Test the test configuration
  * Set `skipTests` to true. Run and verify they tests are skipped
  * Set `behaveOptions` to `--tags test-tag`. Run and verify the behave command has the tag appended
  * Set `outputCucumberStyleTestReports` to false. Run and verify the formatting of the output has changed
  * Set `behaveExcludeManualTag` to false. Run and verify the behave command does not exclude the manual tag
* Remove behave from the project and sync the app
* Run the tests and verify behave was not found and installed

woods-cpointe:
DoD completed with @meliz19 @jaebchoi

[...]

PR #269, opened 2025-03-03
[#260] Update BehaveBddTestMojo to handle different package managers

woods-cpointe:
Created an abstract behave test class for both uv and poetry to extend. The Mojo just determines which to use. Currently their executions are the same so the main execute functionality is in the parent abstract
Signed-off-by: woods-cpointe <cwoods@cpointe-inc.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -729,7 +729,8 @@ Default: `${project.basedir}/src`
 
 #### testDirectory ####
 
-Folder in which Python test files are located - should align with Poetry's project structure conventions. Developers will typically **not** modify this property but is made available for customization to support unanticipated scenarios.
+Folder in which Python test files are located - should align with standard project structure conventions. 
+Developers will typically **not** modify this property but is made available for customization to support unanticipated scenarios.
 
 Default: `${project.basedir}/tests`
 
--- a/examples/habushu-uv-package/pom.xml
+++ b/examples/habushu-uv-package/pom.xml
@@ -48,23 +48,6 @@
             <plugin>
                 <groupId>org.technologybrewery.habushu</groupId>
                 <artifactId>habushu-maven-plugin</artifactId>
-                <!-- Utilize the run-command-in-virtual-env goal to demonstrate how
-                    commands/scripts may be executed via "poetry run" and bound to the build
-                    lifecycle phases -->
-                <executions>
-                    <execution>
-                        <configuration>
-                            <runCommandArgs>python -m grpc_tools.protoc -I=src
-                                --python_out=src/habushu_poetry_package/generated src/person.proto
-                            </runCommandArgs>
-                        </configuration>
-                        <id>generate-protobuf-bindings</id>
-                        <phase>compile</phase>
-                        <goals>
-                            <goal>run-command-in-virtual-env</goal>
-                        </goals>
-                    </execution>
-                </executions>
             </plugin>
             <plugin>
                 <groupId>net.masterthought</groupId>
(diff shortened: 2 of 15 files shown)


examples/habushu-uv-package/pom.xml
```diff
@@ -48,23 +48,6 @@
             <plugin>
                 <groupId>org.technologybrewery.habushu</groupId>
                 <artifactId>habushu-maven-plugin</artifactId>
-                <!-- Utilize the run-command-in-virtual-env goal to demonstrate how
```
woods-cpointe, line 51: i: This mojo is not implemented yet so removing the invalid test from the example project. If its actually needed then it should be added as part of that Mojo update

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractBehaveBddTest.java
```diff
@@ -0,0 +1,105 @@
...
+            this.log.info(String.format("%s dependency not specified in pyproject.toml - installing now...",
+                    BEHAVE_PACKAGE));
+            this.commandHelper.installDevelopmentDependency(BEHAVE_PACKAGE);
+        }
+    }
+
+    protected boolean hasTests(File behaveDirectory) throws MojoExecutionException {
+        if (behaveDirectory.exists()) {
+            try (Stream<Path> path = Files.list(behaveDirectory.toPath())) {
+                return path.findAny().isPresent();
```
aaron-gary: A: minor - typically it's good to only have one return statement.  i.e. could set a variable and return at the end, but this method is so small, it's not too hard to read.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractBehaveBddTest.java
```diff
@@ -0,0 +1,105 @@
...
+                throw new MojoExecutionException("Could not load behave features directory", e);
+            }
+        }
+        return false;
+    }
+
+    protected void appendBehaveOptions(List<String> executeBehaveTestArgs) {
+        if (this.behaveBddTestMojo.isOutputCucumberStyleTestReports()) {
+            this.commandHelper.installDevelopmentDependency(BEHAVE_CUCUMBER_FORMATTER);
+            executeBehaveTestArgs.add("--format=kappa_maki.kappa_maki_formatter:PrettyCucumberJSONFormatter");
```
aaron-gary: A: I noticed there was a constant for BEHAVE_CUCUMBER_FORMATTER.  Could a similar constant be used for this formatter string?

aaron-gary (approved):

README.md
```diff
@@ -729,7 +729,8 @@ Default: `${project.basedir}/src`
 
 #### testDirectory ####
 
-Folder in which Python test files are located - should align with Poetry's project structure conventions. Developers will typically **not** modify this property but is made available for customization to support unanticipated scenarios.
+Folder in which Python test files are located - should align with Poetry's and uv's project structure conventions.
```
meliz19: I (informational): uv doesn't generate a test directory like Poetry.

examples/habushu-uv-package/pom.xml
jaebchoi: Good Call!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/AbstractCommandHelper.java
```diff
@@ -152,6 +152,23 @@ public Integer executePackageManagerCommandAndLogAfterTimeout(List<String> argum
         }
     }
 
+    /**
+     * Returns whether the specified dependency package is installed within this
+     * project's virtual environment (and pyproject.toml).
+     *
+     * @param packageName the package name to check
+     * @return weather the package is installed
```
jaebchoi:
```suggestion
     * @return whether the package is installed
```

jaebchoi (approved):
looks good, nit comments

Issue #270, opened 2025-03-04
Replace black with ruff: remove mentions of black, pylint, add missing documentation to README


cpointe-ibllanos:
# Context
#191 added support for `ruff` as the formatter in place of `black`. We should verify that

# Definition of Done
1. [ ] All mentions of `black` and `pylint` are removed
2. [ ] Newly added mojo parameters for `useFormatter` are included in the README

cpointe-ibllanos:
OTS completed with @aaron-gary

jaebchoi:
Test Passed:
- Verified there is no call for Black or pylint 
- Added Mojo param useFormatter exist in README
- Verified -PBootstrap default and uv profile build succeed

PR #278, opened 2025-03-06
[#270] Remove mentions of "black" and "pylint" dependencies and add "useFormatter" mojo param to README

cpointe-ibllanos:
#270

### DIFF
--- a/examples/habushu-managed-dependencies/pyproject.toml
+++ b/examples/habushu-managed-dependencies/pyproject.toml
@@ -13,10 +13,8 @@ dynamic = ["version", "dependencies"]
 
 [tool.poetry.group.dev.dependencies]
 behave = ">=1.2.6"
-pylint = ">=3.1.0"
-black = ">=24.0.0"
+ruff = ">=0.9.9"
 kappa-maki = ">=1.0.0"
-ruff = "^0.9.9"
 
 [build-system]
 requires = ["poetry-core>=1.6.0"]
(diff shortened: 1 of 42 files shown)


habushu-maven-plugin/src/test/resources/specifications/dependency-management-poetry.feature
```diff
@@ -55,14 +55,14 @@ Feature: Test Poetry dependency management capabilities to help align package ve
       | packageBarTest | ^0.9.0             |
 
   Scenario Outline: Inactive managed dependencies are skipped
-    Given a Habushu configuration with a poetry managed dependency of "<package>" and "<operatorAndVersion>"
+    Given a Habushu configuration with a poetry inactive managed dependency of "<package>" and "<operatorAndVersion>"
     When Habushu executes with poetry
-    Then the poetry pyproject.toml file has updates
+    Then the poetry pyproject.toml file has no updates
```
cpointe-ibllanos, line 60: I: Seems this was using the wrong test steps. Kinda curious how it was passing before.

examples/habushu-managed-dependencies/pyproject.toml
```diff
@@ -13,10 +13,8 @@ dynamic = ["version", "dependencies"]
 
 [tool.poetry.group.dev.dependencies]
 behave = ">=1.2.6"
-pylint = ">=3.1.0"
-black = ">=24.0.0"
-kappa-maki = ">=1.0.0"
 ruff = "^0.9.9"
```
jaebchoi: A: do we need to update >= just like what we changed in other pyprojec.toml?

jaebchoi (approved):

examples/habushu-managed-dependencies/pyproject.toml
cpointe-ibllanos: It's not necessary, but you're right it should be more consistent. I switched most mentions to use `>=` where it makes sense.

Issue #277, opened 2025-03-06
Update Run command in venv mojo to support uv


woods-cpointe:
Update the RunCommandInVirtualEnv mojo to support both Poetry and uv in a way that makes it easy to add more package managers in the future.

## DoD
- [x] RunCommandInVirtualEnvMojo completes for both uv and Poetry
- [x] Create a dedicated example project for this config and remove it from the base poetry example
- [x] Update the docs to indicate it is not poetry specific

## Testing
* Build the plugin - `mvn clean install -Pbootstrap`
* Build the example projects - `mvn clean install`
* Build the uv example projects - `mvn clean install -Puv`
* Verify the build completes successfully
* Verify the `runCommandArgs` have been moved to the example readme

woods-cpointe:
DoD completed with @jaebchoi and @aaron-gary

woods-cpointe:
OTS completed with @colinpalmer-pro

jaebchoi:
Test Passed

- [x] Build the plugin - mvn clean install -Pbootstrap
- [x] Build the example projects - mvn clean install
- [x] Build the uv example projects - mvn clean install -Puv
- [x] Verify the build completes successfully
- [x] Verify the runCommandArgs have been moved to the example readme

PR #279, opened 2025-03-07
[#277] Update run command mojo to use abstract package manager

woods-cpointe:
[#277] Create dedicated eample project


examples/habushu-running-python-scripts/src/habushu_running_python_scripts/generated/__init__.py
aaron-gary, line 1: A: Does this generated file need to be committed?

woods-cpointe: No, we want the protobuf bindings to be generated every build by the python command in the pom config

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
```diff
@@ -32,12 +32,12 @@ public class RunCommandInVirtualEnvMojo extends AbstractHabushuMojo {
 
     @Override
     public void doExecute() throws MojoExecutionException, MojoFailureException {
-	PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
-	List<String> poetryRunCommandArgs = new ArrayList<>(Arrays.asList(StringUtils.split(runCommandArgs)));
-	poetryRunCommandArgs.add(0, "run");
+        AbstractCommandHelper commandHelper = getCommandHelper();
```
aaron-gary: Q: It seems strange to instantiate a class that has "Abstract" in the name.  Is there an interface that should be used here instead?

woods-cpointe: Great point. We can create in interface for the command helper and instantiate that instead

examples/habushu-running-python-scripts/src/habushu_running_python_scripts/generated/__init__.py
meliz19: I'm pretty sure I tried to gitignore this file, but it breaks the build because the `generated` directory has to already exist.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
meliz19: Hmm, that's a good point. I created that class, but I also didn't like that Abstract was in the name. Maybe this is a candidate for an interface instead. 🤔

meliz19 (approved):

aaron-gary (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/RunCommandInVirtualEnvMojo.java
aaron-gary: @meliz19 - Having a class with Abstract in the name is fine, but should only be used when we extend it in a concrete class. When it comes time to instantiate the class, the variable should be declared using the interface.  This keeps the Abstract class encapsulated and hidden from the user.

examples/habushu-running-python-scripts/src/habushu_running_python_scripts/generated/__init__.py
woods-cpointe: Ah sorry I think I misunderstood the comment. The init does need to exist for the issue Margaret mentioned. The other file that gets generated on build is `person_pb2.py` which we do not want to commit

PR #285, opened 2025-03-13
[#284] :thead: catch race condition around poetry's pyproject.toml fi…

### LINKED ISSUE #284: Running habushu with mvnd with many simultaneous executions results in errors ~10% of the time
When running a project with *+ simultaneous habushu executions, the following error is encountered:

```
[INFO] --- habushu:2.18.0:validate-pyenv-and-poetry (default-validate-pyenv-and-poetry) @ booz-allen-public-license-python ---
[INFO] Using Python 3.11.4 (managed by pyenv)
[INFO] Found Poetry 2.1.1
[INFO] Configuring Poetry to use the pyenv-activated Python binary...
[INFO] Executing Poetry command: poetry config --local virtualenvs.use-poetry-python false
[INFO] Checking for updates to poetry-monorepo-dependency-plugin...
[INFO] Executing Poetry command: poetry self show plugins
[ERROR] 
The Poetry configuration is invalid:
  - Either [project.name] or [tool.poetry.name] is required in package mode.
  - Either [project.version] or [tool.poetry.version] is required in package mode.
```

This is an improvement over 2.17.x, but still would be nice to make it work all the time.

## Steps to Recreate:
* Check out a project with a lot of habushu 2.18.0 projects and run `mvnd initialize`
   * you may need to turn caching off, if you are using it on the project: `-Dmaven.build.cache.skipCache=true`
* If you run a project with 5+ modules that all use habushu somewhat simultaneously, you'll hit this error about one in ten runs

## Test Steps
* Check out a project with a lot of habushu 2.18.0 projects and run `mvnd initialize`
   * you may need to turn caching off, if you are using it on the project: `-Dmaven.build.cache.skipCache=true`
* Update it to use habushu 2.18.1-SHAPSHOT (ensure you have built this locally as well)
* Run the command at least ten times and verify you don't see the error. 
* Run `poetry self remove poetry-monorepo-dependency-plugin`
* Run `mvnd initialize` again, verify no errors occur

d-ryan-ashcraft:
…le and retry once on occurance

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
@@ -20,6 +20,7 @@ import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.technologybrewery.habushu.HabushuException;
 
 /**
  * Facilitates the execution of Poetry commands.
@@ -36,6 +37,8 @@ public class PoetryCommandHelper {
 
     protected File workingDirectory;
 
+    protected volatile String showPluginsResult;
+    protected static final Object pluginSystemLock = new Object();
 
     public PoetryCommandHelper(File workingDirectory) {
         this.workingDirectory = workingDirectory;
@@ -236,26 +239,26 @@ public class PoetryCommandHelper {
         return result;
     }
 
-    private synchronized int performInstallPoetryPlugin(String name) throws MojoExecutionException {
+    private int performInstallPoetryPlugin(String name) throws MojoExecutionException {
         int result = EXIT_SUCCESS;
-        if (pluginNeedsInstalling(name)) {
-            List<String> args = new ArrayList<>();
-            args.add("self");
-            args.add("add");
-            args.add(name);
-            result = this.executeAndLogOutput(args);
-
+        List<String> args = new ArrayList<>();
+        args.add("self");
+        args.add("add");
+        args.add(name);
+        synchronized (pluginSystemLock) {
+            if (pluginNeedsInstalling(name)) {
+                result = this.executeAndLogOutput(args);
+            }
         }
+
+        // un-cache known set of plugins:
+        showPluginsResult = null;
 
         return result;
     }
 
     private boolean pluginNeedsInstalling(String name) throws MojoExecutionException {
-        List<String> args = new ArrayList<>();
-        args.add("self");
-        args.add("show");
-        args.add("plugins");
-        String showPluginsResult =  this.execute(args);
+        executeShowPlugins();
 
         String pluginNameWithoutVersion = name;
         if (name.contains(VERSION_DELIMITER)) {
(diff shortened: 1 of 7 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -36,6 +37,7 @@ public class PoetryCommandHelper {
 
     protected File workingDirectory;
```
d-ryan-ashcraft, line 39: I (Informational): we are doing this in a low-grade fashion.  Cache result of which plugins are there (since we don't support different versions of poetry being used in a single project).  If this gets set a couple times, so be it.  Still executes faster that way than with a locking strategy.  But use `volatile` keywork to help out.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -36,6 +37,8 @@ public class PoetryCommandHelper {
...
 
+    protected volatile String showPluginsResult;
+    protected static final Object pluginSystemLock = new Object();
```
d-ryan-ashcraft, line 41: I: adding a specific lock file on plugin install to isolate the synchronization down to the specific command and provide a common lock for a second try on showing plugins.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -265,6 +268,30 @@ private boolean pluginNeedsInstalling(String name) throws MojoExecutionException
         return !showPluginsResult.contains(pluginNameWithoutVersion);
     }
 
+    private void executeShowPlugins() throws MojoExecutionException {
+        if (StringUtils.isBlank(showPluginsResult)) {
+            List<String> args = new ArrayList<>();
+            args.add("self");
+            args.add("show");
+            args.add("plugins");
+            try {
```
d-ryan-ashcraft, line 277: I: FA - let's throw caution to the wind and assume it's all going to go according to plan.  This is a **lot** faster (~2-5 seconds).  If it blows up, we'll deal with things more carefully.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PoetryCommandHelper.java
```diff
@@ -265,6 +268,30 @@ private boolean pluginNeedsInstalling(String name) throws MojoExecutionException
...
+                    if (StringUtils.isBlank(showPluginsResult)) {
+                        showPluginsResult = this.execute(args);
+                    }
+            } catch (HabushuException e) {
+                logger.info("Plugin status could not be determined. This is normally a race condition on another"
+                        + " thread touching poetry's underlying pyproject.toml, trying one more time...");
+                // let's be more careful round two and assume that
+                synchronized (pluginSystemLock) {
```
d-ryan-ashcraft, line 285: I: FO - okay, if we caught this at the right moment when the race condition occurred, be careful and make sure it works for round 2

cpointe-ibllanos (approved):

PR #289, opened 2025-03-14
[#288] Refactor containerize dependencies to check full project scope for habushu dependency

### LINKED ISSUE #288: Simplify Habushu dependency resolution in containerization mojo
# Context

Our current containerization process is confusing and misleading.Our current implementation is Poetry focused and requires the user to know and provide what projects constitute Habushu dependencies, the complexity of which scares potential users from using the feature. The goal is to streamline this behavior so that Habushu dependencies are resolved even if they are not provided as part of limited maven project scope using commands like `-pl`.

Before doing any refactoring or redesigning the mojo for Habushu, we also want to look into how UV expects a developer to containerize its dependencies and what existing methods exist for it. 

# Definition of Done

- ✅ Investigate whether UV's venv dependency and wheel file resolution is identical enough to Poetry's to leverage current logic in the existing containerization mojo
- ✅ Change mojo logic to leverage existing wheel files from previous build rather than forcing parameterization of individual dependencies in `-pl` statement for every build
  - ℹ This was meant to handle an edge case that comes up internally, but it's counter-intuitive to use otherwise
  - ℹ️ We are ok if wheel files fall behind the module version
  - ✅ Fix bdd tests that broke as a result
- Should there be disagreements on this approach and a push to maintain the current functionality, we can:
  - ✅ at least include a message with direction as to why a build providing just the main module in the `pl` command would fail and provide the dependency names needed for a successful build
- ❌ Adapt mojo to be package manager agnostic and support uv
  - Adapt dockerfile generation to leverage Velocity templating as well
  - To be handled in #207 

# Test Steps
1. Run `mvn clean install -Pbootstrap` without failure
2. Run `mvn clean install -pl :examples,:add-habushu-to-new-or-existing-poetry-project`
3. Run `mvn clean install -pl :habushu-poetry-containerize`
  - The build should succeed
  - You should *NOT* see a failure and error that says:
    - "Habushu project to containerize was not included in the provided projects (-pl). Ensure the Habushu project is in the build and the POM dependencies are configured correctly."

cpointe-ibllanos:
#288

### DIFF
--- a/examples/habushu-poetry-containerize/README.md
+++ b/examples/habushu-poetry-containerize/README.md
@@ -1,0 +1,64 @@
+### Leveraging the containerize-dependencies Goal to Prepare a Containerized Virtual Environment ###
+The `containerize-dependencies` goal will collect a single `habushu` dependency specified in the project's `pom.xml`,
+including all transitive habushu-packaged dependencies. After collecting the set of necessary dependencies, Habushu will
+copy the project files for each dependency to a staging directory, while preserving the original structure of the
+dependency modules to ensure that any path-based dependencies can be leveraged as-is. This directory can then be copied
+onto a Docker container and used to create a virtual environment capable of running the target Habushu project.
+
+The plugin will automatically inject logic for building and using this virtual environment into a pre-existing
+Dockerfile by default. The `dockerfile` configuration must be set to the target Dockerfile. To disable the Dockerfile
+update altogether, set the `updateDockerfile` configuration to `false`. Because the virtual environment that is created
+is dependent on the platform for which it is built, Habushu defaults to using `python:3.12` to build the virtual
+environment and  `python:3.12-slim` as the final image that packages/runs the virtual environment. This can be
+customized with the `dockerBuilderBase`, `dockerFinalBase`, `dockerUser`, `dockerPoetryVersion`, `dockerPoetryPluginBundleVersion`, and `dockerPoetryMonorepoDependencyPluginVersion` configurations, but care must be taken to
+ensure the builder image platform is sufficiently similar to the final image platform so that the virtual environment is
+compatible.
+
+```xml
+<plugin>
+    <groupId>org.technologybrewery.habushu</groupId>
+    <artifactId>habushu-maven-plugin</artifactId>
+    <executions>
+        <execution>
+            <id>containerize-deps</id>
+            <goals>
+                <goal>containerize-dependencies</goal>
+            </goals>
+        </execution>
+    </executions>
+    <configuration>
+        <dockerfile>${project.basedir}/src/main/docker/Dockerfile</dockerfile>
+    </configuration>
+</plugin>
+```
+
+
+To control the exact insertion location of the Dockerfile build logic, use the `#HABUSHU_BUILDER_STAGE`and
+`#HABUSHU_FINAL_STAGE` comment tags in the `Dockerfile` in the preferred location. The `#HABUSHU_BUILDER_STAGE` tag will
+be replaced with the builder stage logic to copy the generated dependencies files onto a Docker container and then create
+the virtual environment by building the target Habushu project. The `#HABUSHU_FINAL_STAGE` tag will be replaced with the
+final stage logic to copy over the built virtual environment to the final Docker build stage.
+```dockerfile
+#HABUSHU_BUILDER_STAGE
+
+FROM redhat/ubi9-minimal:latest AS builder
+RUN microdnf install -y python3.11
+
+#HABUSHU_FINAL_STAGE
+
+ENTRYPOINT ["/opt/venv/bin/python3.11", "-m", "pets.main", "--enable_docs_url", "True"]
+```
+
+The plugin will only examine dependencies that are of the type `habushu` in the dependencies block of the `pom.xml` file.
+```xml
+<dependencies>
+    <dependency>
+        <groupId>your-group-id</groupId>
+        <artifactId>your-artifact-id</artifactId>
+        <version>your-version</version>
+        <type>habushu</type>
+    </dependency>
+</dependencies>
+```
+Any transitive monorepo dependencies should also use this convention to ensure they are captured by the plugin.
+
--- a/examples/habushu-poetry-containerize/pom.xml
+++ b/examples/habushu-poetry-containerize/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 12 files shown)


pom.xml
```diff
@@ -56,6 +56,7 @@
     <properties>
         <version.cucumber.reporting>5.8.4</version.cucumber.reporting>
         <version.cucumber.reporting.plugin>5.8.5</version.cucumber.reporting.plugin>
+        <skipTests>false</skipTests>
```
cpointe-ibllanos, line 59: Kind of curious if this is worth defining in the TechnologyBrewery parent pom.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -239,11 +244,8 @@ protected ProjectCollectionResult getHabushuProjects() {
      * @param collectionResult the result object to add the projects to
      */
     protected ProjectCollectionResult collectHabushuDependenciesAsProjects(MavenProject currentProject, ProjectCollectionResult collectionResult) {
-        Set<String> habushuDeps = currentProject.getDependencies().stream()
-                .filter(d -> HabushuUtil.HABUSHU.equals(d.getType()))
-                .map(ContainerizeDepsMojo::toGav)
-                .collect(Collectors.toSet());
-        for (MavenProject project : getSession().getProjects()) {
+        Set<String> habushuDeps = getProjectHabushuDependencies(currentProject);
+        for (MavenProject project : getSession().getAllProjects()) {
```
cpointe-ibllanos, line 248: Main change: we now check all projects when determining the habushu dependency to containerize.

examples/habushu-poetry-containerize/README.md
```diff
@@ -0,0 +1,64 @@
...
+The `containerize-dependencies` goal will collect a single `habushu` dependency specified in the project's `pom.xml`,
+including all transitive habushu-packaged dependencies. After collecting the set of necessary dependencies, Habushu will
+copy the project files for each dependency to a staging directory, while preserving the original structure of the
+dependency modules to ensure that any path-based dependencies can be leveraged as-is. This directory can then be copied
+onto a Docker container and used to create a virtual environment capable of running the target Habushu project.
+
+The plugin will automatically inject logic for building and using this virtual environment into a pre-existing
+Dockerfile by default. The `dockerfile` configuration must be set to the target Dockerfile. To disable the Dockerfile
+update altogether, set the `updateDockerfile` configuration to `false`. Because the virtual environment that is created
+is dependent on the platform for which it is built, Habushu defaults to using `python:3.11` to build the virtual
```
meliz19:
```suggestion
is dependent on the platform for which it is built, Habushu defaults to using `python:3.12` to build the virtual
```

The default python version for habushu has been updated to 3.12.9.

examples/habushu-poetry-containerize/README.md
```diff
@@ -0,0 +1,64 @@
...
+environment and  `python:3.11-slim` as the final image that packages/runs the virtual environment. This can be
```
meliz19:
```suggestion
environment and  `python:3.12-slim` as the final image that packages/runs the virtual environment. This can be
```

examples/habushu-poetry-containerize/README.md
```diff
@@ -0,0 +1,64 @@
...
+final stage logic to copy over the built virtual environment to the final Docker build stage.
+```dockerfile
+#HABUSHU_BUILDER_STAGE
+
+FROM redhat/ubi9-minimal:latest AS builder
+RUN microdnf install -y python3.11
+
+#HABUSHU_FINAL_STAGE
+
+ENTRYPOINT ["/opt/venv/bin/python3.11", "-m", "pets.main", "--enable_docs_url", "True"]
```
meliz19, line 49:
```suggestion
ENTRYPOINT ["/opt/venv/bin/python3.12", "-m", "pets.main", "--enable_docs_url", "True"]
```

examples/habushu-poetry-containerize/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+                <executions>
+                    <execution>
+                        <id>dockerize</id>
+                        <goals>
+                            <goal>containerize-dependencies</goal>
+                        </goals>
+                        <phase>prepare-package</phase>
+                        <configuration>
+                            <dockerfile>src/resources/Dockerfile</dockerfile>
+                            <dockerBuilderBase>docker.io/python:3.11</dockerBuilderBase>
```
meliz19, line 58:
```suggestion
                            <dockerBuilderBase>docker.io/python:3.12</dockerBuilderBase>
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -166,7 +166,12 @@ protected Path stageHabushuProjects(Path sourceRoot, ProjectCollectionResult pro
             stageSourcesForProject(sourceRoot, destRoot, sourceFileSet, relativeProjectPath);
         }
         if( primaryProjectPath == null ) {
-            throw new HabushuException("Primary project was not included in the set of projects. Ensure the Habushu project is in the build and the POM dependencies are configured correctly.");
+            String habushuDependency = getProjectHabushuDependencies(getProject()).toString().split(":")[1];
+            StringBuilder message = new StringBuilder()
+                    .append("Primary project was not included in the set of projects.")
```
meliz19: S: I'm not sure what we mean by this error message. I'm not sure if this is maven terminology or not. Could we rephrase it to something along the lines of "The <module/project name> was not included as a habushu dependency. Ensure the <module/project name> is included as a module in the appropriate parent pom and as a dependency in the appropriate dependent pom." (I'm not sure if that's the right sentiment or not, but hopefully that gives you an idea of what might be more helpful for Python developers who encounter this error message.)

cpointe-ibllanos: Good call. Since the README specifies it is meant to containerize a single habushu project as its dependency, I much prefer the language used to more closely match that.

habushu-maven-plugin/pom.xml
```diff
@@ -242,6 +242,9 @@
             <plugin>
                 <groupId>net.masterthought</groupId>
                 <artifactId>maven-cucumber-reporting</artifactId>
+                <configuration>
+                    <skip>${skipTests}</skip>
```
woods-cpointe, line 246: q: Is there ever a time where this is true or where we want it to be true?

cpointe-ibllanos: Only when adding it manually for testing purposes as per `-DskipTests`. It makes the build fail without the config as the plugin checks for a report that does not exist.

woods-cpointe (approved):

Issue #290, opened 2025-03-14
Remove unnecessary complexity when setting python version in uv projects


meliz19:
## Description
When I first wrote the logic for getting the Python version from a uv project, I incorrectly thought that uv projects did not have a `.python-version` file generated on project inception. However, I have since learned that that is incorrect. To simplify the logic of the python version selection process, I am going to remove the checks regarding the `.venv` and only check a uv project's python version via the `uv python pin` command.

## DoD
- [x] Remove unnecessary logic regarding extracting the python version from a `.venv` for uv projects

## Test Steps
- `mvn clean install -Pbootstrap`
- In one of the uv example projects, e.g., `examples/add-habushu-to-new-or-existing-uv-project`, remove the `.python-version` file. 
- From the root project level, run  `mvn clean install -Dmaven.build.cache.skipCache=true -Puv`
- Verify you seen the following lines in the log for the project in which you deleted the `.python-version` file.
```
[INFO] Executing uv command: uv python pin
[ERROR] error: No pinned Python version found

[INFO] The Python version is not currently set to the desired version. Setting the version...
[INFO] Executing uv command: uv python pin 3.11.9
```

meliz19:
Completed DoD/OTS with @jaebchoi

woods-cpointe:
FInal test: Completed successfully. Version output matches testing steps

PR #291, opened 2025-03-17
[#290] Remove unnecessary uv python version setting logic

meliz19:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -79,16 +82,8 @@ public Pair<Boolean, String> getIsUvInstalledAndVersion() {
      *
      * @return
      */
-    public String getCurrentPythonVersion() throws MojoExecutionException {
-        String currentPythonVersion = pythonVersionFileContents();
```
meliz19, line 83: I: The `uv python pin` command reads the python version from the `.python-version` if it exists, so this custom command isn't necessary.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -79,16 +82,8 @@ public Pair<Boolean, String> getIsUvInstalledAndVersion() {
...
-        if (currentPythonVersion.isEmpty()) {
-            currentPythonVersion = pythonVersionFromVirtualEnvironment();
```
meliz19, line 85: I: When uv projects are initialized, the `.python-version` file is generated, so this is an unnecessary edge case to consider that's making the code needlessly complex.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -145,27 +139,33 @@ public List<String> createSyncCommand() {
         return arguments;
     }
 
-    public List<String> createRunCommand(List<String> additional_arguments) {
```
meliz19, line 148: The `createRunCommand` isn't used anywhere, so I'm deleting it.

d-ryan-ashcraft (approved):

Issue #294, opened 2025-03-17
Task: Clean up initialization of abstract classes


woods-cpointe:
We are initialization abstract classes when we should be using the concrete class or an interface if the desired child is not known

**DoD:**
* Remove direct calls to `AbstractPythonPackageAndDependencyManagerSetup`

**Testing:**
* Build the plugin - `mvn clean install -Pbootstrap`
* Build the example projects `mvn clean install -Pdefault,uv`

woods-cpointe:
DoD and OTS completed with @meliz19

d-ryan-ashcraft:
## Test Results: ✅ PASSED
- ✅ Build the plugin - `mvn clean install -Pbootstrap`
- ✅ Build the example projects `mvn clean install -Pdefault,uv`

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for habushu::Root 3.0.0-SNAPSHOT:
[INFO] 
[INFO] habushu::Root ...................................... SUCCESS [  0.066 s]
[INFO] habushu::Examples .................................. SUCCESS [  0.011 s]
[INFO] habushu::Examples::Add Habushu to a new or existing Poetry project SUCCESS [ 13.380 s]
[INFO] habushu::Examples::Build a Poetry Package .......... SUCCESS [ 14.379 s]
[INFO] habushu::Examples::Consume a Local Poetry Package .. SUCCESS [ 13.245 s]
[INFO] habushu::Examples::Managed Dependencies ............ SUCCESS [ 11.800 s]
[INFO] habushu::Examples::Default Python Strategy ......... SUCCESS [ 11.755 s]
[INFO] habushu::Examples::Running Python Scripts .......... SUCCESS [ 14.946 s]
[INFO] habushu::Examples::Containerize Poetry Dependencies  SUCCESS [  9.487 s]
[INFO] habushu::Examples::Add Habushu to a new or existing uv project SUCCESS [  9.973 s]
[INFO] habushu::Examples::Disable exportRequirementsWithoutPathDependencies in uv SUCCESS [  9.772 s]
[INFO] habushu::Examples::Build an uv Package ............. SUCCESS [ 14.242 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  27.859 s (Wall Clock)
[INFO] Finished at: 2025-03-17T12:17:46-04:00
[...]

PR #295, opened 2025-03-17
[#294] fix usage of AbstractPythonPackageAndDependencyManagerSetup

woods-cpointe:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PythonPackageAndDependencyManagerSetup.java
```diff
@@ -0,0 +1,23 @@
+package org.technologybrewery.habushu;
+
+import java.util.List;
+
+import org.apache.maven.plugin.MojoExecutionException;
+
```
d-ryan-ashcraft, line 6: A (Administrative): There are all very easy to understand, but I generally think we should comment interfaces for completeness, especially as these will propagate down to several implementations.

d-ryan-ashcraft (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PythonPackageAndDependencyManagerSetup.java
woods-cpointe: Good call. Ill add comments

jaebchoi (approved):

Issue #283, opened 2025-03-12
aiSSEMBLE Lite/Habushu: Implement PublishToPyPiRepoMojo for UV/Poetry


jaebchoi:
## Description

As we support UV as additional package manager for habushu, we need to implement/refactor PublishToPyPiRepoMojo

## DOD
- [x] implement UV/Poetry switch logic for PublishToPyPiRepoMojo
- [x] implement UV specific logic for PublishToPyPiRepoMojo
- [x] habushu plugin for poetry can still function as before.
- [x] make sure UV PublishToPyPiRepoMojo passes and can be published.

Stretch Goal
- [ ] Implement [uv-monorepo-dependency-tool](https://github.com/TechnologyBrewery/uv-monorepo-dependency-tool/issues/2) into PublishToPyPiRepoMojo for uv if it is available.

## Testing Steps

**Prerequisite**
- If you don't have test-pypi account, create one [here](https://test.pypi.org/account/register/)
- Go to account Setting, and add API Token
![Image](https://github.com/user-attachments/assets/7bd3de6d-c85d-4705-bd48-f321b9229e50)
- Copy your API Token and add it to .m2/settings.xml server credentials like this 
NOTE: id should be `dev-pypi`, and username should be` __token__`
![Image](https://github.com/user-attachments/assets/a297c812-c0f8-4491-b454-3e9b5d02a4fa)

**Test publish to test pypi**
- Pull` dev` branch 
- On the root of habushu project, `run mvn clean install -Pbootstrap -Dmaven.build.cache.enabled=false` to build habushu maven plugin
- On the root of habushu project, `run mvn clean install -Puv -Dmaven.build.cache.enabled=false` to test UV install.
- On examples/habushu-uv-package/ run `mvn clean deploy ` Verify default-publish-to-pypi-repo MOJO is successful
NOTE: Build Failure in the (injected-nexus-deploy) is expected this is only intended for CI environment.

jaebchoi:
DoD Completed with @woods-cpointe

jaebchoi:
OTS Completed with @aaron-gary

meliz19:
Test passed!

PR #299, opened 2025-03-18
[#283] implement publishToPypiRepo for UV and Poetry

jaebchoi:


examples/add-habushu-to-new-or-existing-uv-project/pyproject.toml
```diff
@@ -40,3 +28,18 @@ dev = [
...
+name = "pypi"
+url = "https://pypi.org/simple/"
+publish-url = "https://pypi.org/simple/"
+
+
+# Added by habushu-maven-plugin at 2025-03-18T10:40:05.178201 to use https://test.pypi.org/simple/ as source PyPi repository for installing dependencies
+[[tool.uv.index]]
+name = "dev-pypi"
+url = "https://test.pypi.org/simple/"
+publish-url = "https://test.pypi.org/legacy/"
```
jaebchoi, line 44: I: This is needed when we want to publish to specific index. https://docs.astral.sh/uv/guides/package/#publishing-your-package

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractInstallDependencies.java
```diff
@@ -110,15 +112,21 @@ protected void prepareRepositoryForInstallation(String repoId, String repoUrl, S
...
                         String.format("name = \"%s\"",
                                 StringUtils.isNotEmpty(repoId) && !PUBLIC_PYPI_REPO_ID.equals(repoId)
                                         ? repoId
                                         : "private-pypi-repo"),
-                        String.format("url = \"%s\"", pypiRepoSimpleIndexUrl), "priority = \"supplemental\"");
+                        String.format("url = \"%s\"", pypiRepoSimpleIndexUrl)));
+                if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile())  == PackageManager.UV ) {
+                    newPypiRepoSourceConfig.add(String.format("publish-url = \"%s\"", getPublishUrl(repoUrl, !PUBLIC_PYPI_REPO_ID.equals(repoId))));
+                }
+                newPypiRepoSourceConfig.add("priority = \"supplemental\"");
```
jaebchoi, line 116: I: This is logic to add publish-url

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/InstallDependenciesMojo.java
```diff
@@ -95,6 +95,26 @@ public class InstallDependenciesMojo extends AbstractHabushuMojo {
...
+    @Parameter(property = "habushu.pypiUploadSuffix", defaultValue = "")
+    protected String pypiUploadSuffix = "";
+
+    /**
+     * {{@link #pypiUploadSuffix repositoryUploadSuffix} contains critical information.  The main difference is that
+     * this dev repository url path defaults to "legacy/" as the most common scenario when overriding the dev
+     * repository URL is to leverage test.pypi.org, which needs this configuration.
+     */
+    @Parameter(property = "habushu.devRepositoryUrlUploadSuffix", defaultValue = "legacy/")
+    protected String devRepositoryUrlUploadSuffix;
```
jaebchoi, line 116: I: needed those params to make sure we populate publish-url

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoUv.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        List<Pair<String, Boolean>> publishToRepoWithCredsArgs = Collections.emptyList();
+
+        if (StringUtils.isNotEmpty(username) && StringUtils.isNotEmpty(password)) {
+
+            publishToRepoWithCredsArgs = new ArrayList<>();
+
+            //Need to specify specific files as by default it would publish all tar.gz and whl files in dist folder.
+            //When publishing locally, uv will throw error if file name is same but hash is different which fails case of -dev0 https://github.com/astral-sh/uv/issues/7917#issuecomment-2495160227
```
jaebchoi, line 118: I: as I stated in the comment here, UV publishes all files in dist folder by default whereas poetry only publishes its project version artifacts. 
Yes UV does have functionality to skip upload file if there's file with same name but if the file will only skip if
it has same file name and have same exact content (i.e. same hash). Otherwise it will error out with this message `local and remote hash don't match` and and Fails publish.


in case of running publish locally or testing, we won't always have same content artifacts especially -dev0 artifacts in which fails publish MOJO.

Therefore, I have it so that  it only uploads artifacts with project version numbers.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractInstallDependencies.java
```diff
@@ -110,15 +112,21 @@ protected void prepareRepositoryForInstallation(String repoId, String repoUrl, S
...
                                 "# Added by habushu-maven-plugin at %s to use %s as source PyPi repository for installing dependencies",
                                 LocalDateTime.now(), pypiRepoSimpleIndexUrl),
                         String.format("[[%s]]", packageIndexPath),
                         String.format("name = \"%s\"",
                                 StringUtils.isNotEmpty(repoId) && !PUBLIC_PYPI_REPO_ID.equals(repoId)
                                         ? repoId
                                         : "private-pypi-repo"),
-                        String.format("url = \"%s\"", pypiRepoSimpleIndexUrl), "priority = \"supplemental\"");
+                        String.format("url = \"%s\"", pypiRepoSimpleIndexUrl)));
+                if (HabushuUtil.checkPythonPackageManager(getPyProjectTomlFile())  == PackageManager.UV ) {
```
aaron-gary: minor suggestion:  This line could be extracted to a method called bool isUV() or something like that to make it easier to read the code.  Just a suggestion.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractPublishToPyPiRepo.java
```diff
@@ -0,0 +1,134 @@
...
+        this.log = log;
+        this.publishToPyPiRepoMojo = mojo;
+    }
+
+
+    public abstract void doExecute() throws MojoExecutionException, MojoFailureException;
+
+
+
+    protected void invokePublish(AbstractCommandHelper commandHelper, List<Pair<String, Boolean>> publishToRepoWithCredsArgs) {
```
aaron-gary: Is there a concrete class you can use instead of AbstractCommandHelper?  Abstract classes shouldn't be used directly, but through classes that extend them.

jaebchoi: yeah I think this method should live in its own impl instead of [AbstractPublishToPyPiRepo.java](https://github.com/TechnologyBrewery/habushu/pull/299/files/f3bbd1b2ec98029f7aaa2e91fe5f9a55d51e4ef6#diff-f41f07cc51419e7f0708410e2021aded76ef0a6649b8c3f3500830112f995105) as I see some logic that is poetry only, good call!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/AbstractInstallDependencies.java
jaebchoi: Sounds good, Ill update to add those under HabushuUtil!

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoPoetry.java
```diff
@@ -0,0 +1,205 @@
...
+        super(baseDir, log, mojo);
+    }
+
+
+    @Override
+    public void doExecute() throws MojoExecutionException, MojoFailureException {
+        PoetryCommandHelper poetryHelper = new PoetryCommandHelper(baseDir);
+
+        String pomVersion = publishToPyPiRepoMojo.getProject().getVersion();
+        if (publishToPyPiRepoMojo.overridePackageVersion() && AbstractHabushuMojo.isPomVersionSnapshot(pomVersion)) {
```
aaron-gary: Q: Is there a non-abstract method we should use here?  Even if there is just a concrete class that's called DefaultHabushuMojo that is very basic, but extends AbstractHabushuMojo, that would be preferable

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoPoetry.java
```diff
@@ -0,0 +1,205 @@
...
+                    Arrays.asList("config", "--local", String.format("repositories.%s", repoId), repoUrl));
+        }
+
+        List<Pair<String, Boolean>> publishToRepoWithCredsArgs = Collections.emptyList();
+
+        if (StringUtils.isNotEmpty(username) && StringUtils.isNotEmpty(password)) {
+
+            publishToRepoWithCredsArgs = new ArrayList<>();
+
+            if (!AbstractHabushuMojo.PUBLIC_PYPI_REPO_ID.equals(repoId)) {
```
aaron-gary: Please replace AbstractHabushuMojo with a concrete class.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoUv.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+
+    @Override
+    public void doExecute() throws MojoExecutionException, MojoFailureException {
+
+        UvCommandHelper uvCommandHelper = new UvCommandHelper(baseDir);
+
+        String pomVersion = publishToPyPiRepoMojo.getProject().getVersion();
+        if (publishToPyPiRepoMojo.overridePackageVersion() && AbstractHabushuMojo.isPomVersionSnapshot(pomVersion)) {
```
aaron-gary: Please replace AbstractHabushuMojo with a concrete class

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/TestRetryPublishToPyPiRepoMojo.java
```diff
@@ -35,8 +33,8 @@ public TestRetryPublishToPyPiRepoMojo(int numberOfRetries, int finalRetryNumber)
 
     }
 
-    @Override
-    protected Callable<Boolean> getPyPiPushCallable(PoetryCommandHelper poetryHelper, List<Pair<String, Boolean>> publishToOfficialPypiRepoArgs) {
+
+    protected Callable<Boolean> getPyPiPushCallable(AbstractCommandHelper commandHelper, List<Pair<String, Boolean>> publishToOfficialPypiRepoArgs) {
```
aaron-gary: I'm a little less concerned with using an Abstract class since this is a test file, but at the same time, it's better to use a concrete class.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -269,4 +269,14 @@ public static File getShellConfigFile() {
...
     }
+
+
+    /**
+     * Finds current Python Package Manager and return whether it is UV.
+     *
+     * @return true if current Package Manager is UV.
+     */
+    public static boolean isCurrentPackageManagerUv(File pyProjectTomlFile) {
+        return HabushuUtil.checkPythonPackageManager(pyProjectTomlFile)  == PackageManager.UV;
```
woods-cpointe: a:
```suggestion
        return checkPythonPackageManager(pyProjectTomlFile)  == PackageManager.UV;
```

aaron-gary (approved):
Be sure to squash before merging (or select squash and merge in GH)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoPoetry.java
jaebchoi: I think we can move those helper method from AbstractHabushuMojo to HabushuUtil. I can spin up next ticket to refactor those for cleaner use.

woods-cpointe (approved):

Issue #303, opened 2025-03-20
Refactor containerize-dependencies Mojo to support UV as a package manager option


cpointe-ibllanos:
# Context

The containerize-dependencies mojo needs to be abstracted somewhat to support UV-based projects in addition to Poetry. 

# Definition of Done
* [x] Expand containerize-dependencies mojo to allow UV dockerfile support
* [x] Confirm UV-based dockerfile can be deployed and used
* [x] Create a new examples/readme to highlight differences between Poetry containerization
  * [x] Example builds as expected
  * [x] Readme updated for UV-specific parameters

# Steps to Test

## Build the dockerfile
1. Run `mvn clean install -Pbootstrap`
2. Run `mvn clean install -Puv`
3. Confirm the following file exists: `examples/habushu-containerize/habushu-uv-containerize/src/resources/Dockerfile`
4. Create a terminal at `examples/habushu-containerize/habushu-uv-containerize/`
5. Build the docker image successfully: `docker build -t myapp . -f ./src/resources/Dockerfile --no-cache`

## Confirm the deployment
1. After following the steps above run: `docker run --rm myapp ls /opt/venv/lib/python3.12/site-packages`
2. Confirm monorepo `add_habushu_to_new_or_existing_uv_project` is one of the packages present
3. run `docker run -it myapp`
4. In the python terminal that opens run:
  i. `from add_habushu_to_new_or_existing_uv_project import helloworld`
  ii. `helloworld.hello_world()`
5. Confirm the output: 'This is the example project that demonstrates how to add Habushu to a new or existing uv project.'

habibimoiz:
DoD LGTM!

chang-annie:
OTS completed with @aaron-gary

chang-annie:
all test steps are passing, closing ticket.

Issue #207, opened 2025-02-07
Refactor `containerize-dependencies` Mojo to use Apache Velocity for creating dockerfiles


chang-annie:
# Context
The containerize-dependencies mojo needs to be abstracted somewhat to support UV-based projects in addition to Poetry. Additionally, the programmatic nature of our dockerfile generation process is getting messy and would probably be more suitable to a templating solution such as Velocity.

# Definition of Done
* ❌ Expand `containerize-dependencies` mojo to allow UV support (follow-on)
* ✅ Refactor the `containerize-dependencies` Mojo to use Apache Velocity Engine

# Steps to Test

## Velocity refactor
1. run `mvnd clean install -Pbootstrap` and confirm success
2. run `mvnd clean install` and confirm success
3. run `mvnd clean install -pl :habushu-poetry-containerize -Dhabushu.dockerPoetryVersion=TEST`
4. open the file: `examples/habushu-poetry-containerize/src/resources/Dockerfile`
5. Confirm line 5 points to poetry version "poetry==TEST"

cpointe-ibllanos:
Adding a note based on standup convo: we should look and see if our approach to creating Dockerfiles and containerizing is compatible / optimal for our use case. The first thing to do would be to see if this is even compatible with a `uv` project setup before attempting to refactor the existing implementation to support multiple package managers.

EDIT: This has been handled in #288 . This issue will now again focus on refactoring the containerize dependencies mojo to support UV and use Velocity templating in the dockerfile generation.

jaebchoi:
Completed DoD with @jaebchoi @cpointe-ibllanos

cpointe-ibllanos:
DoD performed with @jaebchoi

cpointe-ibllanos:
In the interest of atomizing issues, I'm moving the UV portion of this to #303

cpointe-ibllanos:
OTS completed with @colinpalmer-pro

chang-annie:
all test steps passing, closing issue.

PR #304, opened 2025-03-21
#207 Refactor containerize-dependencies Mojo to use Apache Velocity for creating dockerfiles

cpointe-ibllanos:
#207 
_Refactor containerize-dependencies Mojo to use Apache Velocity for creating dockerfiles_

Contains some small test refactoring in prep for #303 
_Refactor containerize-dependencies Mojo to support UV as a package manager option_


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -34,6 +32,7 @@
 public class ContainerizeDepsMojo extends AbstractHabushuMojo {
 
     private static final Logger logger = LoggerFactory.getLogger(ContainerizeDepsMojo.class);
+    private static final String DOCKERFILE_POETRY_VM = "templates/dockerfile.vm";
```
d-ryan-ashcraft: S: I'd recommend making this a maven-enabled parameter with this as a default.  This would allow it to be overridden if appropriate.

d-ryan-ashcraft (approved):

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/TestUtils.java
```diff
@@ -0,0 +1,6 @@
+package org.technologybrewery.habushu;
```
meliz19: Q/S: Instead of creating this new class, could we call `PackageManager` in its place?

meliz19 (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
cpointe-ibllanos: `templates/dockerfile.vm` lives in the classpath so I'm not sure what use this would have to a downstream developer. 

Would we want to give the dev the option of using their own template / changing the existing template? If so, I can expand the velocity engine setup to look in the downstream's file system for a custom dockerfile template file first before deferring to the default in the habushu jar, then make additional documentation for how to do this in the README. Although I imagine someone with extensive customizations would instead disable the dockerfile generation portion before going willy-nilly on it with edits.

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/TestUtils.java
cpointe-ibllanos: Good catch! Yeah there is a way we can convert the existing enumerator to strings.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
epwilkins: I don't think we need to default to looking on the file system first. The idea is to support extension I think. In which case it'd be pretty simple to get the custom template on Habushu's classpath, or change the Velocity configuration to add a new resource loader for template resolution.

Issue #313, opened 2025-03-28
Remove unused repo storage credential methods


meliz19:
## Description
In the process of #312, I realized that we don't use the methods listed in DoD. The purpose of this ticket is to remove those methods from the codebase. 

## DoD
- Remove the following methods:
  - [x] `registerRepositoryToSupportAuthenticatedDependencyResolution`
  - [x] `configurePrivateDevPyPiRepositoryCredentials`
  - [x] `configurePrivatePyPiRepositoryCredentials`
  - [x] `logSkipRationale`

## Test Steps
**Prerequisite**
- If you don't have test-pypi account, create one [here](https://test.pypi.org/account/register/)
- Go to account Setting, and add API Token
![Image](https://github.com/user-attachments/assets/7bd3de6d-c85d-4705-bd48-f321b9229e50)
- Copy your API Token and add it to .m2/settings.xml server credentials like this 
NOTE: id should be `dev-pypi`, and username should be` __token__`
![Image](https://github.com/user-attachments/assets/a297c812-c0f8-4491-b454-3e9b5d02a4fa)

- Get clean builds for
  - `mvn clean install -Pbootstrap`
  - `mvn clean install`
- Run the following commands. In both cases, the command will ultimately fail because you're not authorized to publish these packages. See the note below for the expected logs. 
  - `cd examples/habushu-poetry-package && mvn clean deploy`
  - `cd examples/habushu-uv-package && mvn clean deploy`

Expected Logs
```
[INFO] --- habushu:3.0.0-SNAPSHOT:publish-to-pypi-repo (default-publish-to-pypi-repo) @ habushu-poetry-package ---
[INFO] Executing poetry command: poetry version -s
[INFO] Setting version of Poetry package to publish to 3.0.0.dev1743158606
[INFO] Executing poetry command: poetry version 3.0.0.dev1743158606
[INFO] Bumping version from 3.0.0.dev to 3.0.0.dev1743158606
[INFO] Publishing to dev repository (useDevRepository=true, dev version being published)
[INFO] Adding repository configuration to poetry.toml for dev-pypi at https://test.pypi.org/legacy/
[INFO] Executing poetry command: poetry config --local repositories.dev-pypi https://test.pypi.org/legacy/
[INFO] Publishing archives to https://test.pypi.org/legacy/ 
[INFO] Executing poetry command: poetry publish --build --repository dev-pypi --username __token__ --password XXXX
[...]

PR #314, opened 2025-03-28
[#313] Removed unused repo storage credential methods

meliz19:
#313

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
@@ -152,29 +152,6 @@ public class PyenvAndPoetrySetup extends AbstractPythonPackageAndDependencyManag
     }
 
     /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void registerRepositoryToSupportAuthenticatedDependencyResolution(String repoId, String username, String password) {
-        PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
-
-        if (StringUtils.isEmpty(username) || StringUtils.isEmpty(password)) {
-            log.info(String.format("Did not find username and password for the server with <id> %s. Will use existing configuration.", repoId));
-        } else {
-            String configKey = String.format("http-basic.%s", repoId);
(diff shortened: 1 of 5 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -151,29 +151,6 @@ public String pythonSourceMessage() {
...
-            String configKey = String.format("http-basic.%s", repoId);
-            log.info(String.format("Adding username and password configuration for %s", repoId));
-
-            List<Pair<String, Boolean>> credentialConfigurationArgs = new ArrayList<>();
-            credentialConfigurationArgs.add(new ImmutablePair<>("config", false));
-            credentialConfigurationArgs.add(new ImmutablePair<>(configKey, false));
-            credentialConfigurationArgs.add(new ImmutablePair<>(username, false));
-            credentialConfigurationArgs.add(new ImmutablePair<>(password, true));
-
-            poetryHelper.executeWithSensitiveArgsAndLogOutput(credentialConfigurationArgs);
```
meliz19, line 173: I: We don't use this method of accessing the repo credentials within habushu. So, I'm deleting this method and its related methods across habushu to help with maintainability.

d-ryan-ashcraft: -How is it happening in that case?  I would have thought this would be used?- I misread, my question is OBE

epwilkins: I think a key element missing in the issue test steps is pulling a dependency. Looks like we're verifying that the push does not use the credentials being configured here, but this is explicitly for sourcing dependencies from a repository that requires authentication.

d-ryan-ashcraft (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
meliz19: Good catch. thank you @ewilkins-csi!

meliz19:
Closing as OBE

PR #316, opened 2025-04-01
[#303] UV containerize depencies mojo support

cpointe-ibllanos:
#303


examples/habushu-uv-containerize/README.md
```diff
@@ -0,0 +1,64 @@
+### Leveraging the containerize-dependencies Goal to Prepare a Containerized Virtual Environment ###
```
epwilkins: S: I don't think we should duplicate all of this information. We could make `examples/habushu-containerize/habushu-uv-containerize` and `examples/habushu-containerize/habushu-poetry-containerize` with a common README. That can then link to the two specific READMEs for info that differs.

examples/habushu-uv-containerize/README.md
```diff
@@ -0,0 +1,64 @@
...
+copy the project files for each dependency to a staging directory, while preserving the original structure of the
+dependency modules to ensure that any path-based dependencies can be leveraged as-is. This directory can then be copied
+onto a Docker container and used to create a virtual environment capable of running the target Habushu project.
+
+The plugin will automatically inject logic for building and using this virtual environment into a pre-existing
+Dockerfile by default. The `dockerfile` configuration must be set to the target Dockerfile. To disable the Dockerfile
+update altogether, set the `updateDockerfile` configuration to `false`. Because the virtual environment that is created
+is dependent on the platform for which it is built, Habushu defaults to using `python:3.12` to build the virtual
+environment and  `python:3.12-slim` as the final image that packages/runs the virtual environment. This can be
+customized with the `dockerBuilderBase`, `dockerFinalBase`, `dockerUser`, `dockerUvVersion`, and `dockerfileTemplateUv` configurations, but care must be taken to
```
epwilkins: A: Looks like we've got a 120 wrap in this readme.

examples/habushu-uv-containerize/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.technologybrewery.habushu</groupId>
+                <artifactId>habushu-maven-plugin</artifactId>
+                <configuration>
+                    <managedDependencies>
```
epwilkins: Q: Is this configuration relevant to the example? If not seems like we should remove it to cut down on noise.

examples/habushu-uv-containerize/src/resources/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+USER root
+WORKDIR /work-dir/containerize-support/examples/add-habushu-to-new-or-existing-uv-project
+
+ENV UV_COMPILE_BYTECODE=1
+ENV UV_LINK_MODE=copy
+ENV UV_CACHE_DIR="/.cache/uv"
+
+# Export target project's virtual environment to /opt/venv
+RUN --mount=type=cache,target=/.cache/uv/ \
```
epwilkins: A: You should be able to use `target=${UV_CACHE_DIR}`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -214,10 +235,15 @@ protected FileSet getSourceSet() {
...
         fileSet.addInclude(relativeSrc +"/**");
         fileSet.addInclude("pyproject.toml");
-        fileSet.addInclude("poetry.toml");
-        fileSet.addInclude("poetry.lock");
         fileSet.addInclude("README.md");
 
+        if (PackageManager.POETRY.equals(manager)) {
+            fileSet.addInclude("poetry.toml");
+            fileSet.addInclude("poetry.lock");
+        } else { // UV
```
epwilkins: A: Might consider explicitly checking that its `uv` and throwing an exception in an else block if it's neither.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -285,17 +311,27 @@ protected void setUpdateDockerfile(boolean update) {
...
+        String template;
+
+        context.setAnchorDirectory(outputDir.toString());
+        context.setSingleRepoProjectDir(targetProjectPath.toString());
+        context.setOwner(dockerUser);
+        context.setBuilderBaseImage(dockerBuilderBase);
+        context.setFinalBaseImage(dockerFinalBase);
+
+        if (PackageManager.POETRY.equals(manager)) {
+            context.setPoetryMonorepoDependencyPluginVersion(dockerPoetryMonorepoDependencyPluginVersion);
```
epwilkins: S: Consider polymorphism so consumers of the class don't have to guess what conditions cause certain properties to be valid/invalid.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileContext.java
```diff
@@ -0,0 +1,58 @@
+package org.technologybrewery.habushu.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.velocity.VelocityContext;
+
+public class ContainerizeDepsDockerfileContext extends VelocityContext {
```
epwilkins: A: Docker Context has a very specific meaning, and this name doesn't indicate it's for use with Velocity. 
```suggestion
public class ContainerizeDepsVelocityContext extends VelocityContext {
```

epwilkins (approved):

Issue #323, opened 2025-04-04
Habushu: Fix inability to set devRepositoryUrlUploadSuffix to empty value


jaebchoi:
### Description
Currently we are unable to set devRepositoryUrlUploadSuffix to empty value due to way maven property is structured.
If we set devRepositoryUrlUploadSuffix to empty or null, it will always go to default value which is /legacy.

We should create additional property called enableDevRepositoryUrlUploadSuffix to allow null value.

if enableDevRepositoryUrlUploadSuffix is set to false, upload suffix will always be empty regardless of devRepositoryUrlUploadSuffix value.

if enableDevRepositoryUrlUploadSuffix is set to true, upload suffix will be devRepositoryUrlUploadSuffix value, and setting devRepositoryUrlUploadSuffix empty or null value will set it to default value.

### DoD

- [x] update to add enableDevRepositoryUrlUploadSuffix whereever devRepositoryUrlUploadSuffix exist and add logic to allow/omit devRepositoryUrlUploadSuffix based on enableDevRepositoryUrlUploadSuffix flag.
- [x] Add README for enableDevRepositoryUrlUploadSuffix

### Testing Instruction
- Pull dev of Habushu
- In the `habushu/examples/habushu-uv-devrepository-package/pom.xml`, Verify `enableDevRepositoryUrlUploadSuffix` is set to true.
- In the `habushu/examples/habushu-uv-devrepository-package/pyproject.toml,` Delete whole `[[tool.uv.index]]` sections with `name = "dev-pypi"`
- on the root of project, run `mvn clean install -Puv -Dmaven.build.cache.enabled=false`
- In the `habushu/examples/habushu-uv-devrepository-package/pyproject.toml`, verify `[[tool.uv.index]]` is regenerated and `publish-url` is `"https://test.pypi.org/legacy/"`
- In the `habushu/examples/habushu-uv-devrepository-package/pom.xml`, Set `enableDevRepositoryUrlUploadSuffix` to false.
- In the `habushu/examples/habushu-uv-devrepository-package/pyproject.toml`, Delete whole `[[tool.uv.index]] `sections with `name = "dev-pypi"`
- on the root of project, run `mvn clean install -Puv -Dmaven.build.cache.enabled=false`
- In the `habushu/examples/habushu-uv-devrepository-package/pyproject.toml`, verify `[[tool.uv.index]]` is regenerated and `publish-url` is `"https://test.pypi.org/"`

jaebchoi:
[...]

PR #325, opened 2025-04-04
[#323] Fix inability to set devRepositoryUrlUploadSuffix to empty value

jaebchoi:
https://github.com/TechnologyBrewery/habushu/issues/323

### DIFF
--- a/examples/habushu-uv-devrepository-package/.gitignore
+++ b/examples/habushu-uv-devrepository-package/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in generated package with the exception of __init__.py - protobuf bindings are generated
+# into this package as a part of the automated build and should *not* be version-controlled
+src/habushu_uv_package/generated/__init__.py
+!src/habushu_uv_package/generated/__init__.py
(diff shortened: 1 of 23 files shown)


examples/habushu-uv-package/pom.xml
```diff
@@ -40,6 +40,7 @@
                                 <active>true</active>
                             </packageDefinition>
                         </managedDependencies>
+                        <enableDevRepositoryUrlUploadSuffix>true</enableDevRepositoryUrlUploadSuffix>
```
meliz19: S: I suggest we remove the config from this example project and instead create a stand alone example for the `DevRepository`-related configs. If you want to create that example in a separate ticket to minimize the PR size, I think that's okay IMO.

meliz19 (approved):

examples/habushu-uv-package/pom.xml
jaebchoi: Sounds good, Added one example project for devRepository!

examples/habushu-uv-devrepository-package/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in generated package with the exception of __init__.py - protobuf bindings are generated
```
jaebchoi, line 1: I: habushu-uv-devrepository-package files are simple example project intended to test enableDevRepositoryUrlUploadSuffix, can only review pom file for this directory.

Issue #342, opened 2025-04-11
Integrate uv monorepo tool into Habushu


meliz19:
## Description
The purpose of this ticket is to enable the `rewriteLocalPathDepsInArchives` functionality for uv-based projects.

## DoD
- [x] Integrate the [uv-monorepo-dependency-tool](https://github.com/TechnologyBrewery/uv-monorepo-dependency-tool) into BuildDeploymentArtifactsUv
- [x] Integrate the [uv-monorepo-dependency-tool](https://github.com/TechnologyBrewery/uv-monorepo-dependency-tool) into UvPublishToPyPiRepoUv
- [x] Update the README docs 
- [x] Add an example project 

## Test Steps 
- Build the habushu plugin: `mvn clean install -Pbootstrap`
- Prerequisite: Clone and build the latest dev version of the [uv-monorepo-dependeny-tool](https://github.com/TechnologyBrewery/uv-monorepo-dependency-tool)

**All of the remaining steps will be in the habushu repo.** 
- Verify all the example projects build successfully: `mvn clean install`
- Open `examples/uv/habushu-uv-package-consumer/pom.xml` and change the `rewriteLocalPathDepsInArchives` to `true`. 
- Rebuild the module: `mvn clean install -pl :habushu-uv-package-consumer`
- Check that the `uv-monorepo-dependency-tool` was installed. Search the logs for the following line:
```bash
[INFO] Executing uv command: uv tool install uv-monorepo-dependency-tool@latest
```
- Check that the `habushu-uv-package-consumer` module DID rewrite local path dependencies in the wheel file: 
  - Unzip `examples/uv/habushu-uv-package-consumer/dist/habushu_uv_package_consumer-3.0.0.dev0-py3-none-any.whl`
  - `cat examples/uv/habushu-uv-package-consumer/dist/habushu_uv_package_consumer-3.0.0.dev0-py3-none-any/habushu_uv_package_consumer-3.0.0.dev0.dist-info/METADATA | grep "Requires-Dist: habushu-uv-package"`. The result should be
```bash
Requires-Dist: habushu-uv-package==3.0.0.dev
```
- Verify that we can disable the `rewriteLocalPathDepsInArchives`. To do this, open the `examples/uv/habushu-uv-package-consumer/pom.xml` and update `rewriteLocalPathDepsInArchives` to false, like below.  
```xml
<rewriteLocalPathDepsInArchives>false</rewriteLocalPathDepsInArchives>
```
[...]

PR #345, opened 2025-04-14
[#342] integrate uv-monorepo-dep-tool and added example

meliz19:
#342

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ In order to use Habushu, the following prerequisites must be installed:
 
 * Maven 3.9+ or mvnd 1.0.2+
 * Java 11+
+
+If you would like to use Habushu with Poetry-based projects, you must install:
 * [Poetry 1.5+](https://python-poetry.org/)
 * [Pyenv](https://github.com/pyenv/pyenv)
 
@@ -30,6 +32,7 @@ If you would like to use Habushu with uv-based projects, you must install:
 Additionally, Habushu may install and manage:
 
 * [poetry-monorepo-dependency-plugin](https://pypi.org/project/poetry-monorepo-dependency-plugin/) for Poetry-based projects
+* [uv-monorepo-dependency-tool](https://pypi.org/project/uv-monorepo-dependency-tool/) for uv-based projects
 
 ## Usage ##
 
@@ -234,7 +237,7 @@ Default: `true`
 
 When `pythonVersion` is not explicitly specified, this strategy helps decide where to get the default version. If set to 
 `PYTHONVERSION`, then Habushu will use the existing project's Python version. If set to `POM`, then Habushu will use 
-the default value for `pythonVerion`
+the default value for `pythonVersion`
 
 Default: `PYTHONVERSION`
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -30,6 +32,7 @@ If you would like to use Habushu with uv-based projects, you must install:
 Additionally, Habushu may install and manage:
 
 * [poetry-monorepo-dependency-plugin](https://pypi.org/project/poetry-monorepo-dependency-plugin/) for Poetry-based projects
+* [uv-monorepo-dependency-tool](https://pypi.org/project/uv-monorepo-dependency-tool/) for uv-based projects
```
meliz19, line 35: I: Adding this now since we're going to release the uv-monorepo-dependency-tool within the next few days.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/UvSetup.java
```diff
@@ -90,7 +89,9 @@ public List<String> validatePackageAndDependencyManagerInstallationAndVersion(
      */
     @Override
     public void finalizePythonPackageAndDependencyManagerConfiguration() {
-        // Currently no monorepo dependency plugin to install
+        if (rewriteLocalPathDepsInArchives) {
```
meliz19, line 92: I: Poetry installs the plugin whether it's used or not. I broke that pattern here and only install the tool if we will use it. Any objections?

README.md
```diff
@@ -653,6 +650,10 @@ Default: `${project.basedir}/target`
 
 Controls whether locally pathed dependencies should be excluded from the requirements.txt export file.
 
+For Poetry projects, Habushu uses the [poetry-monorepo-dependency-plugin]()'s `export-without-path-deps` commands in place of Poetry's `export` command.
+
+[//]: # (TODO: For uv projects, Habushu uses the [uv-monorepo-dependency-tool]&#40;&#41;'s `export-without-path-deps` commands in place of uv's `export` command.)
```
meliz19, line 655: Will uncomment once this command is added to the `uv-monorepo-dependency-tool`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsUv.java
```diff
@@ -32,14 +32,22 @@ public void doExecute() {
...
+            uvCommandHelper.executeAndLogOutput(uvCommandHelper.createToolRunCommand(buildCommand));
+        } else {
+            String buildCommand;
+            buildCommand = "build";
+            buildLogMessage = "Building source and wheel archives...";
+            uvCommandHelper.executeAndLogOutput(Arrays.asList(buildCommand));
+        }
 
         log.info(buildLogMessage);
-        uvCommandHelper.executeAndLogOutput(Arrays.asList(buildCommand));
```
epwilkins, line 42: A: It's a bit odd to have the log statement happen after the command. Could we assign the list to a `command` variable or move the `log.info` into the if/else block?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsUv.java
```diff
@@ -32,14 +32,22 @@ public void doExecute() {
...
-        buildLogMessage = "Building source and wheel archives...";
+        if (buildDeploymentArtifactsMojo.rewriteLocalPathDepsInArchives()) {
+            List<String> buildCommand = new ArrayList<>();
+            buildCommand.add("uv-monorepo-dependency-tool");
+            buildCommand.add("build-rewrite-path-deps");
+            buildLogMessage = "Building source and wheel archives with uv-monorepo-dependency-tool...";
+            uvCommandHelper.executeAndLogOutput(uvCommandHelper.createToolRunCommand(buildCommand));
+        } else {
+            String buildCommand;
+            buildCommand = "build";
```
epwilkins: A: I don't know that we're getting any use out of this being a variable. I'd just inline it.  I think it was probably a variable so the if/else logic could assign it based on whether rewrite was active. But we just haven't structured the code that way so there's no need for it.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoUv.java
```diff
@@ -138,9 +137,21 @@ protected void publishPackage(UvCommandHelper uvCommandHelper, boolean rebuildPa
...
+                    List<String> buildCommand = new ArrayList<>();
+                    buildCommand.add("uv-monorepo-dependency-tool");
+                    buildCommand.add("build-rewrite-path-deps");
+                    buildLogMessage = "Building source and wheel archives with uv-monorepo-dependency-tool...";
+                    uvCommandHelper.executeAndLogOutput(uvCommandHelper.createToolRunCommand(buildCommand));
+                } else {
+                    String buildCommand;
+                    buildCommand = "build";
+                    buildLogMessage = "Building source and wheel archives...";
+                    uvCommandHelper.executeAndLogOutput(Arrays.asList(buildCommand));
```
epwilkins: A: Same comments here about log ordering and `buildCommand` variable.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BuildDeploymentArtifactsUv.java
meliz19: That's an excellent point. I thought I was being efficient, but I didn't use my brain to think that through all the way. 🤦‍♀️

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PublishToPyPiRepoUv.java
epwilkins: Also just for info as it really doesn't impact this code, `List.of` is a bit more performant for lists <= 2. But the efficiency gain is minuscule and would only matter if you were creating millions of tiny lists.

epwilkins (approved):

Issue #346, opened 2025-04-15
Deprecate exportRequirementsWithoutPathDependencies


meliz19:
## Description
When habushu was first developed, we created the `exportRequirementsWithoutPathDependencies` configuration to generate `requirements.txt` files without relative paths/editable dependencies for use in containerized environments. We have since transitioned to use the `containerize-dependencies` goal instead. The `exportRequirementsWithoutPathDependencies` isn't the best solution in practice, so we are deprecating this configuration in this ticket. Users who want to use packages with editable dependencies in a containerized environment should use the `containerize-dependencies` goal. 

## DoD
- [x] Remove all mentions of the `exportRequirementsWithoutPathDependencies` from habushu

## Test Steps
- Build habushu: `mvn clean install -Pbootstrap`
- Build the example projects: `mvn clean install`
- Search the habushu repo for any mentions of the `exportRequirementsWithoutPathDependencies` configuration. Verify that the only reference is in the README, which notes that the configuration is deprecated.

meliz19:
Completed DoD/OTS with @aaron-gary

chang-annie:
all test steps are passing! closing ticket

PR #347, opened 2025-04-15
[#346] Deprecate exportRequirementsWithoutPathDependencies

meliz19:
#346


epwilkins (approved):

epwilkins:
Have we been keeping up with the draft release notes? Seems like it'd be good to note this as a breaking change.

meliz19:
> Have we been keeping up with the draft release notes? Seems like it'd be good to note this as a breaking change.

I was thinking we'd just use the auto-generated release notes, but I'm happy to put together a manual version if you think that's better. Just let me know!

Issue #358, opened 2025-04-23
Create poetry example for withGroups and withoutGroups


meliz19:
## Description 
In doing #348, we identified several configs that could use example projects to illustrate how the configs work. The purpose of this ticket is to develop an example for `withGroups` and `withoutGroups` in Poetry.

## DoD
- [x] Develop a Poetry project that uses the `withGroups` and `withoutGroups` habushu configurations
- [x] Add/Update README documentation where relevant

## Test Steps
1. Build the habushu-maven-plugin
```bash
mvnd clean install -Pbootstrap
```

2. Build the example project without the build cache so that the `pom.xml` updates are captured.
```bash
mvnd clean install -pl :habushu-poetry-dependency-groups -Dmaven.build.cache.skipCache
```

3. In the `examples/poetry/habushu-poetry-dependency-groups/pom.xml` we specify that `groupA` is to be included in the virtual environment sync via the `withGroups` configuration and `groupB` is NOT to be included in the virtual environment sync via the `withoutGroups` configuration . To verify that the `groupA` dependency group was included and the `groupB` dependency group was not included, open  the `examples/poetry/habushu-poetry-dependency-groups/.venv/lib/python3.11/site-packages` and 
    - Verify that the `groupA` dependency (`numpy`) is included 
    - Verify that the `groupB` dependency (`six`) is NOT included

4. Open `examples/poetry/habushu-poetry-dependency-groups/pom.xml` and comment out the `withGroups` and `withoutGroups` configurations.
5. Re-run the build command in step 2. 
6. In the `examples/poetry/habushu-poetry-dependency-groups/.venv/lib/python3.11/site-packages`,
    - Verify that the `groupA` dependency (`numpy`) is NOT included (Note: the groupA dependency group is marked as optional in the `pyproject.toml` and therefore will not be installed by default).
    - Verify that the `groupB` dependency (`six`) is included (Note: the groupB dependency group is NOT marked as optional in the `pyproject.toml` and therefore will be installed by default).

meliz19:
Completed DoD/OTS with @woods-cpointe & @lemuelwatts

chang-annie:
all test steps are passing! closing issue.

PR #359, opened 2025-04-23
[#358] Add Poetry example for the withGroups and withoutGroups Habush…

meliz19:
…u configs

#358


examples/poetry/habushu-poetry-dependency-groups/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+    <packaging>habushu</packaging>
+
+    <build>
+        <directory>dist</directory>
+        <plugins>
+            <plugin>
+                <groupId>org.technologybrewery.habushu</groupId>
+                <artifactId>habushu-maven-plugin</artifactId>
+                <configuration>
+                    <forceSync>true</forceSync>
```
meliz19, line 24: I: I'm not calling this usage out in the examples docs bc in a follow-on ticket, we will set the `forceSync` default to true. The reason for this is that Poetry recommends to use `poetry sync` instead of `poetry install` in their latest documentation.

d-ryan-ashcraft (approved):

chang-annie (approved):

Issue #354, opened 2025-04-18
Poetry v2 Migration to remove empty sections from pyproject.toml


chang-annie:
After the Poetry v2 Migrations are run, some of the sections in the resulting `pyproject.toml` file are empty. The migration should account for these empty headers and remove them.

Example - `examples/poetry/habushu-poetry-add-to-project/pyproject.toml`
```
[tool.poetry]
packages = [{include = "habushu_poetry_add_to_project", from = "src"}]

[project]
name = "habushu-poetry-add-to-project"
version = "3.0.0.dev"
description = "This example demonstrates how to add Habushu to a new or existing Poetry project."
authors = [{name = "Margaret Black", email = "black_margaret@bah.com"}]
dynamic = ["version", "dependencies"]
readme = "README.md"
requires-python = ">=3.12,<4"

[tool.poetry.dependencies]

[tool.poetry.group.dev.dependencies]
behave = "^1.2.6"
kappa-maki = "^1.0.2"
ruff = ">=0.9.9"
...
```

### Definition of Done
- [x] Update Poetry v2 migrations to check and remove empty headers
- [x] Run Poetry v2 migration against `examples` directory
- [x] Create unit tests
- [x] Update Poetry v2 changes section of `README`, if necessary

### Test Steps
1. Pull the latest `dev` branch and build using the bootstrap profile - `mvn clean install -Pbootstrap` (you must be on Poetry v2.0+ to build Habushu)
2. Build the example projects - `mvn clean install`
3. Change directory to `examples/poetry/habushu-poetry-package`. We will use the `pyproject.toml` in this directory as our test project
4. Add just the `[tool.poetry]` header into the file and run `mvn habushu:baton-migrate` to trigger the migration
> **Before**
> ```
> [tool.poetry]
> 
> [project]
> name = "habushu_poetry_package"
> version = "3.0.0.dev"
> ...
> ```

5. Confirm the empty `[tool.poetry]` header was removed by the migration
> **After**
> ```
> [project]
> name = "habushu_poetry_package"
> version = "3.0.0.dev"
> ...
> ```

6. Remove everything under the `[tool.poetry.dependencies]` header and run `mvn habushu:baton-migrate`
> **Before**
> ```
> [project]
> name = "habushu_poetry_package"
> version = "3.0.0.dev"
> ...
> 
> [tool.poetry.dependencies]
> 
> [tool.poetry.group.dev.dependencies]
> behave = ">=1.2.6"
> ...
> ```

[...]

PR #360, opened 2025-04-24
[#354] Poetry v2 Migration - remove empty section headers

chang-annie:
#354 

- `PoetryRemoveEmptyTomlMigration` to remove empty `[tool.poetry]` and `[tool.poetry.dependencies]` headers from `pyproject.toml` if they exist
- update documentation to include info about new migration
- wrote unit tests and refactored previous Poetry v2 Migration tests to share context for checking if Poetry version at least 2.0.0
- ran migration against all example Poetry `pyproject.toml` files

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryRemoveEmptyTomlMigration.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryRemoveEmptyTomlMigration.java
@@ -1,0 +1,115 @@
+package org.technologybrewery.habushu.migration.poetryv2migrations;
+
+import com.electronwill.nightconfig.core.Config;
+import com.electronwill.nightconfig.core.file.FileConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.baton.BatonException;
+import org.technologybrewery.habushu.util.TomlUtils;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Provides functionality to remove empty headers from pyproject.toml file
+ * Runs when on Poetry v2.0.0 or later.
+ */
+public class PoetryRemoveEmptyTomlMigration extends AbstractPoetryMigration{
+    private static final Logger logger = LoggerFactory.getLogger(PoetryRemoveEmptyTomlMigration.class);
+    private boolean isToolPoetryEmpty;
+    private boolean isToolPoetryDepsEmpty;
+
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        if (isPoetryProject(file) && isPoetryVersionAtLeast2){
+            try(FileConfig tomlFileConfig = FileConfig.of(file)){
+                tomlFileConfig.load();
+                List<String> allTomlSectionHeaders = TomlUtils.extractTomlSectionHeaders(file);
+
+                Optional<Config> toolPoetryGroupOpt = tomlFileConfig.getOptional(TomlUtils.TOOL_POETRY);
+                boolean hasToolPoetryHeader = allTomlSectionHeaders.contains(TomlUtils.TOOL_POETRY);
+
+                // only check for emptiness if the header actually exists
+                if (hasToolPoetryHeader && toolPoetryGroupOpt.isPresent()){
+                    isToolPoetryEmpty = TomlUtils.isSectionEmpty(toolPoetryGroupOpt.get(), TomlUtils.TOOL_POETRY, allTomlSectionHeaders);
+                }
+
+                Optional<Config> toolPoetryDepsGroupOpt = tomlFileConfig.getOptional(TomlUtils.TOOL_POETRY_DEPENDENCIES);
+                boolean hasToolPoetryDepsHeader = allTomlSectionHeaders.contains(TomlUtils.TOOL_POETRY_DEPENDENCIES);
+
+                if (hasToolPoetryDepsHeader && toolPoetryDepsGroupOpt.isPresent()) {
+                    isToolPoetryDepsEmpty = TomlUtils.isSectionEmpty(toolPoetryDepsGroupOpt.get(), TomlUtils.TOOL_POETRY_DEPENDENCIES, allTomlSectionHeaders);
+                }
+            }
+        }
+        return isToolPoetryEmpty || isToolPoetryDepsEmpty;
+    }
+
+
+
+    @Override
+    protected boolean performMigration(File pyProjectTomlFile) {
+        boolean inToolPoetrySection = false;
+        boolean inToolPoetryDepsSection = false;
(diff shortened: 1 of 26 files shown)


habushu-maven-plugin/src/main/resources/migrations.json
```diff
@@ -79,6 +79,20 @@
...
+      {
+        "name": "poetry-v2-remove-empty-toml-migration",
+        "implementation": "org.technologybrewery.habushu.migration.poetryv2migrations.PoetryRemoveEmptyTomlMigration",
+        "fileSets": [
+          {
+            "includes": [
+              "pyproject.toml"
+            ],
+            "excludes": [
+              "**/examples/poetry/habushu-poetry-install-from-dev-repo/pyproject.toml"
```
chang-annie, line 92: I: excluded this example toml file specifically because it contains a `[tool.poetry.dependencies]` header with only commented out code 
```
[tool.poetry.dependencies]
#    UNCOMMENT TO ENABLE FUNCTIONAL EXAMPLE
#habushu-poetry-publish-to-dev-repo = {source = "private-repo"}
```
we can either keep as excluded or add an arbitrary dependency like `krausening = ">=20"` to keep the migration from deleting the header completely

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryV2MigrationContext.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.File;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Shared context for BDD steps to verify whether the Poetry version
+ * is at least 2.0.0 or not.
+ */
+public class PoetryV2MigrationContext {
```
chang-annie, line 16: I: created a shared context for the Poetry v2 Migration tests so that we can reuse the `Given` steps to check if the poetry version is at least 2. previously, we had different variations of the `Given` step to avoid having duplicate step errors

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/TomlUtils.java
```diff
@@ -376,4 +380,82 @@ public static boolean hasComparators(String string) {
...
     }
 
+    /**
+     * Reads a TOML file and extracts every section header (the text between '[' and ']').
+     *
+     * @param file  the TOML file to scan
+     * @return      a list of section names without surrounding brackets
+     */
+    public static List<String> extractTomlSectionHeaders(File file){
+        Pattern headerPattern = Pattern.compile("^\\[+(.+?)\\]+$");
```
chang-annie: I: the regex looks in the toml file for lines that look like `[x.y.z]` or `[[x.y.z]]` and will return `x.y.z` as the capture group

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/TomlUtils.java
```diff
@@ -376,4 +380,82 @@ public static boolean hasComparators(String string) {
...
+     * @param parentHeaderName  the full section name used to match headers
+     *                          in allTomlSectionHeaders (e.g. "tool.poetry")
+     * @param allTomlHeaders    list of all section headers in the file (strings between '[' and ']')
+     * @return                   {@code true} if the section has neither literal
+     *                          values nor inline tables; {@code false} otherwise
+     */
+
+    public static boolean isSectionEmpty(Config sectionConfig, String parentHeaderName, List<String> allTomlHeaders){
+        // extract the names of any nested Configs under parent section
+        List<String> childConfigKeys = getConfigKeys(sectionConfig);
```
chang-annie, line 446: I: for example, if we have:
```
[tool.poetry]

[tool.poetry.dependencies]
ruff = "xyz"

[tool.poetry.group.dev.dependencies]
ruff = "xyz"
```
then,
`allTomlHeaders = [tool.poetry, tool.poetry.dependencies, tool.poetry.group.dev.dependencies]`
`childConfigKeys = [dependencies, group]`

looking at each nested config, we see that `dependencies` matches `tool.poetry.dependencies` and `group` matches `tool.poetry.group.`, so we know that there are no inline tables and therefore the original `[tool.poetry]` section is empty.

-------------

if we have:
```
[tool.poetry]
dependencies = { ruff = "xyz" }

[tool.poetry.group.dev.dependencies]
ruff = "xyz
```

then, 
`allTomlHeaders = [tool.poetry, tool.poetry.group.dev.dependencies]`
`childConfigKeys = [dependencies, group]`

looking at each nested config, we note that `dependencies` does NOT match any of the toml headers, so therefore it is an inline table and `[tool.poetry]` is not empty

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/TomlUtils.java
lemuelwatts: S: I think maybe we'd want to parameterize the regex here like we have for `PACKAGE_AND_VERSION_REGEX` and `OPERATOR_AND_VERSION_REGEX`

chang-annie: added `TOML_SECTION_HEADER_REGEX` and updated accordingly!

lemuelwatts (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryRemoveEmptyTomlMigration.java
```diff
@@ -0,0 +1,115 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Provides functionality to remove empty headers from pyproject.toml file
+ * Runs when on Poetry v2.0.0 or later.
+ */
+public class PoetryRemoveEmptyTomlMigration extends AbstractPoetryMigration{
+    private static final Logger logger = LoggerFactory.getLogger(PoetryRemoveEmptyTomlMigration.class);
```
d-ryan-ashcraft, line 22: A: no need to change, but Java member variables are false by default

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/TomlUtils.java
```diff
@@ -376,4 +387,82 @@ public static boolean hasComparators(String string) {
...
     }
 
+    /**
+     * Reads a TOML file and extracts every section header (the text between '[' and ']').
+     *
+     * @param file  the TOML file to scan
+     * @return      a list of section names without surrounding brackets
+     */
+    public static List<String> extractTomlSectionHeaders(File file){
+        Pattern headerPattern = Pattern.compile(TOML_SECTION_HEADER_REGEX);
```
d-ryan-ashcraft, line 397: A: minor, but you get better performance when you store the Pattern as a static final, when possible.

d-ryan-ashcraft (approved):

Issue #333, opened 2025-04-08
Help users diagnose/fix PyEnv Python installation issues


lemuelwatts:
## Summary
We would like to add a Pyenv version enforcer to ensure that users of Habushu are using a version of Pyenv that supports the minimum required Python version (3.9), which is also the lowest version supported by our minimum required Poetry version.

Provide functionality to enforce a specific PyEnv version like uv (#308) and poetry (#302)

The current error displayed is hidden and not immediately obvious especially when running with `mvnd`.

Error message when running pyenv 2.3.25 using python 3.11.4:
```
[INFO] --- enforcer:3.5.0:enforce (enforce-poetry) @ habushu-poetry-add-to-project ---
[INFO] Rule 0: org.technologybrewery.habushu.enforcer.RequirePoetryVersionRule(requirePoetryVersion) passed
[INFO] 
[INFO] --- habushu:3.0.0-SNAPSHOT:validate-python-package-and-dependency-manager (default-validate-python-package-and-dependency-manager) @ habushu-poetry-add-to-project ---
[INFO] Found Poetry 2.1.2
[INFO] Executing pyenv command: pyenv versions --bare
[INFO] Could not find Python version 3.12.9 in following versions [3.11.4] that are installed via pyenv. Installing version 3.12.9 now...
[INFO] Executing pyenv command: pyenv install 3.12.9
[ERROR] python-build: definition not found: 3.12.9

See all available versions with `pyenv install --list'.

If the version you need is missing, try upgrading pyenv:

  brew update && brew upgrade pyenv
```

## DOD

- [x] Add `RequirePyenvVersionRule` with logic to enforce a specific version/range of acceptable pyenv versions
     - Check if installed pyenv supports requested version: `pyenv install --list`
     - `>=3.12.9`?
     - Make sure version of pyenv we enforce is compatible with lowest required versions of poetry/uv/python
- [x] add example in `examples` folder under `poetry` (make sure it runs immediately after `habushu-poetry-enforcer-rule`)
- [x] Update relevant documentation with pyenv/python enforcer information

## Test Steps
**Initial Setup**
1. Pull latest version of habushu and run the following to build the latest version of the plugin: `mvn clean install -Pbootstrap`
2. Check your pyenv version by running `pyenv --version` and you should receive output similar to: `pyenv 2.5.4`

**Validate enforcer success**

[...]

PR #362, opened 2025-04-25
[#333] Add optional Maven Enforcer Rule for pyenv version

lemuelwatts:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -110,28 +99,43 @@ public List<String> validatePackageAndDependencyManagerInstallationAndVersion(
             ValidationTrackingStatus validationTracker, List<String> missingRequiredToolMsgs) {
             missingRequiredToolMsgs = validatePoetryInstallationAndVersion(validationTracker, missingRequiredToolMsgs);
             if (usePyenv) {
-                missingRequiredToolMsgs = validatePyenvInstallation(missingRequiredToolMsgs);
-                missingRequiredToolMsgs = validatePyenvConfiguration(missingRequiredToolMsgs);
+                missingRequiredToolMsgs = validatePyenvInstallationandVersion(missingRequiredToolMsgs);
```
lemuelwatts: I: Consolidated `validatePyenvInstallation` and `validatePyenvConfiguration` into a single function

chang-annie: we should rename `validatePyenvInstallationandVersion` to `validatePyenvInstallationAndVersion`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -110,28 +99,43 @@ public List<String> validatePackageAndDependencyManagerInstallationAndVersion(
...
 
-    private List<String> validatePyenvConfiguration(List<String> missingRequiredToolMsgs) {
+    private List<String> validatePyenvInstallationandVersion(List<String> missingRequiredToolMsgs) {
+        PyenvCommandHelper pyenvHelper = createPyenvCommandHelper();
+        log.debug("Validating pyenv installation/version and configuration...");
+        Pair<Boolean, String> pyenvInstallStatusAndVersion = pyenvHelper.getIsPyenvInstalledAndVersion();
+        // Check if pyenv is installed and with correct version
+        if (Boolean.FALSE.equals(pyenvInstallStatusAndVersion.getLeft())) {
+            missingRequiredToolMsgs.add(
+                    "'pyenv' is not currently installed! Please install pyenv and try again. Visit https://github.com/pyenv/pyenv for more information.");
```
chang-annie, line 114: q: why do we duplicate the messages here and in `RequirePyenvVersionRule`? can we pull in the messages from there instead?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PyenvCommandHelper.java
```diff
@@ -192,4 +193,13 @@ protected ProcessExecutor createPyenvExecutor(List<String> arguments) {
         fullCommandArgs.addAll(arguments);
         return new ProcessExecutor(workingDirectory, fullCommandArgs, Platform.guess(), null);
     }
+
+    public Pair<Boolean, String> getIsPyenvInstalledAndVersion() {
```
chang-annie, line 183: looks like this new method replaces `isPyenvInstalled`. we could probably remove it since it looks like it's not being used anywhere

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
meliz19: I'm leaning against the idea of pulling in the messages from the `RequirePyenvVersionRule`. Ther reason is that this set of code will run even when the enforcer rule is "off". If we pull in methods from the enforcer here, I'm concerned that it might lead to confusion down the road.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
lemuelwatts: Updated function name with capitalized A!

meliz19 (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/PyenvCommandHelper.java
lemuelwatts: good catch. removing isPyenvInstalled

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
lemuelwatts: Going to leave the message as is

Issue #363, opened 2025-04-28
Create uv example for withGroups configuration


meliz19:
## Description 
In doing #348, we identified several configs that could use example projects to illustrate how the configs work. The purpose of this ticket is to develop an example for `withGroups` in uv.

## DoD
- [x] Develop a uv project that uses the `withGroups` habushu configuration
- [x] Add/Update README documentation where relevant

## Test Steps
1. Build the habushu-maven-plugin
```bash
mvnd clean install -Pbootstrap
```

2. Build the example project without the build cache so that the `pom.xml` updates are captured.
```bash
mvnd clean install -pl :habushu-uv-dependency-groups -Dmaven.build.cache.skipCache
```

3. In the `examples/uv/habushu-uv-dependency-groups/pom.xml` we specify that `groupa` is to be included in the virtual environment sync via the `withGroups` configuration. To verify that the `groupa` dependency group was included, open  the `examples/uv/habushu-uv-dependency-groups/.venv/lib/python3.11/site-packages` and 
    - Verify that the `groupa` dependency (`numpy`) is included 

4. Open `examples/uv/habushu-uv-dependency-groups/pom.xml` and comment out the `withGroups` configuration.
5. Re-run the build command in step 2. 
6. In the `examples/uv/habushu-uv-dependency-groups/.venv/lib/python3.11/site-packages`,
    - Verify that the `groupa` dependency (`numpy`) is NOT included (Note: the groupA dependency group is optional and therefore will not be installed by default).

## Notes
- In the process of doing this ticket, I was unable to come up with a useful example illustrating the use of the `withoutGroups` configuration for a uv project. Instead of removing this functionality from Habushu, we are opting to keep it in and warn users that uv's `uv sync --no-group` functionality is currently underdeveloped and therefore does not possess strong useful functionality at this time.

meliz19:
Completed Definition of Done (DoD) & Over the Shoulder (OTS) with @colinpalmer-pro

chang-annie:
all test steps are passing! closing ticket as done.

- build with bootstrap profile
<img width="727" alt="Image" src="https://github.com/user-attachments/assets/2536ea22-80f0-4dda-87db-e967d71c33e0" />

- build module
<img width="742" alt="Image" src="https://github.com/user-attachments/assets/0c22c8d1-5a3b-496c-b67c-04d820f950cb" />

- confirm `withGroups` works as expected
[...]

PR #369, opened 2025-05-02
[#363] Add uv example demonstrating withGroups configuration

meliz19:
#363

### DIFF
--- a/docs/CONFIGURATION_README.md
+++ b/docs/CONFIGURATION_README.md
@@ -288,8 +288,7 @@ Default: None
 
 **Examples:**
 - [Poetry Dependency Groups](../examples/poetry/habushu-poetry-dependency-groups/README.md)
-
-[//]: # (- [uv Dependency Groups]&#40;TODO&#41;)
+- [uv Dependency Groups](../examples/uv/habushu-uv-dependency-groups/README.md)
 
 ### withoutGroups
 Specifies dependency groups to exclude within the installation.  Example usage:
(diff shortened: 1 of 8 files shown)


docs/CONFIGURATION_README.md
```diff
@@ -314,11 +313,12 @@ or
 
 Default: None
 
+**NOTE:** While Habushu does support this configuration for `uv`, `uv sync --no-group <group>` is underdeveloped and does not possess
```
meliz19, line 316: I: Added this note to let users know that while we do support this config, uv support for `--no-groups` is underdeveloped.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvAuthenticationCommandHelper.java
```diff
@@ -86,6 +86,7 @@ public void executeSyncCommand(String[] withGroups, String[] withoutGroups, int
             arguments.add(groupName);
         }
         for (String groupName : withoutGroups) {
+            logger.warn("While Habushu does support this configuration, `uv sync --no-group` is underdeveloped and does not possess strong useful functionality at this time.");
```
meliz19, line 89: I: Added this warning too to let users know that while we do support this config, uv support for `--no-groups` is underdeveloped.

lemuelwatts (approved):

PR #377, opened 2025-05-13
371 - Remove references to black and pylint

aaron-gary:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/HabushuUtil.java
```diff
@@ -136,6 +137,20 @@ public static void writeLinesToFile(String commands, String filePath) {
         }
     }
 
+    /**
+     * Writes the given content to the given file
+     * @param targetFile The file to write content to
+     * @param fileContent The content to write to file
+     * @throws IOException
+     */
+    public static void writeFile(File targetFile, String fileContent) throws IOException {
```
aaron-gary, line 146: I: There was another method in this class that wrote contents to a file, but it used a string path, whereas, in our case, we already had a File type.

habushu-maven-plugin/src/test/resources/migration/dev-dependencies/with-pylint-and-black-dependencies.toml
```diff
@@ -0,0 +1,26 @@
+[tool.poetry]
+name = "with-dependencies-is-both-groups"
+version = "2.9.0.dev"
+description = "Test moving dev-dependencies dependecies to group.dev.dependencies group when group.dev.dependencies group already exists"
+authors = ["Test <blah@foobar.com>"]
+license = "MIT License"
+
+[tool.poetry.dependencies]
+python = "^3.11"
+#dependencies = {black = ">=22.1.0", pylint = "^3.1.0"}
```
epwilkins, line 10: Q: Why is this commented out?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/RemovePylintCommentsMigration.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Provides functionality to remove pylint comments (for pylint filtering) from python files
+ */
+public class RemovePylintCommentsMigration extends AbstractPoetryMigration {
+    public static final String PYLINT_COMMENT = "# pylint";
```
epwilkins, line 22: Q: Any chance the is more or less than 1 space character?

epwilkins (approved):

habushu-maven-plugin/src/test/resources/migration/dev-dependencies/with-pylint-and-black-dependencies.toml
aaron-gary: Good catch.  I need to remove that line.  I tried to replicate the inline table of dependencies for black and pylint, but poetry throws a syntax exception when it tries to read the pyproject.toml.  My rational is that by using poetry to remove dependencies (in the migrate method) then we can trust that poetry knows how to handle whatever we throw at it (with valid syntax)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/RemovePylintCommentsMigration.java
aaron-gary: Good question.  Let me investigate

aaron-gary: Pylint syntax wise, all the examples are single space, but ultimately, users can add spaces and the migration would miss it.  So.. I'll add a slack task to my backlog to address this.

Issue #375, opened 2025-05-12
Update containerize deps Dockerfile velocity templates


meliz19:
## Description
The purpose of this ticket is to continue the containerize dependencies Mojo works started in #367.

## DoD
- [x] Replace pip with pipx in the builderStage Dockerfiles for Poetry
- [x] Add the `chomd` option with 744 as the default in the finalStage Dockerfiles for Poetry & uv

Follow ons:
- Add unit tests for the uv containerize dependencies goal
- Abstract ContainerizeDependenciesHelper and extend for uv & Poetry

## Test Steps
1. Build the habushu-maven-plugin
```bash
mvnd clean install -Pbootstrap
```

2. Build the Poetry & uv examples
```bash
mvnd clean install
```

3. Open `examples/poetry/habushu-poetry-containerize/src/main/resources/docker/Dockerfile` and verify `pipx` is installed in the `HABUSHU_BUILDER_STAGE` and you see `COPY --from=habushu_builder --chown=1001 --chmod=744 /opt/venv /opt/venv/` in the  `HABUSHU_FINAL_STAGE`.

4. Open `examples/uv/habushu-uv-containerize/src/main/resources/docker/Dockerfile` and verify you see `COPY --from=habushu_builder --chown=1001 --chmod=744` in the `HABUSHU_FINAL_STAGE`.

5. Build and run the `habushu-poetry-containerize` docker image
```bash
docker build -t habushu-poetry-containerize -f ./examples/poetry/habushu-poetry-containerize/src/main/resources/docker/Dockerfile ./examples/poetry/habushu-poetry-containerize
```

```bash
docker run --rm -it habushu-poetry-containerize
```

6. Verify you see `Worker is running`.
7. Build and run the `habushu-uv-containerize` docker image
```bash
docker build -t habushu-uv-containerize -f ./examples/uv/habushu-uv-containerize/src/main/resources/docker/Dockerfile ./examples/uv/habushu-uv-containerize
```

```bash
docker run --rm -it habushu-uv-containerize
```

8. Verify you see `Worker is running`.
9. Clean up images
```bash
docker rmi $(docker images | grep habushu | tr -s ' ' | cut -d ' ' -f 3)
```

meliz19:
Completed DoD/OTS with @nartieri

lemuelwatts:
Testing completed!

Screenshot of successful `habushu-poetry-containerize` image and output:
[...]

PR #376, opened 2025-05-12
updated README, added containerize-deps user configs and simplified u…

meliz19:
…pdateDockerfileWithContainerStageLogic parameters

#372

Issue #372, opened 2025-05-09
Simplify updateDockerfileWithContainerStageLogic method, convert dockerfile .vm paths to habushu configs & update readme configs


meliz19:
## Description
The purpose of this ticket is to continue the containerize dependencies Mojo works started in #367.

## DoD
- [x] Add Dockerfile .vm template parameters so that users can configure their own templates, if desired.
- [x] Update `updateDockerfileWithContainerStageLogic` to accept ContainerizeDependenciesMojo instead of all the current parameters
- [x] Update README documentation to include all possible Mojo configurations

## Test Steps
1. Build the habushu-maven-plugin
```bash
mvn clean install -Pbootstrap
```

2. Open the following files and replace the content except for the last line (with the `CMD` command) with the txt snippet below.
  - `examples/poetry/habushu-poetry-containerize/src/main/resources/docker/Dockerfile` and 
  - `examples/uv/habushu-uv-containerize/src/main/resources/docker/Dockerfile` 
```txt
#HABUSHU_BUILDER_STAGE

#HABUSHU_FINAL_STAGE
```

3. Build the Poetry & uv examples
```bash
mvn clean install
```

4. Build and run the `habushu-poetry-containerize` docker image
```bash
docker build -t habushu-poetry-containerize -f ./examples/poetry/habushu-poetry-containerize/src/main/resources/docker/Dockerfile ./examples/poetry/habushu-poetry-containerize
```

```bash
docker run --rm -it habushu-poetry-containerize
```

5. Verify you see `Worker is running`.
6. Build and run the `habushu-uv-containerize` docker image
```bash
docker build -t habushu-uv-containerize -f ./examples/uv/habushu-uv-containerize/src/main/resources/docker/Dockerfile ./examples/uv/habushu-uv-containerize
```

```bash
docker run --rm -it habushu-uv-containerize
```

8. Verify you see `Worker is running`.
9. Build the Poetry & uv containerize dependencies examples using the `custom-docker-template` profile
```bash
mvn clean install -Dmaven.build.cache.skipCache -pl :habushu-poetry-containerize,:habushu-uv-containerize -Pcustom-docker-template
```
10. Build the `habushu-poetry-containerize` docker image
```bash
docker build -t habushu-poetry-containerize -f ./examples/poetry/habushu-poetry-containerize/src/main/resources/docker/Dockerfile ./examples/poetry/habushu-poetry-containerize
```
[...]

PR #378, opened 2025-05-14
[#375] Replace pip with pix & add chmod=744 to Dockerfile templates

meliz19:
#375

### DIFF
--- a/docs/CONFIGURATION_README.md
+++ b/docs/CONFIGURATION_README.md
@@ -689,6 +689,12 @@ The user to set as the owner of the virtual env. This is useful when the Docker 
 
 Default: `1001`
 
+### dockerVenvDirectoryPermissions
+
+The directory permissions for the virtual env. Set to an empty string to disable.
+
+Default: `744`
+
 ### dockerTemplatePath
 
 Overwrite with Docker template path if a custom template is preferred.
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
@@ -60,11 +60,12 @@ public class ContainerizeDepsDockerfileHelper {
     }
 
     private void setSharedVelocityTemplateContext(AbstractContainerizeDepsVelocityContext context,
-                                                  String anchorDirectory, String singleRepoProjectDir, String owner,
+                                                  String anchorDirectory, String singleRepoProjectDir,
                                                   String baseImage) {
         context.setAnchorDirectory(anchorDirectory);
         context.setSingleRepoProjectDir(singleRepoProjectDir);
-        context.setOwner(owner);
+        context.setOwner(containerizeDepsMojo.getDockerUser());
+        context.setVenvDirectoryPermissions(containerizeDepsMojo.getDockerVenvDirectoryPermissions());
         context.setBaseImage(baseImage);
     }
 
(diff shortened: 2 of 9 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
```diff
@@ -60,11 +60,12 @@ private void setVelocityPropertiesForFilePath(File templateDir) {
     }
 
     private void setSharedVelocityTemplateContext(AbstractContainerizeDepsVelocityContext context,
-                                                  String anchorDirectory, String singleRepoProjectDir, String owner,
```
meliz19, line 63: I: Fixed missed parameter simplifications in the last issue (#372)

lemuelwatts (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
```diff
@@ -60,11 +60,12 @@ private void setVelocityPropertiesForFilePath(File templateDir) {
...
+                                                  String anchorDirectory, String singleRepoProjectDir,
                                                   String baseImage) {
         context.setAnchorDirectory(anchorDirectory);
         context.setSingleRepoProjectDir(singleRepoProjectDir);
```
epwilkins, line 66: A: In the same vein as replacing the huge list of method parameters with just passing the mojo, it might be worth just setting the mojo on the context.

habushu-maven-plugin/src/main/resources/templates/dockerfile_poetry_builder_stage_template.vm
```diff
@@ -1,8 +1,20 @@
 FROM $baseImage AS habushu_builder
 
-# Poetry and supporting plugin installations
+# Install pipx
+RUN apt-get update && \
```
epwilkins: C: Using apt makes this template way less portable. I think we need to assume the base image has what it needs. If that isn't true for our default base image we may need to discuss alternatives.

habushu-maven-plugin/src/main/resources/templates/dockerfile_poetry_builder_stage_template.vm
```diff
@@ -1,8 +1,20 @@
...
+    apt-get install -y --no-install-recommends \
+    ca-certificates && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN pip install --no-cache-dir --break-system-packages pipx
```
epwilkins: A: A bit weird to call pip and then call ensurepip on the next line

docs/CONFIGURATION_README.md
```diff
@@ -689,6 +689,12 @@ The user to set as the owner of the virtual env. This is useful when the Docker
 
 Default: `1001`
 
+### dockerVenvDirectoryPermissions
+
+The directory permissions for the virtual env. Set to an empty string to disable.
+
+Default: `744`
+
 ### dockerTemplatePath
```
epwilkins, line 698: Q: What's the benefit of setting the path separately from the file name? I'd think its easy enough to make a pom property if needed.

habushu-maven-plugin/src/main/resources/templates/dockerfile_poetry_builder_stage_template.vm
```diff
@@ -1,8 +1,20 @@
...
 RUN python -m ensurepip --upgrade && \
-    pip install poetry==$poetryVersion && \
+    python -m pipx ensurepath
```
epwilkins: A: This isn't necessary, since the PATH update won't impact the Dockerfile commands, and we're updating the PATH env var on line 14.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/ContainerizeDepsDockerfileHelper.java
meliz19: As part of this ticket series effort, I'm going to reconfigure this file to be an abstract method and extend it for poetry and uv. I'll look into this suggestion in that ticket.

docs/CONFIGURATION_README.md
meliz19: See the `setVelocityPropertiesForFilePath` method in this [PR](https://github.com/TechnologyBrewery/habushu/pull/376/files#diff-d466e0619bc9e8d62697deef647963486ee6e39a5cccbe69ae41419dc71f73a9).

PR #391, opened 2025-06-04
[#390] add pytest test support

### LINKED ISSUE #390: Habushu should natively support pytest
## Description
Enable habushu to support the Pytest test for uv based project

## DOD
Acceptance criteria required to realize the requested feature
- [x] Create PytestBddTestMojo class to handle `pytest-test` goal
- [x] Include the `pytest-test` into the habushu default lifecycle
- [x] Create PytestTestUv class for pytest uv specific command
- [x] Update the README docs
- [x] Add a `habushu-uv-pytest` example

## Test Strategy/Script
How will this feature be verified?
Have the build succeed and example test passes

## References/Additional Context
Add any other context, links, or screenshots about the feature request here.

csun-cpointe:


docs/CONFIGURATION_README.md
```diff
@@ -444,6 +454,93 @@ Controls whether skipped tests should be completely omitted from test reports ra
...
+        <configuration>
+          <pytestTestEnvironmentVariables>
+            <ENV_VAR>VALUE</ENV_VAR>
+          </pytestTestEnvironmentVariables>
+        </configuration>
+      </execution>
+    </executions>
+</plugin>
+```
+To pass environment variables in via the command line using the `-D` option, use a property placeholder in the pom.xml:
```
epwilkins, line 496: A: This feels a little too far into documenting how Maven works. I'm on the fence though and can see an argument for its usefulness.

docs/CONFIGURATION_README.md
```diff
@@ -219,6 +224,12 @@ Developers will typically **not** modify this property but is made available for
 
 Default: `${project.basedir}/tests`
 
+### testPackage
+
+The test package used for testing.
+
+Default: `behave`
```
epwilkins, line 231: S: Recommend trying to infer from the `pyproject.toml` and fall back to `behave` if neither behave nor pytest are present.

examples/uv/habushu-uv-pytest/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+from typing import Literal
+
+import pytest
+
+
+@pytest.fixture()
+def sample_fixture() -> Literal[True]:
+    """Placeholder fixture available to all tests.
+
+    This is part of the repo template and can be deleted once used.
```
epwilkins: Q/A: Was this docstring from Pytest and if so should we update it?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/BehaveBddTestMojo.java
```diff
@@ -103,14 +103,18 @@ public Map<String, String> getBehaveTestEnvironmentVariables() {
...
+                BehaveBddTestUv behaveBddTestUv = new BehaveBddTestUv(getLog(), this,
+                        createUvCommandHelper());
+                behaveBddTestUv.doExecute();
+            }
         } else {
-            BehaveBddTestUv behaveBddTestUv = new BehaveBddTestUv(getLog(), this,
-                    createUvCommandHelper());
-            behaveBddTestUv.doExecute();
+            getLog().info(String.format("This Mojo is not used for %s test.", testPackage));
         }
```
csun-cpointe, line 118: I: view with hide whitespace for easier review

examples/uv/habushu-uv-pytest/tests/conftest.py
csun-cpointe: yeah.. I can remove this docstring.

docs/CONFIGURATION_README.md
csun-cpointe: There could be no `behave`/`pytest` dep in the `pyproject.toml` file when we run test, which will be automatically added during test if it's not installed. I can have a check in the code to ensure that it always fall back to `behave` if it's neither `behave` or `pytest`

epwilkins (approved):

docs/CONFIGURATION_README.md
```diff
@@ -444,6 +454,75 @@ Controls whether skipped tests should be completely omitted from test reports ra
...
+```
+Then you can specify the value from the cli: `mvn clean install -Dhabushu.ENV_VAR=customValue`. A default value can be
+set in the `<properties>` tag of the pom.xml.
+
+Default: None
+
+**Examples:**
+- [Pytest with uv](../examples/uv/habushu-uv-pytest)
+
+### skipTests
```
woods-cpointe, line 518: s: Might want to also call out that this will skip the behave tests and vice versa

docs/CONFIGURATION_README.md
epwilkins: We discussed this further offline.
The current flow is (where `testPackage` is a POM configuration):
```
if testPackage is pytest
  if pytest is not installed
    install pytest
  run pytest
else
  if behave is not installed
    install behave
  run behave
```
I suggested this flow:
```
if pytest is not installed and behave is not installed
  if testPackage is pytest
    install pytest
  else
    install behave
if behave installed
   run behave
if pytest installed
   run pytest
```
The benefit of this is that, if someone uses their Python build tool directly to add tests, Habushu just transparently picks them up and executes them. If they preferred configuring directly via the POM, then Habushu will install the missing test package for them to bootstrap the project. However, this could be counter-intuitive if someone started a project using pytest, then added Habushu later and omitted the `testPackage` configuration. This would result in Habushu installing and running behave, even though the project is using pytest.

Ultimately, it felt like there were confusing scenarios for both code flows, so we opted to leave it configuration-driven for now. It is easy enough add "detection" logic in the future if we want.

woods-cpointe: s: I question if the configuration is needed. We could always attempt to execute both and if one or both of them cant find tests then it will skip the mojo. Effectively the same as this config but a little later. 

One downside is it would cause excessive logging. If you have behave tests, it could be confusing to see the pytest mojo warn that there are no tests. Could update it to a log that just states no pytest tests found

docs/CONFIGURATION_README.md
csun-cpointe: good catch!!

PR #412, opened 2025-07-09
#410 wheel containerization staging

### LINKED ISSUE #410: Containerization: Stage wheels from the build output directory for use by the Dockerfile
The [initial attempt](https://github.com/TechnologyBrewery/habushu/issues/149) at creating a virtual environment to package a Python application on a Docker container built the application and all monorepo dependencies from scratch on the target container.  This causes issues in use cases that include source code generation, and results in longer build times.  Instead, we should use the same logic we currently have for collecting the source code and instead collect the wheels from the build output directory.  Then install the wheels such that each monorepo dependency wheel is installed before its dependent project.

**Definition of Done**
- ✅ Collect the wheels instead of the source code in the containerization-support directory
  - What if there are multiple wheels? 
    - ✅ Hard-code *-none-any.whl
    - ❌ accept any *.whl and use the first
    - ❌ add configs for wheel name parts and default to *-none-any
  - Can assume output dir is `dist` as there is no way to set `--output-directory` during build phase currently
- ✅ Update CONFIGURATION.md to remove deprecated configs
- ✅ Add `dockerVenvDirectoryPermissions` to [TOC](https://github.com/TechnologyBrewery/habushu/blob/dev/docs/CONFIGURATION_README.md#configuration-options)

**Test Steps**
1. Build the snapshot of the plugin
    ```sh
    mvn clean install -Pbootstrap
    ```
2. Create a test project with a Python library module, a Python application module, and a module to package the application as a Docker image
3. Add the containerization goal to the Docker module
4. Run the build
5. Ensure the library and application wheels were copied to the staging directory of the Docker module. (Default is _target/containerization-support_)

epwilkins:
Split into two commits for easier review:
 - (8dd683d) [410] code cleanup/refactor
    Pull logic for checking the python package version into the
    CommandHelper of each package manager.  Unify `UvCommandHelper` and
    `UvAuthenticationCommandHelper`.
 - (3c49f5f) [410] update c14n staging logic to grab wheels
    Update containerization logic that stages necessary Habushu projects so
    that it grabs the built wheel files in addition to the source code. Also:
     - update version to 3.2.0-SNAPSHOT
     - add dockerVenvDirectoryPermissions to TOC in CONFIGURATION.md


examples/habushu-default-python-strategy/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.technologybrewery.habushu</groupId>
         <artifactId>examples</artifactId>
-        <version>3.1.2-SNAPSHOT</version>
+        <version>3.2.0-SNAPSHOT</version>
```
epwilkins, line 8: I: Bumped to 3.2.0 given the significant change to containerization

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -240,6 +249,35 @@ protected void stageSourcesForProject(Path sourceRoot, Path destRoot, FileSet so
...
+            String wheelGlob = normalizedName + "-" + version + "*-*-none-any.whl";
+            logger.info("Searching for wheels matching {}", wheelGlob);
+            try (DirectoryStream<Path> ds = Files.newDirectoryStream(distDir, wheelGlob)) {
+                for (Path wheel : ds) {
+                    if (copiedWheel != null) {
+                        throw new HabushuException(
+                                String.format("Multiple wheels found for project [%s]! found %s and %s",
+                                        projectDir.getFileName(),
+                                        copiedWheel,
+                                        wheel.getFileName()));
```
epwilkins, line 273: I: I considered logging this as a warning, but reasoned that in any scenario where we match more than one wheel it could create hard to track bugs.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -123,7 +151,148 @@ public List<String> createToolRunCommand(List<String> additionalArguments) {
         return arguments;
     }
 
-    public String executePythonPinCommand() {
+    public void executeLockCommand(AbstractHabushuMojo mojo, boolean skipUvLockRefresh, boolean skipUvCheck) {
```
epwilkins, line 154: I: This was copied basically as-is from `UvAuthenticationCommandHelper` except that instead of the relevant mojo being stored on the helper, it's passed in as a parameter.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -123,7 +151,148 @@ public List<String> createToolRunCommand(List<String> additionalArguments) {
...
+    private String getUvIndexPasswordEnvironmentVariable(String repoId){
+        return String.format("UV_INDEX_%s_PASSWORD", getExportParameter(repoId));
+    }
+
+    private Boolean IsUvIndexUsernameEnvironmentVariableSet(String repoId){
+        String uvIndexUsernameEnvironmentVariable = getUvIndexUsernameEnvironmentVariable(repoId);
+        return !StringUtils.isEmpty(System.getenv(uvIndexUsernameEnvironmentVariable));
+    }
+
+    private Boolean IsUvIndexPasswordEnvironmentVariableSet(String repoId){
```
woods-cpointe: a: couple of capitalized function names

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
```diff
@@ -123,7 +151,148 @@ public List<String> createToolRunCommand(List<String> additionalArguments) {
...
+
+        if (executeCommandWithCredentials(mojo)) {
+            Map<String, String> privateRepoCredentials = getPrivateRepoCredentials(mojo, getRepoId(mojo));
+            execute(arguments, privateRepoCredentials);
+        } else {
+            execute(arguments);
+        }
+    }
+
+    public void executeLockCommandAndLogAfterTimeout(InstallDependenciesMojo mojo, boolean skipUvCheck, int timeout, TimeUnit timeUnit, String messageToDisplay) {
```
woods-cpointe, line 165: a: this function seems almost identical to the other lock. Could probably simplify it a bit

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -444,4 +482,25 @@ public MavenProject getPrimaryProject() {
             return primaryProject;
         }
     }
+
+    private static class ProjectInfo {
```
woods-cpointe: a: can be made public if we are exposing it in `getAllProjects`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -193,20 +200,21 @@ protected void doExecute() throws MojoExecutionException, MojoFailureException {
...
-            if (project.equals(projectCollection.getPrimaryProject())) {
+            if (info.getProject().equals(projectCollection.getPrimaryProject())) {
                 primaryProjectPath = relativeProjectPath;
             }
 
             FileSet sourceFileSet = getSourceSet();
             sourceFileSet.setDirectory(projectPath.toString());
-            stageSourcesForProject(sourceRoot, destRoot, sourceFileSet, relativeProjectPath);
+            Path wheelName = stageProject(sourceRoot, destRoot, sourceFileSet, relativeProjectPath);
+            info.setWheelFile(wheelName);
```
woods-cpointe, line 217: q: do we use this projectInfo.wheelFile anywhere?

woods-cpointe: Greater question is why update it from `MavenProject` to `ProjectInfo` if we dont use the wheelFile

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/exec/UvCommandHelper.java
epwilkins: I can do a bit of cleanup but trying not to change the code too much to avoid introducing regressions from the refactor.

woods-cpointe (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
epwilkins: We will use the wheel file in the follow on when we update the Dockerfile to install the wheels. At that point, it will be important to know the exact name of each wheel staged.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
epwilkins: `ProjectStack` is protected (effectively private) so I reasoned this should be similarly scoped.  I could make it protected so it matches though.

PR #421, opened 2025-08-19
HAB-420 Resolve Dependabot Alerts

### LINKED ISSUE #420: Resolve Dependabot alerts
Resolve current [Dependabot Alert](https://github.com/TechnologyBrewery/habushu/security/dependabot)s (currently 3)

### DoD

- [x] Update dependency versions to resolve [Dependabot alerts](https://github.com/TechnologyBrewery/habushu/security/dependabot)

### Test Steps
Build the project. Ensure the following commands succeed:

1. Pull down most recent changes
2. `mvnd clean install -Pbootstrap`
3. `mvnd clean install -Pdefault`

Inspect mvn dependency tree

1. Run the following command:
```
mvn dependency:tree \
-Dincludes=com.google.code.gson:gson,org.apache.commons:commons-lang3
```

2. In the terminal search for `com.google.code.gson:gson` and ensure it does not appear in the dependency output
3. In the terminal search for `org.apache.commons:commons-lang3` and ensure each entry has version 3.18.0

colinpalmer-pro:
Resolving dependabot alerts:

- com.google.code.gson:gson 2.8.1 -> 2.8.9
- org.apache.commons:commons-lang3 3.15.0 -> 3.18.0

### DIFF
--- a/habushu-maven-plugin/pom.xml
+++ b/habushu-maven-plugin/pom.xml
@@ -57,7 +57,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.15.0</version>
+            <version>3.18.0</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
@@ -119,11 +119,6 @@
                     <artifactId>guava</artifactId>
                 </exclusion>
             </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>com.moandjiezana.toml</groupId>
-            <artifactId>toml4j</artifactId>
-            <version>0.7.2</version>
         </dependency>
         <dependency>
             <groupId>org.apache.maven</groupId>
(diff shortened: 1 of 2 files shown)


habushu-maven-plugin/pom.xml
```diff
@@ -121,9 +121,9 @@
             </exclusions>
         </dependency>
         <dependency>
```
colinpalmer-pro, line 123: i: artifact moved locations - see note at https://mvnrepository.com/artifact/com.moandjiezana.toml/toml4j

epwilkins: We should take care here.  This wasn't just a moved group.  The original project is no longer maintained, and this is a fork from the Hotmoka ecosystem.  It doesn't look like they are trying to take over maintenance but instead forked it for their own use.  That could make this fragile in the long run if they aren't expecting any outside users.  Also their domain is not over SSL and ATT at least has it marked as a phishing site. (Probably incorrect flag on ATT's part, but still raises some flags worth investigating.)

epwilkins: It looks like this is only used in `HabushuUtil.checkPythonPackageManager`. Everywhere else we use electronwill nightconfig.  I wonder if we should just convert this method to use the existing library and remove this dependency entirely.

_Edit:_ Yeah we definitely should move to NightConfig. It is under active maintenance, is on version 3.8 instead of 0.7, and supports TOML 1.0 instead of 0.4. I think the switch is pretty easy to make since we're just reading a config value.

colinpalmer-pro: Alright I'll take a look at that

epwilkins (approved):

PR #341, opened 2025-04-10
[#340] Add the dev repo authentication examples for poetry

meliz19:

PR #425, opened 2025-09-09
[#423] add private/dev indexes to dockerfile if set

### LINKED ISSUE #423: Containerization should be compatible with private repositories
**Background**
Follow on to #411. If the user has overridden pypiRepoUrl or has the dev repository enabled, the resulting dockerfile logic should account for that so packages from private repositories can be installed on the image. We will intentionally not support installing dependencies from multiple repositories (outside of the dev repo support) to comply with [pip's suggestion](https://github.com/pypa/pip/issues/5045#issuecomment-369521345). Users will need to configure their private repo's to "pass through" to any additional repos (e.g. pypi.org).

For now, we will intentionally leave out authentication against the private repo as our prior approach (before #410) did not support authentication and it was not an issue for the year or so that it was in place.  On request, we should be able to add authentication to our existing approach.

**Definition of Done**
- [x] Introduce a more centralized, consistent way to process/configure Python repository settings (with unit tests)
  - For now, we will just use this logic for our new case, but in the future we should refactor the logic in the following places to use the new centralized approach:
    - `InstallDependenciesMojo`
    - `PublishToPyPiMojo`
    - `ValidatePackageAndDependencyManagerMojo`
- [x] If `habushu.pypiRepoUrl` is not pypi.org, the containerization logic should add `--index-url <habushu.pypiRepoUrl>`
- [x] If `habushu.useDevRepo` is set to true, the containerization logic should add `--extra-index-url <habushu.devRepositoryUrl>`

**Test Steps**
1. Stand-up a private repository
2. Publish a package to the repository
3. Create a Habushu module that depends on the private package
4. Use the `containerize-dependencies` goal to create a Docker image for the module
5. Verify the image is successfully created and the private package is installed

epwilkins:
Also reverts change from #341 (c85f486) as authentication is not always required)

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
@@ -166,7 +166,7 @@ public class PyenvAndPoetrySetup extends AbstractPythonPackageAndDependencyManag
         PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
 
         if (StringUtils.isEmpty(username) || StringUtils.isEmpty(password)) {
-            throw new HabushuException("Your credentials for " + repoId + " must be set in your ~/.m2/settings.xml");
+            log.info(String.format("Did not find username and password for the server with <id> %s. Will use existing configuration.", repoId));
         } else {
             String configKey = String.format("http-basic.%s", repoId);
             log.info(String.format("Adding username and password configuration for %s", repoId));
--- a/habushu-maven-plugin/src/main/resources/templates/dockerfile_builder_stage_template.vm
+++ b/habushu-maven-plugin/src/main/resources/templates/dockerfile_builder_stage_template.vm
@@ -4,6 +4,13 @@ RUN --mount=type=bind,source=$stagingDirectory,target=/tmp/wheels \
     python3 -m venv /opt/venv && \
     /opt/venv/bin/python -m ensurepip --upgrade && \
     /opt/venv/bin/python -m pip install  \
+#if($privateRepoUrl)
+    --index-url="$privateRepoUrl" \
+#end
+#if($devRepoUrl)
+    # use caution, should not be used with public repos. See https://github.com/pypa/pip/issues/8606
+    --extra-index-url="$devRepoUrl" --pre \
+#end
 #foreach($wheel in $extraWheels)
     /tmp/wheels/$wheel \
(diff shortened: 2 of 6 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/PyenvAndPoetrySetup.java
```diff
@@ -166,7 +166,7 @@ public void registerRepositoryToSupportAuthenticatedDependencyResolution(String
         PoetryCommandHelper poetryHelper = createPoetryCommandHelper();
 
         if (StringUtils.isEmpty(username) || StringUtils.isEmpty(password)) {
-            throw new HabushuException("Your credentials for " + repoId + " must be set in your ~/.m2/settings.xml");
+            log.info(String.format("Did not find username and password for the server with <id> %s. Will use existing configuration.", repoId));
```
epwilkins, line 169: I: Reversion noted in summary -- not all private repos/actions require authentication

habushu-maven-plugin/src/main/resources/templates/dockerfile_builder_stage_template.vm
```diff
@@ -4,6 +4,13 @@ RUN --mount=type=bind,source=$stagingDirectory,target=/tmp/wheels \
     python3 -m venv /opt/venv && \
     /opt/venv/bin/python -m ensurepip --upgrade && \
     /opt/venv/bin/python -m pip install  \
+#if($privateRepoUrl)
+    --index-url="$privateRepoUrl" \
+#end
+#if($devRepoUrl)
+    # use caution, should not be used with pubic repos. See https://github.com/pypa/pip/issues/8606
+    --extra-index-url="$devRepoUrl" --pre \
```
epwilkins, line 12: I: `--pre` enables installation of pre-release versions, as would be typical with a dev repo.

jaebchoi (approved):

aaron-gary (approved):

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/ContainerizeDepsSteps.java
```diff
@@ -2,6 +2,7 @@
 
 import io.cucumber.java.After;
 import io.cucumber.java.Before;
+import io.cucumber.java.en.And;
```
cwoods-cpointe: a: unused import

woods-cpointe (approved):

PR #430, opened 2025-09-11
[#426] select newest wheel during c14n for snapshot deployment

### LINKED ISSUE #426: Containerization: An error is thrown when deploying a snapshot
## Description
This should really only happen for development snapshot builds that are being pushed/deployed.  However, during snapshot publishing, each Habushu module will build twice by default: once during `install` with the pre-release version of `dev` (i.e. `dev0`), and again during `deploy` with the pre-release version of `dev<TS>` (e.g. `dev1757432284715`) in order to ensure a unique package version is being pushed.  The current containerization logic simply throws an error if multiple wheels match the search pattern.

Technically the search pattern does include the version as defined in `pyproject.toml` of the dependency, but it does not add the implied `0` to the end of `dev`.  To resolve this we have 2 options:
 1. implement logic to fully resolve the version to what would be expected in the wheel name
    - technically more "correct" to the state of the project at containerization time
    - potentially automatically handles other Python versioning patterns like build descriptors
 2. choose whatever the latest wheel file is
    - easier to implement
    - uses published wheel in image (though theoretically they should be identical)
    - potentially easier to work around undiscovered edge cases

Because option 2 has more pros than 1, we should target that solution for now.

## Definition of Done
- [x] When multiple wheels match, choose the latest wheel by creation time

## Test Steps
1. Clone and build the latest Habushu `dev` changes: `mvn clean install -Pbootstrap`
2. Create a Maven project on a snapshot version (e.g. 1.0.0-SNAPSHOT)
    - or use the attached test project and skip to step 5:
    - [hab-426-test.tar.gz](https://github.com/user-attachments/files/22279982/hab-426-test.tar.gz)

3. Create a module with packaging `habushu` using Habushu 3.3.0-SNAPSHOT
4. Create a Docker module that containerizes the Habushu module
5. Configure Habushu's `default-publish-to-pypi-repo` execution to run during the `install` phase
6. Start a PyPI server on localhost:7000 so the `publish-to-pypi-repo` goal succeeds
    - For the test project, simply run `./start-local-pypi.sh` in a separate terminal, which will download and run [pypiserver](https://github.com/pypiserver/pypiserver?tab=readme-ov-file#pypiserver---minimal-pypi-server-for-use-with-pipeasy_install)
7. Ensure Docker is running and exposed via DOCKER_HOST
8. Run `mvn clean install`
    - Due to #428 you may get an error if you don't have credentials in your settings.xml file. For the test project, simply use the included settings i.e. `mvn clean install -s settings.xml`
9. Verify two wheels were created under the Habushu module's `dist` directory: one for dev0 and one with the current timestamp
10. Verify the Docker build installed the timestamped wheel using one or more of the indicators below:
    - the logs should indicate the timestamped wheel is being chosen over the dev0 wheel
    - the target/containerize-support directory should have the timestamped wheel
    - the updated Dockerfile should be calling `pip install` on the timestamped wheel
11. Tear down the local PyPI server with Ctrl+C

epwilkins:
When multiple wheels are found during containerization, choose the most recently modified wheel.

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
@@ -10,7 +10,6 @@ import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.model.fileset.FileSet;
-import org.codehaus.plexus.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.technologybrewery.habushu.exec.CommandHelper;
@@ -30,10 +29,12 @@ import java.nio.file.StandardCopyOption;
 import java.util.ArrayDeque;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Deque;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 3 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -193,16 +198,28 @@ protected Path stageProject(Path sourceRoot, Path destRoot, Path relativeProject
...
-                        String message = String.format("Multiple wheels found for project [%s]! found %s and %s",
-                                projectDir.getFileName(), copiedWheel, wheel.getFileName());
-                        throw new HabushuException(message);
-                    }
+                ds.forEach(wheels::add);
+            }
+            Path sourceWheel = null;
+            for (Path wheel : wheels) {
+                Path unusedWheel;
+                if (sourceWheel == null || Files.getLastModifiedTime(sourceWheel).compareTo(Files.getLastModifiedTime(wheel)) <= 0) {
```
epwilkins, line 208: I: Must compare against the source wheel as the copied wheel does not preserve the modified time

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/ContainerizeDepsSteps.java
```diff
@@ -47,15 +49,17 @@ public void configureMavenTestSession() throws Exception {
         // important for registering Habushu's mojos to AbstractTestCase.mojoDescriptors,
         // which ensures that lookupConfiguredMojo will return the configured mojo
         mojoTestCase.configurePluginTestHarness();
+        File sourceTestProject = new File("target/test-classes/containerize-dependencies/default-single-monorepo-dep");
+        File tempTestProject = new File(testMonorepoDepPath);
+        FileUtils.createParentDirectories(tempTestProject);
+        FileUtils.copyDirectory(sourceTestProject, tempTestProject);
```
epwilkins, line 54: I: Copying the entire test repo to a new location makes resetting the test much easier.

jaebchoi (approved):

csun-cpointe (approved):

PR #432, opened 2025-09-15
[#429] Add requirements file relocator for containerization

### LINKED ISSUE #429: Containerization: Relocate path-based dependencies from requirements.txt file to support containerization
When `rewriteLocalPathDepsInArchives` is set to `false`, the containerization process writes local file paths into the METADATA file inside Poetry generated wheel files that reference monorepo dependencies. During Docker builds, these file paths cannot be resolved, causing build failures. We need to investigate strategies to reliably work around this issue. UV projects do not run into the same issue.

**Example METADATA file:**
```
Metadata-Version: 2.3
Name: python_containerization_experiment_lib
Version: 0.1.0.dev0
Summary: Util code for python containerization experiment lib
Requires-Python: >=3.12,<4
Classifier: Programming Language :: Python :: 3
Classifier: Programming Language :: Python :: 3.12
Classifier: Programming Language :: Python :: 3.13
Requires-Dist: python-containerization-experiment-util @ file:///Users/anniechang/experiment-python-containerization/util
```

**Example error message:**
```
[INFO] DOCKER> 66.56 Processing /Users/anniechang/experiment-python-containerization/util (from python-containerization-experiment-lib==0.1.0.dev0)
[INFO] DOCKER> 66.56 ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/Users/anniechang/experiment-python-containerization/util'
```

## Steps to Reproduce
1. Create a Poetry project with a Python application module and a docker module to package the application as a Docker image
2. In the docker module, add the `containerize-dependencies` goal 
3. In the application module, set the `rewriteLocalPathDepsInArchives` to `false`
4. Build the project and note that the build will fail at the docker module because it cannot resolve the local file path that was included in the generated wheel file
5. (Optional) Unzip the generated wheel file - `unzip path/to/generated_wheel.whl` and note that there is local file path included in the `METADATA` file
```
Requires-Dist: python-containerization-experiment-util @ file:///Users/anniechang/experiment-python-containerization/util
```

## Expected Behavior
Poetry project with monorepo dependencies should be able to successfully containerize when the `rewriteLocalPathDepsInArchives` is set to `false

## Actual Behavior
Poetry project with monorepo dependency fails to containerize because local file path included in the generated wheel file.

**Definition of Done:**
- [x] Poetry projects should successfully containerize even with path dependencies
  - [x] investigate and implement solution such that final Poetry generated wheel files do not include local file paths when `rewriteLocalPathDepsInArchives` is set to `false`
  - [x]  leave comment with potential solutions

**Test Strategy**
- follow steps 1-3 in Steps to Reproduce
- use the `containerize-dependencies` goal to create a Docker image for the for the docker module
- confirm docker module is able to package the application as a Docker image and containerize successfully

chang-annie:
#429 
- updates containerization approach - use requirements.txt file as single-source of truth
  - parse requirements.txt file and pull out any path-based requirements
  - locate and stage wheel files into wheel house
  - update requirements.txt file to point to new location in wheelhouse
  - in Dockerfile, install requirements.txt then pip install wheel with `--no-deps`


csun-cpointe (approved):
Q: Do we need to update the readme file? since we now require the requirement.txt in the mono dependency file

chang-annie:
> Q: Do we need to update the readme file? since we now require the requirement.txt in the mono dependency file

spoke offline - no need to update docs because `requirements.txt` is automatically generated

examples/poetry/habushu-poetry-package-consumer/habushu_poetry_package_consumer-3.3.0.dev0.dist-info/METADATA
epwilkins, line 1: Q/S: Shouldn't this be generated by poetry and not checked in? Same for the other files in this directory.

epwilkins:
> > Q: Do we need to update the readme file? since we now require the requirement.txt in the mono dependency file
> 
> spoke offline - no need to update docs because `requirements.txt` is automatically generated

I don't think it's _always_ generated. Exporting requirements is a configuration flag. (Though I do believe it's on by default, it could still be changed.)

examples/poetry/habushu-poetry-package-consumer/habushu_poetry_package_consumer/verify_install.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/python
+from habushu_poetry_package.reusable_module.worker import SubWorker
+
+if __name__ == "__main__":
```
epwilkins: Q: Why is this being added? I don't see it being called, or referenced in documentation.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/RequirementsFileHelper.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.stream.Stream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+import org.apache.commons.codec.binary.Hex;
+
+import org.codehaus.plexus.util.StringUtils;
+import org.technologybrewery.habushu.HabushuException;
+
+public class RequirementsFileHelper {
```
epwilkins, line 26: A: Add java docs. I think `relocatePathRequirements` in particular would benefit from documentation.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/RequirementsFileHelper.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+        return path;
+    }
+
+    private static String calculateSha256(Path file) throws IOException, NoSuchAlgorithmException {
+        try {
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] hash = digest.digest(Files.readAllBytes(file));
+            return Hex.encodeHexString(hash);
+        } catch (IOException | NoSuchAlgorithmException e) {
```
epwilkins, line 183: A: We're catching these exceptions but the method header says we throw them.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -298,10 +358,13 @@ protected void setUpdateDockerfile(boolean update) {
...
     protected void performDockerfileUpdateForVirtualEnvironment(ProjectStack result) {
-        List<String> orderedProjectWheels = result.getAllProjects().stream()
+        String primaryWheel = result.getAllProjects().stream()
+                .filter(ProjectInfo::isPrimaryProject)
                 .map(ProjectInfo::getWheelName)
-                .collect(Collectors.toList());
-        ContainerizeDepsDockerfileHelper helper = new  ContainerizeDepsDockerfileHelper(this, orderedProjectWheels);
+                .collect(Collectors.toList()).get(0);
+
+        ContainerizeDepsDockerfileHelper.Wheel projectWheel = new ContainerizeDepsDockerfileHelper.Wheel(primaryWheel, "requirements.txt");
```
epwilkins: S: Use the `requirementsFilePath` from `ProjectInfo`

epwilkins: I actually wonder if it's necessary to have both `Wheel` and `ProjectInfo`. They seem a bit redundant with each other.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
```diff
@@ -298,10 +358,13 @@ protected void setUpdateDockerfile(boolean update) {
     }
 
     protected void performDockerfileUpdateForVirtualEnvironment(ProjectStack result) {
```
epwilkins: S: I think we can remove `ProjectStack` as an object and just pass the primary project info.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/RequirementsFileHelper.java
```diff
@@ -0,0 +1,181 @@
...
+    public Path getRequirementsFilePath() {
+        return requirementsFilePath;
+    }
+
+    public Path getRequirementsFileBaseDir() {
+        return requirementsFileBaseDir;
+    }
+
+    public List<Path> getPathBasedRequirements(){
+        return new ArrayList<>(getResolvedPathRequirements().values());
```
epwilkins, line 49: A: Might be worth making a note that this is a defensive copy, since Map is modifiable from the `values()` Collection

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
epwilkins, line 1: A: We can remove `getExtraWheels` from the mojo as well

examples/poetry/habushu-poetry-package-consumer/habushu_poetry_package_consumer-3.3.0.dev0.dist-info/METADATA
chang-annie: yes this was committed by mistake! removing this directory completely.

examples/poetry/habushu-poetry-package-consumer/habushu_poetry_package_consumer/verify_install.py
chang-annie: spoke offline, these files were added by mistake and will be removed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/ContainerizeDepsMojo.java
chang-annie: removed

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/RequirementsFileHelper.java
chang-annie: added javadocs to a few of the methods in this class

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/RequirementsFileHelper.java
chang-annie: updated header accordingly

PR #439, opened 2025-09-19
[#438] during c14n only add hashes if they are already present

### LINKED ISSUE #438: Containerization: Add support for hashless requirements files
## Background
When updating the requirements file with relocated paths for c14n, we always add the SHA-256 hash of the file since Habushu's requirements export includes hashes by default.  If the user configured `habushu.exportRequirementsWithHashes` to `false`, pip would complain that only some requirements have hashes.  We should only add the relocated path hashes if other requirements have hashes as well.

## Definition of Done
- [x] only add hashes if hashes are already present in file

## Test Steps
1. Create a Maven project with a Habushu module, and a Docker module that containerizes it
2. Include a local path dependency in the pyproject.toml of the Habushu module
3. Build with `exportRequirementsWithHashes` set to `false`
    ```sh
    mvn clean install -Dhabushu.exportRequirementsWithHashes=false
    ```
4. Verify the docker build succeeds.

epwilkins:


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/RequirementsFileHelper.java
```diff
@@ -9,9 +9,12 @@
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
```
epwilkins: A: unused import

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/util/RequirementsFileHelper.java
```diff
@@ -45,8 +50,8 @@ public Path getRequirementsFileBaseDir() {
     }
 
     // Defensive copy of resolved path-based requirements since the Map returned by getResolvedPathRequirements is modifiable
-    public List<Path> getPathBasedRequirements(){
-        return new ArrayList<>(getResolvedPathRequirements().values());
+    public Collection<Path> getPathBasedRequirements(){
+        return new HashSet<>(getResolvedPathRequirements().values());
```
epwilkins, line 51: I: Duplicate output made test failures harder to understand. The logic is identical whether duplicates exist or not. (Can be argued that it's more accurate to return only unique paths as well.)

habushu-maven-plugin/src/test/java/org/technologybrewery/habushu/ContainerizeDepsUpdateRequirementsSteps.java
```diff
@@ -91,8 +101,8 @@ public void theFollowingPathsAreReturnedPathBasedRequirements(List<String> paths
                 extra.add(path.toString());
             }
         }
-        Assertions.assertTrue(notFound.isEmpty(), "The following paths were not found: " + notFound);
-        Assertions.assertTrue(extra.isEmpty(), "The following paths were not expected: " + extra);
+        Assertions.assertEquals(notFound, List.of(), "Expected paths were not found.");
+        Assertions.assertEquals(List.of(), extra, "Extra unexpected paths were found");
```
epwilkins, line 103: I: Slightly more readable test output

habushu-maven-plugin/src/test/resources/specifications/containerize-dependencies-update-requirements.feature
```diff
@@ -88,4 +88,24 @@ Feature: Test updating requirements.txt file to rewrite path-based dependencies
       https://remote-wheelhouse/direct-package.whl
       # Random extra wheel checked into project
       file://${WHEEL_HOUSE}/leftpad-1.0.0-py3-none-any.whl
-      """
\ No newline at end of file
+      """
+
+    Scenario: Relocated paths include hashes if hashes are already present
```
epwilkins, line 93: I: This seems like the inverse scenario of the ticket, but before we weren't creating a test file to hash and just skipping adding a hash if the file didn't exist in the implementation. To support hashes we just always create a test file now, which caused this scenario to immediately pass but broke the old scenarios that excluded hashes.

woods-cpointe (approved):

lemuelwatts (approved):

PR #441, opened 2025-09-23
[437] fix poetry 2 migration

aaron-gary:
Fixes a bug where an empty line is mistaken for a new section.
Also fixes a bug where properties that are multi-line and need to be migrated gets hashed.

## Summary by Sourcery

Refactor PoetryToProjectMigration to accurately migrate entries by parsing the TOML into a section-to-lines map, tracking multi-line blocks, and regenerating the file from that structure

Bug Fixes:
- Stop treating empty lines as new section headers during migration
- Preserve and properly migrate multi-line array properties instead of hashing them

Enhancements:
- Replace line-by-line flag parsing with a tomlMap structure and introduce TomlLocationBlock to record multi-line property ranges

Tests:
- Update test fixture to use block-style arrays for multi-line fields
- Adjust feature spec expected counts to reflect additional migrated entries

### DIFF
--- a/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
+++ b/habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
@@ -5,7 +5,7 @@ import com.electronwill.nightconfig.core.file.FileConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.technologybrewery.baton.BatonException;
-import org.technologybrewery.habushu.util.TomlReplacementTuple;
+import org.technologybrewery.habushu.util.TomlLocationBlock;
 import org.technologybrewery.habushu.util.TomlUtils;
 
 import java.io.BufferedReader;
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
+import java.util.HashMap;
 
 /**
  * Provides functionality to migrate configuration entries from the [tool.poetry] section to the [project] section of a TOML file.
@@ -26,14 +27,13 @@ import java.util.LinkedHashMap;
  */
 public class PoetryToProjectMigration extends AbstractPoetryMigration {
     private static final Logger logger = LoggerFactory.getLogger(PoetryToProjectMigration.class);
-    private boolean hasProjectGroup = false;
-    private final List<String> entriesToMigrate = Arrays.asList("name", "description", "version", "authors", "maintainers", "license");
-    private final List<String> existingProjectKeys = new ArrayList<>();
-    private final Map<String, TomlReplacementTuple> replacements = new LinkedHashMap<>();
+    public static final String PROJECT_SECTION_WITH_BRACKETS = "[" + TomlUtils.PROJECT + "]";
+    public static final String NO_SECTION_HEADER = "no_section_header";
+    private final List<String> entriesToMigrate = Arrays.asList("name", "description", "version", "license");
+    private final Map<String, List<String>> tomlMap = new LinkedHashMap<>();
 
     @Override
     protected boolean shouldExecuteOnFile(File file) {
-        replacements.clear();
         boolean shouldExecute = false;
         Config.setInsertionOrderPreserved(true);
 
@@ -48,33 +48,10 @@ public class PoetryToProjectMigration extends AbstractPoetryMigration {
 
                         // check if any of the entries under [tool.poetry] match one of "name", "description", or "version"
                         String groupEntryName = groupEntry.getKey();
-                        Object groupEntryRhs = groupEntry.getValue();
-                        String groupEntryRhsAsString = null;
 
                         if (entriesToMigrate.stream().anyMatch(e -> e.equalsIgnoreCase(groupEntryName))) {
-                            if (groupEntryRhs instanceof List<?>){
-                                List<String> listValues = poetryGroup.get(groupEntryName);
-                                List<String> listOfStrings = new ArrayList<>(listValues);
-                                groupEntryRhsAsString = convertListOfStringsToString(listOfStrings);
-                            } else {
-                                groupEntryRhsAsString = (String) groupEntryRhs;
-                            }
-
-                            logger.info("Found [{}] group entry to migrate! ({} = {})", TomlUtils.TOOL_POETRY, groupEntryName, groupEntryRhsAsString);
-                            TomlReplacementTuple replacementTuple = new TomlReplacementTuple(groupEntryName, groupEntryRhsAsString, "");
-                            replacements.put(groupEntryName, replacementTuple);
                             shouldExecute = true;
                         }
-                    }
-                }
-                // process the [project] section and record existing keys
-                Optional<Config> projectGroupEntries = tomlFileConfig.getOptional(TomlUtils.PROJECT);
-                if (projectGroupEntries.isPresent()){
-                    hasProjectGroup = true;
-                    Config projectConfig = projectGroupEntries.get();
-                    // iterate over the entries and add each key to the set
-                    for (Config.Entry entry : projectConfig.entrySet()) {
-                        existingProjectKeys.add(entry.getKey());
                     }
                 }
             }
@@ -84,61 +61,64 @@ public class PoetryToProjectMigration extends AbstractPoetryMigration {
 
     @Override
     protected boolean performMigration(File pyProjectTomlFile) {
-        boolean injectAfterNextEmptyLine = false;
-        boolean inToolPoetrySection = false;
-        boolean inProjectSection = false;
-        boolean dependenciesInjected = false;
-        StringBuilder fileContent = new StringBuilder();
+        Map<String, TomlLocationBlock> multiLineProperties = findMultilineProperties(pyProjectTomlFile);
 
         try (BufferedReader reader = new BufferedReader(new FileReader(pyProjectTomlFile))) {
-            String line = reader.readLine();
+            loadTomlFileToMap(reader, multiLineProperties);
 
+            StringBuilder fileContent = generateNewTomlFileContentsFromUpdatedStructure();
+
+            TomlUtils.writeTomlFile(pyProjectTomlFile, fileContent.toString());
+        } catch (IOException e) {
+            throw new BatonException("Problem reading pyproject.toml for migration!", e);
+        }
+
+        return true;
+    }
+
+    public Map<String, TomlLocationBlock> findMultilineProperties(File pyProjectTomlFile) {
+        /*
+         * Parses the Toml file and identifies where items are represented on multiple lines.
+         * If an item is multi-line, the start and end locations are determined and added
+         * to a list.
+         */
+        Map<String, TomlLocationBlock> multiLineProperties = new HashMap<>();
+        try (BufferedReader reader = new BufferedReader(new FileReader(pyProjectTomlFile))) {
+            boolean inArray = false;
+            String currentSection = null;
+            String currentName = null;
+            int startLine = -1;
+
+            String line = reader.readLine();
+            int lineCount = 1;
             while (line != null) {
-                boolean addLine = true;
-                boolean isEmptyLine = line.isBlank();
                 String trimmedLine = line.strip();
 
                 if (trimmedLine.startsWith("[") && trimmedLine.endsWith("]")){
-                    if (trimmedLine.equals("[" + TomlUtils.TOOL_POETRY + "]")) {
-                        inToolPoetrySection = true;
-                        inProjectSection = false;
-                    } else if (trimmedLine.equals(("[" + TomlUtils.PROJECT + "]"))) {
-                        inToolPoetrySection = false;
-                        inProjectSection = true;
-                    } else {
-                        inToolPoetrySection = false;
-                        inProjectSection = false;
-                    }
+                    currentSection = trimmedLine.replaceAll(" ", "");
                 }
-                if (trimmedLine.contains(TomlUtils.EQUALS)) {
-                    if (inToolPoetrySection) {
-                        String key = line.substring(0, line.indexOf(TomlUtils.EQUALS)).strip();
-                        TomlReplacementTuple matchedTuple = replacements.get(key);
-                        if (matchedTuple != null) {
-                            // skip this line, we will add it back to [project] later
-                            addLine = false;
+
+                if (!inArray) {
+                    if (trimmedLine.matches("^[a-zA-Z0-9_.-]+\\s*=\\s*\\[$")) {
+                        // matches lines like the following:
+                        // packages = [
+                        inArray = true;
+                        currentName = getKeyFromLine(trimmedLine);
+                        startLine = lineCount;
+                    }
+                } else {
+                    if (trimmedLine.equals("]")) {
+                        inArray = false;
+
+                        // We only want to create a location object for items under [tool.poetry]
+                        // and in the list of items we are migrating
+                        if(currentSection.equals("[" + TomlUtils.TOOL_POETRY + "]") && entriesToMigrate.contains(currentName)) {
+                            multiLineProperties.put(currentName, new TomlLocationBlock(currentName, startLine, lineCount));
                         }
                     }
-                    if (!hasProjectGroup || (inProjectSection && !replacements.isEmpty())) {
-                        injectAfterNextEmptyLine = true;
-                    }
                 }
-                if (isEmptyLine && injectAfterNextEmptyLine){
-                    if (hasProjectGroup){
-                        fileContent = injectDependencies(fileContent, existingProjectKeys, replacements);
-                        dependenciesInjected = true;
-                        injectAfterNextEmptyLine = false;
-                    } else if (!dependenciesInjected){
-                        fileContent.append("\n[").append(TomlUtils.PROJECT).append("]\n");
-                        fileContent = injectDependencies(fileContent, existingProjectKeys, replacements);
-                        injectAfterNextEmptyLine = false;
-                        hasProjectGroup = true;
-                        dependenciesInjected = true;
-                    }
-                }
-                if (addLine) {
-                    fileContent.append(line).append("\n");
-                }
+
+                lineCount++;
 
                 line = reader.readLine();
             }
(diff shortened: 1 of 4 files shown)


habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -26,10 +28,13 @@
  */
 public class PoetryToProjectMigration extends AbstractPoetryMigration {
     private static final Logger logger = LoggerFactory.getLogger(PoetryToProjectMigration.class);
-    private boolean hasProjectGroup = false;
+    public static final String PROJECT_SECTION_WITH_BRACKETS = "[" + TomlUtils.PROJECT + "]";
+//    private boolean hasProjectGroup = false;
```
epwilkins: A: Remove

epwilkins: Actually this is causing a compile error on line 78

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -26,10 +28,13 @@
...
     private final List<String> entriesToMigrate = Arrays.asList("name", "description", "version", "authors", "maintainers", "license");
     private final List<String> existingProjectKeys = new ArrayList<>();
     private final Map<String, TomlReplacementTuple> replacements = new LinkedHashMap<>();
+    private final Map<String, List<String>> tomlMap = new LinkedHashMap<>();
+    private String currentSection = new String();
```
epwilkins: S: Convert to local variable. Confused me that we were saving this in object state and was looking for where we used it outside of `loadTomlFileToMap`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -26,10 +28,13 @@
  */
 public class PoetryToProjectMigration extends AbstractPoetryMigration {
     private static final Logger logger = LoggerFactory.getLogger(PoetryToProjectMigration.class);
-    private boolean hasProjectGroup = false;
+    public static final String PROJECT_SECTION_WITH_BRACKETS = "[" + TomlUtils.PROJECT + "]";
+//    private boolean hasProjectGroup = false;
     private final List<String> entriesToMigrate = Arrays.asList("name", "description", "version", "authors", "maintainers", "license");
     private final List<String> existingProjectKeys = new ArrayList<>();
     private final Map<String, TomlReplacementTuple> replacements = new LinkedHashMap<>();
```
epwilkins: S: These are now created but never used. If we don't need them we should remove them entirely.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -84,75 +89,247 @@ protected boolean shouldExecuteOnFile(File file) {
...
+                } else {
+                    // Comments and blank lines
+                    tomlMap.get(currentSection).add(trimmedLine);
+                }
+            }
+
+            line = reader.readLine();
+            lineCount++;
         }
+        System.out.println("something");
```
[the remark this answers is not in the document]
aaron-gary: - [x] Remove stray print

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Good call.  The AI bot called this one out as well

- [x] Remove unused comment

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -67,92 +71,253 @@ protected boolean shouldExecuteOnFile(File file) {
...
-                            addLine = false;
-                        }
-                    }
-                    if (!hasProjectGroup || (inProjectSection && !replacements.isEmpty())) {
-                        injectAfterNextEmptyLine = true;
+                } else {
+                    // look for closing bracket
+                    if (trimmedLine.equals("]")) {
+                        arrays.add(new TomlLocationBlock(currentName, startLine, lineCount));
+                        multiLineProperties.put(currentName, new TomlLocationBlock(currentName, startLine, lineCount));
```
epwilkins: Q: Do we know that pyproject.toml keys will only ever be arrays or simple, and not ever maps/objects (`key = {...}`)?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -67,92 +71,253 @@ protected boolean shouldExecuteOnFile(File file) {
...
-                    if (trimmedLine.equals("[" + TomlUtils.TOOL_POETRY + "]")) {
-                        inToolPoetrySection = true;
-                        inProjectSection = false;
-                    } else if (trimmedLine.equals(("[" + TomlUtils.PROJECT + "]"))) {
-                        inToolPoetrySection = false;
-                        inProjectSection = true;
-                    } else {
-                        inToolPoetrySection = false;
-                        inProjectSection = false;
+                if (!inArray) {
```
epwilkins, line 101: A: Unused?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -67,92 +71,253 @@ protected boolean shouldExecuteOnFile(File file) {
...
-                    fileContent.append(line).append("\n");
-                }
+
+                lineCount++;
 
                 line = reader.readLine();
             }
         } catch (IOException e) {
             throw new BatonException("Problem reading pyproject.toml for migration!", e);
         }
```
epwilkins, line 127: Q: What if the same key exists in multiple sections?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: - [x] Move variable to local scope

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -67,92 +71,253 @@ protected boolean shouldExecuteOnFile(File file) {
...
+                            lineCount = checkPropertyForMultilineAndProcess(
+                                    reader,
+                                    multiLineProperties,
+                                    currentKeyName,
+                                    lineCount,
+                                    PROJECT_SECTION_WITH_BRACKETS
+                            );
+                        }
+                    } else {
+                        if(currentSection.equals(PROJECT_SECTION_WITH_BRACKETS)) {
```
epwilkins, line 179: S: We should just not run the migration if they already have a `[project]` section, as that would imply they already updated the files the way they want.  Then `currentSection.equals(PROJECT_SECTION_WITH_BRACKETS)` is never true and we can simplify this logic.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -67,92 +71,253 @@ protected boolean shouldExecuteOnFile(File file) {
...
+    }
+
+    /**
+     * If the current section is not already [project] and the item is in the list of migratable
+     * items then we need to migrate.
+     * @param currentKeyName the item being processed
+     * @return If the item needs to be migrated
+     */
+    private boolean itemNeedsToBeMigrated(String currentKeyName) {
+        return entriesToMigrate.contains(currentKeyName) && !currentSection.equals(PROJECT_SECTION_WITH_BRACKETS);
```
epwilkins: S:
```suggestion
        return currentSection.equals(TomlUtils.TOOL_POETRY) && entriesToMigrate.contains(currentKeyName);
```

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -67,92 +71,253 @@ protected boolean shouldExecuteOnFile(File file) {
...
+                                        multiLineProperties,
+                                        currentKeyName,
+                                        lineCount,
+                                        PROJECT_SECTION_WITH_BRACKETS
+                                );
+
+                            }
+                        } else {
+                            // Non-migrate items
+                            tomlMap.get(currentSection).add(trimmedLine);
```
epwilkins, line 194: A: we should add `line` instead of `trimmedLine` here and ln 185

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -48,111 +48,260 @@ protected boolean shouldExecuteOnFile(File file) {
...
+     * process the next few lines until the whole item is processed.
+     * @param reader the file being processed
+     * @param multiLineProperties a list of items that are multi-line
+     * @param currentKeyName the current item being processed
+     * @param lineCount the current position within the Toml file
+     * @param sectionToAddTo the name of the Toml section to add items to
+     * @return the last line that was processed
+     * @throws IOException
+     */
+    private int checkPropertyForMultilineAndProcess(BufferedReader reader, Map<String, TomlLocationBlock> multiLineProperties, String currentKeyName, int lineCount, String sectionToAddTo) throws IOException {
```
epwilkins: A: Might be more straight-forward to construct and return a List. Then at the call site, `tomlMap.get(currentSection).addAll(returnList); lineCount += returnList.size()`. Would let you drop params `lineCount`, and `sectionToAddTo`.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: I prefer simpler logic, but why do we have test cases where [project] exists and items under [tool.poetry] still need migration?  Is there a use case where both can exist?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Good call.  Removed several lines of unused code.

- [x] remove unused code

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -48,111 +48,260 @@ protected boolean shouldExecuteOnFile(File file) {
...
+                currentSection = trimmedLine;
+            } else {
+                if(trimmedLine.contains(TomlUtils.EQUALS)) {
+                    // Section items
+                    currentKeyName = getKeyFromLine(trimmedLine);;
+
+                    if(itemNeedsToBeMigrated(currentKeyName, currentSection)){
+                        ensureProjectSectionExistsInTomlMap();
+                        // check for duplicate
+                        if(!checkIfItemAlreadyExistsInProjectSection(tomlMap.get(PROJECT_SECTION_WITH_BRACKETS), trimmedLine)) {
```
epwilkins, line 165: S: If we're only migrating when `[project]` is not present, there will never be a duplicate as a duplicate key in `[tool.poetry]` is invalid (poetry fails).

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Good question.  I looked through the night-config repo and they have a few examples where they use json/dict values.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: This is used

epwilkins (approved):

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
epwilkins: I think I got duped by the refreshed code after you removed some line. I meant the local `arrays` list in this method.  We add to it, but never query it or return it I think.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
epwilkins: This is also misaligned from submitting on a stale line number. 😞  I will find where I meant to leave this and update.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Oh!  Gotcha.  I will look into it

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
epwilkins: Let me double-check, but I don't think so.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Good catch, it isn't being used 

- [x] Remove unused arrays variable

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: I'm not sure what you mean.  If the same keys exist in other sections, it will only matter if we are migrating them to the [project] section... in which case it's first come, first served.

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
epwilkins: Yeah. The `project` section was ignored/unsupported prior to Poetry 2.0 (might have even thrown an error.) So we can assume that if it _is_ present, the project has already updated their pyproject.toml to leverage Poetry 2.0 either manually or via this migration. If the updated it manually, they would have intentionally made the decision not to move some properties even if we would have expected them to so it doesn't make sense for us to force it on them.

aaron-gary: But that does sounds like a good DoD item to add in a follow on ticket

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
epwilkins: This was supposed to be on [line 127](https://github.com/TechnologyBrewery/habushu/pull/441/files/4d7d0faf0a24f17e592bb180688721e43714a170#diff-38162720f560536b085b29e87203df73970940f60ef981c12ae8f3c25e791a11R127). (And the comment on line 127 was meant to be on [line 116](https://github.com/TechnologyBrewery/habushu/pull/441/files/4d7d0faf0a24f17e592bb180688721e43714a170#diff-38162720f560536b085b29e87203df73970940f60ef981c12ae8f3c25e791a11R116))

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Yes, that's a very good reason to not migrate when [project] is present

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: I like this suggestion.

- [x] Refactor multiline call

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: good call

- [x] switch to untouched line

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Good suggestion.  One small correction is that we have to wrap TomlUtils.TOOL_POETRY in square brackets.  FYI

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
aaron-gary: Ok, I see the dilemma now

epwilkins: I think we only care about multiline properties under the `tool.poetry` section anyway right? Maybe we could just ignore things outside of that then we don't have to worry about duplicates?

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
```diff
@@ -48,111 +48,281 @@ protected boolean shouldExecuteOnFile(File file) {
...
+    }
+
+    /**
+     * If the current section is not already [project] and the item is in the list of migratable
+     * items then we need to migrate.
+     * @param currentKeyName the item being processed
+     * @return If the item needs to be migrated
+     */
+    private boolean itemNeedsToBeMigrated(String currentKeyName, String currentSection) {
+        return currentSection.equals("[" + TomlUtils.TOOL_POETRY + "]") && entriesToMigrate.contains(currentKeyName);
```
epwilkins, line 242: Remove [whitespace](https://toml.io/en/v1.0.0#table) from `currentSection`

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
epwilkins: The [spec](https://toml.io/en/v1.0.0#inline-table) for TOML does not allow for multi-line, inline tables (had to look at the BNF grammar to verify)

habushu-maven-plugin/src/main/java/org/technologybrewery/habushu/migration/poetryv2migrations/PoetryToProjectMigration.java
epwilkins: We had some crossed wires on this feed back.  The current code handles duplicate section headers, this comment is about duplicate multiline keys between sections. The decision was to update the findMultilineProperties method to only add to the map if the key is something we intend to migrate and it's in the `tool.poetry` section. Otherwise, it's just copied line-by-line like everything else.

epwilkins (approved):

